PR #20, opened 2025-02-07
fix: execute valid bash cmd when running init as root

### LINKED ISSUE #19: init fails with command not found
Upon attempting to initialise a new machine with `uc machine init`, the program exits with an error:

```
$ go run cmd/uncloud/main.go machine init -i ~/.ssh/id_github root@<redacted>
Downloading Uncloud install script: https://raw.githubusercontent.com/psviderski/uncloud/refs/heads/main/scripts/install.sh
bash: line 1: : command not found
curl: (23) Failure writing output to destination
Error: provision machine: download and run install script: run command on remote host: Process exited with status 127
exit status 1
````

Will send a PR for a fix if I come up with one.

### PR DESCRIPTION
**Pull request by cedws**
When executed as root, the bash command generated contained a bunch of quotes due to the empty `sudoPrefix` and `env`. This changes the code to emit plain pipe to bash when executed as root.

Closes #19.

### DIFF
--- a/internal/cli/machine.go
+++ b/internal/cli/machine.go
@@ -17,24 +17,36 @@ type RemoteMachine struct {
 	KeyPath string
 }
 
+func installCmd(user string) string {
+	sudoPrefix := "sudo"
+	// Add the SSH user (non-root) to the uncloud group to allow access to the Uncloud daemon unix socket.
+	env := "UNCLOUD_GROUP_ADD_USER=" + user
+
+	curlBashCmd := fmt.Sprintf(
+		"curl -fsSL %s | %s %s bash", sshexec.Quote(installScriptURL), sudoPrefix, sshexec.Quote(env),
+	)
+
+	if user == "root" {
+		curlBashCmd = fmt.Sprintf(
+			"curl -fsSL %s | bash", sshexec.Quote(installScriptURL),
+		)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/cli/machine.go [hunk 1]
```diff
@@ -17,24 +17,36 @@ type RemoteMachine struct {
...
+
+	installCmd := installCmd(user)
 
 	fmt.Println("Downloading Uncloud install script:", installScriptURL)
-	curlBashCmd := fmt.Sprintf(
-		"curl -fsSL %s | %s %s bash", sshexec.Quote(installScriptURL), sudoPrefix, sshexec.Quote(env),
-	)
-	cmd := sshexec.QuoteCommand("bash", "-c", "set -o pipefail; "+curlBashCmd)
+
+	cmd := sshexec.QuoteCommand("bash", "-c", "set -o pipefail; "+installCmd)
```
psviderski, line 49: It perhaps makes sense to return this final command from the `installCmd` to encapsulate the `set -o pipefail` logic which assumes that the `curlBashCmd` contains pipes. But it's not a big deal and up to you if you'd like to address it.

psviderski (approved):
Thank you for the fix! I can confirm that I have been only testing machine provisioning using non-root users as that was a more complex use case.

PR #21, opened 2025-02-07
feat: add ssh_key_file field to connections config

cedws:
This enables private key authentication by setting the `identity_file` field under a connection in `~/.config/uncloud/config.toml`. It also fixes a possible nil panic in this code path.

```
current_cluster = "default"

[clusters]
[clusters.default]

[[clusters.default.connections]]
ssh = "root@<redacted>"
identity_file = "~/.ssh/id_github"
```

### DIFF
--- a/internal/cli/config/connection.go
+++ b/internal/cli/config/connection.go
@@ -14,11 +14,11 @@ const (
 )
 
 type MachineConnection struct {
-	SSH          SSHDestination `toml:"ssh,omitempty"`
-	TCP          netip.AddrPort `toml:"tcp,omitempty"`
-	Host         string         `toml:"host,omitempty"`
-	PublicKey    secret.Secret  `toml:"public_key,omitempty"`
-	IdentityFile string         `toml:"identity_file,omitempty"`
+	SSH        SSHDestination `toml:"ssh,omitempty"`
+	TCP        netip.AddrPort `toml:"tcp,omitempty"`
+	Host       string         `toml:"host,omitempty"`
+	PublicKey  secret.Secret  `toml:"public_key,omitempty"`
+	SSHKeyFile string         `toml:"ssh_key_file,omitempty"`
 }
 
 // SSHDestination represents an SSH destination string in the canonical form of "user@host:port".
--- a/internal/cli/config/connection.go
+++ b/internal/cli/config/connection.go
@@ -14,10 +14,11 @@ const (
 )
 
 type MachineConnection struct {
-	SSH       SSHDestination `toml:"ssh,omitempty"`
-	TCP       netip.AddrPort `toml:"tcp,omitempty"`
-	Host      string         `toml:"host,omitempty"`
-	PublicKey secret.Secret  `toml:"public_key,omitempty"`
+	SSH          SSHDestination `toml:"ssh,omitempty"`
+	TCP          netip.AddrPort `toml:"tcp,omitempty"`
+	Host         string         `toml:"host,omitempty"`
+	PublicKey    secret.Secret  `toml:"public_key,omitempty"`
+	IdentityFile string         `toml:"identity_file,omitempty"`
 }
 
 // SSHDestination represents an SSH destination string in the canonical form of "user@host:port".
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -13,9 +13,11 @@ func ExpandHomeDir(path string) string {
 		return path
 	}
 	if path[0] == '~' {
-		// TODO: Improve compat with other OSes
-		path = strings.Replace(path, "~", "${HOME}", 1)
-		return os.ExpandEnv(path)
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return path
+		}
+		return strings.Replace(path, "~", home, 1)
 	}
 	return path
 }
--- a/internal/fs/fs.go
+++ b/internal/fs/fs.go
@@ -5,7 +5,20 @@ import (
 	"os"
 	"os/user"
 	"strconv"
+	"strings"
 )
+
+func ExpandHomeDir(path string) string {
+	if len(path) == 0 {
+		return path
+	}
+	if path[0] == '~' {
+		// TODO: Improve compat with other OSes
+		path = strings.Replace(path, "~", "${HOME}", 1)
+		return os.ExpandEnv(path)
+	}
+	return path
+}
 
 // LookupUIDGID returns the user and group IDs for the given username.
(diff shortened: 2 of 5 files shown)


internal/fs/fs.go
```diff
@@ -5,8 +5,21 @@ import (
...
+	"strings"
 )
 
+func ExpandHomeDir(path string) string {
+	if len(path) == 0 {
+		return path
+	}
+	if path[0] == '~' {
+		// TODO: Improve compat with other OSes
+		path = strings.Replace(path, "~", "${HOME}", 1)
```
psviderski: How about using a function from the standard library `os.UserHomeDir` that returns the user's home dir and works on main OSs? It essentially does what you've done here for macos/linux. Please see how it's used in a couple other places in the code base: https://github.com/psviderski/uncloud/blob/f85b5dd01d60d30cb153c6af2b9677e60668f613/cmd/uncloud/main.go#L22-L28

It would be nice to migrate other places to use `ExpandHomeDir` eventually.

internal/cli/config/connection.go
```diff
@@ -14,10 +14,11 @@ const (
...
 type MachineConnection struct {
-	SSH       SSHDestination `toml:"ssh,omitempty"`
-	TCP       netip.AddrPort `toml:"tcp,omitempty"`
-	Host      string         `toml:"host,omitempty"`
-	PublicKey secret.Secret  `toml:"public_key,omitempty"`
+	SSH          SSHDestination `toml:"ssh,omitempty"`
+	TCP          netip.AddrPort `toml:"tcp,omitempty"`
+	Host         string         `toml:"host,omitempty"`
+	PublicKey    secret.Secret  `toml:"public_key,omitempty"`
+	IdentityFile string         `toml:"identity_file,omitempty"`
```
psviderski: I'm thinking whether we should maintain consistency between config option names and flag names. The flag to provide a path to an SSH private key for `machine init|add` commands is the following:
```
  -i, --ssh-key string   path to SSH private key for SSH remote login. (default ~/.ssh/id_*)
```
The short version `-i` was chosen to resemble the commonly known `ssh` command (`machine init|add -i ~/.ssh/id_rsa user@host`). The parameter for the flag in `ssh` CLI is indeed called `identity_file` but I really rarely see people calling it an identity file when talking about SSH. Most of the time it's just an SSH private key or just SSH key, hence the long flag name to make it user friendlier.

What do you think about calling the config option `ssh_key` as well to keep it consistent with the flag? Any better name options you can think of?

psviderski (commented):
Nice one! ❤️ I've left a couple comments to discuss naming. Please take a look and let me know if you have time to address them.

BTW, am I right that we still don't populate the `identity_file` property in the config when initialising or adding a new machine to the cluster? It will work only if it's manually specified, right? If it's true, I will create an issue to not forget to implement this.

internal/fs/fs.go
cedws: Thanks, resolved.

internal/cli/config/connection.go
cedws: Yeah wasn't sure about this, the reason I chose `IdentityFile` was to emphasise the key comes from a *file* and shouldn't be inline in the config.

Other ideas:

* `PrivateKeyFile` - too verbose?
* `PrivateKey` - unclear
* `KeyFile` - unclear, which key?

I wonder if it might be nicer to avoid SSH options in this config entirely and use the user's SSH config.

cedws:
>BTW, am I right that we still don't populate the identity_file property in the config when initialising or adding a new machine to the cluster?

Just added a change to do this. I think other properties still won't be though.

internal/cli/config/connection.go
psviderski: `SSHKeyFile` is pretty clear and doesn't look too bad in my opinion but I'll leave the final decision to you.

To give you some more context on the connections config, originally I tried to use a WireGuard identity for connecting to and managing a cluster. You can see a `WireGuardConnector` [here](https://github.com/psviderski/uncloud/blob/9e44d303289b0d38ce98468598b6441c6a875b79/internal/cli/client/connector/wireguard.go) which uses `Host` and `PublicKey` options. It turned out to be not very reliable as WireGuard or UDP protocol can be blocked in some networks so I decided to stick to SSH for now as it's widely used and easy to troubleshoot. Maybe we should remove the unused params to not confuse developers.

There is also a `TCP` one that is only used by [`ucind`](https://github.com/psviderski/uncloud/blob/8bddbec7bdeb9979e052a184cd9a05325f772f7b/internal/ucind/machine.go) (**u**n**c**loud **in** **d**ocker, a similar idea to [kind](https://kind.sigs.k8s.io/)) to connect to machines running as Docker containers which I use for e2e tests. Not sure if the `TCP` connector will ever be useful for any other use cases.

Also, I haven't come up with a proper multi-user or fine-grained access control model. At the moment it's pretty basic: if you have root access on any of the machines, you get root access to the entire cluster. I've recently discovered that [Incus](https://linuxcontainers.org/incus/docs/main/authorization/) has the same behavior if a user is in `incus-admin` group or is `root`. I don't want to introduce a PKI system for authenticating and authorizing users (e.g. TLS certs, similar to what k8s uses) as it becomes a burden to maintain and it's not very user friendly. The SSH itself is essentially a PKI system that users should already have for accessing their machines so I decide to try to use it and not introduce another one. Maybe we can try to use Linux users for access control or at least to be able to log what action is performed by which user. I think [kamal](https://kamal-deploy.org/) does something similar.

>I wonder if it might be nicer to avoid SSH options in this config entirely and use the user's SSH config.

Do you mean looking for something like this in `~/.ssh/config` and extracting a user and path to the key from there?
```
Host my-machine-ip-or-fqdn
    User foo
    IdentityFile ~/.ssh/id_bar
```

This could work as a heuristic but won't work in all cases. I personally don't really need the "path to a key" option because I always add my key to the ssh-agent. In general, I would like the UX to be nice, namely you init a machine specifying the path to a key (if it's not in the ssh-agent) and then you should instantly be able to run other commands to list machines or run a service without configuring anything manually. So it feels we need to preserve the path to a key somewhere to reuse it later if the user specified it when initializing a machine. Did I get your idea right or you did you mean something different?

cedws: I've raised a discussion about the CLI DX here: https://github.com/psviderski/uncloud/discussions/22

I've taken your suggestion and renamed this field to `SSHKeyFile`.

psviderski: Thanks!

psviderski (approved):

PR #40, opened 2025-03-16
docs: add basic user guide

cedws:


docs/user_guide.md
```diff
@@ -0,0 +1,57 @@
...
+
+* To reserve a domain name, run `uc dns reserve`
+* To release a domain name, run `uc dns release`.
+* To see the domain name, run `uc dns show`
+
+### Running a service
+
+Services on an Uncloud cluster can be managed with `uc service`.
+
+You can run a service with two replicas that expose port 80 like the following:
```
psviderski, line 45: Maybe we need to clarify that this will expose the container port 80 as HTTP via the Caddy reverse proxy that has to be deployed. And also mention the `uc caddy deploy` command to deploy Caddy if it was disabled with `--no-caddy` earlier.

docs/user_guide.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+uc service run --mode replicated -p 80/http --replicas 2 nginxdemos/hello
```
psviderski: The `replicated` mode is the default, do you think we should still specify it here?

psviderski (approved):
Thank you so much for writing this guide. I left a few minor comments. Happy to merge as is and address them later if you like

psviderski (approved):
Nice work!

PR #50, opened 2025-04-25
feat: add version flag to main cmd

### LINKED ISSUE #42: Embed version and commit hash into uc and uncloudd binaries
Implement `version` command for `uc` and `uncloudd` that prints the current version and commit hash the binary has been built from.

cedws:
Closes #42.

This adds a `--version` flag to the main Cobra command.

Since this only applies to the main command this shouldn't conflict `--version` flags in subcommands.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -24,6 +24,8 @@ builds:
     goarch:
       - amd64
       - arm64
+    ldflags:
+      - -s -w -X github.com/psviderski/uncloud/internal/version.version={{ .Version }}
 
   - id: uncloudd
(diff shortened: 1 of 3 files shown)


psviderski (approved):
Thank you! 👍 
BTW, I've recently discovered that Go 1.24 introduced the main module's version (`BuildInfo.Main.Version`) in the compiled binary based on the version control system tag or commit: https://antonz.org/go-1-24/#main-modules-version
Although perhaps it's still better to use one source of truth that stamps the version which is goreleases in our case. This way it will be consistent everywhere. Exactly how you did it 👍

PR #54, opened 2025-04-30
feat: add support for Compose profiles

### LINKED ISSUE #51: Add support for Compose profiles
Profiles are a useful feature that I personally use to mask what services I want to run for local dev. For example, I might only want to start only the database locally and run the application outside of Docker.

https://docs.docker.com/compose/how-tos/profiles/

cedws:
Closes #51.

### DIFF
--- a/cmd/uncloud/deploy.go
+++ b/cmd/uncloud/deploy.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	composecli "github.com/compose-spec/compose-go/v2/cli"
 	"github.com/docker/compose/v2/pkg/progress"
 	"github.com/psviderski/uncloud/internal/cli"
 	"github.com/psviderski/uncloud/pkg/api"
@@ -17,6 +18,7 @@ import (
 
 type deployOptions struct {
 	files    []string
+	profiles []string
 	services []string
 
 	context string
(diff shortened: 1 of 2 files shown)


cmd/uncloud/deploy.go
```diff
@@ -49,9 +53,22 @@ func NewDeployCommand() *cobra.Command {
 	return cmd
 }
 
+// projectOpts returns the project options for the Compose file(s).
+func projectOpts(opts deployOptions) []composecli.ProjectOptionsFn {
+	projectOpts := []composecli.ProjectOptionsFn{}
+
+	if len(opts.profiles) > 0 {
+		projectOpts = append(projectOpts, composecli.WithProfiles(opts.profiles))
```
psviderski: Did you consider using `WithDefaultProfiles` that falls back to profiles specified via the `COMPOSE_PROFILES` environment variable to be more compatible with the default `docker compose` behaviour?

https://github.com/docker/compose/blob/5bb46035cf12e5b2fc2f3b09522a392b7b753120/cmd/compose/compose.go#L418-L419

psviderski (approved):
Looks great, just one little question about choosing the option `WithProfiles` vs `WithDefaultProfiles`

cmd/uncloud/deploy.go
cedws: Nice one, didn't see that. Thanks.

PR #58, opened 2025-05-02
added validition for service name

Tova-Rozovsky:
I added a code snippet to the validate function to ensure that the service name is valid.
I used a regular expression to check that the service name conforms to the expected DNS naming conventions.

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -31,6 +31,7 @@ const (
 )
 
 var serviceIDRegexp = regexp.MustCompile("^[0-9a-f]{32}$")
+var dnsLabelRegexp = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`)
 
 func ValidateServiceID(id string) bool {
 	return serviceIDRegexp.MatchString(id)
@@ -104,13 +105,17 @@ func (s *ServiceSpec) Validate() error {
 		return fmt.Errorf("invalid mode: %q", s.Mode)
 	}
 
+func (s *ServiceSpec) Validate() error {
+	if s.Mode != "someExpectedMode" {
+		return fmt.Errorf("invalid mode: %q", s.Mode)
+	}
+
 	if s.Name != "" {
-		if len(s.Name) > 253 {
-			return fmt.Errorf("service name too long (max 253 characters): %q", s.Name)
+		if len(s.Name) > 63 {
+			return fmt.Errorf("service name too long (max 63 characters): %q", s.Name)
 		}
-		dnsSubdomainRegexp := regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(?:\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
-		if !dnsSubdomainRegexp.MatchString(s.Name) {
-			return fmt.Errorf("invalid service name %q: must be a valid DNS subdomain", s.Name)
+		if !dnsLabelRegexp.MatchString(s.Name) {
+			return fmt.Errorf("invalid service name: %q. Must comply with RFC 1123 label format", s.Name)
 		}
 	}
 
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -104,18 +104,13 @@ func (s *ServiceSpec) Validate() error {
 	default:
 		return fmt.Errorf("invalid mode: %q", s.Mode)
 	}
-
-func (s *ServiceSpec) Validate() error {
-	if s.Mode != "someExpectedMode" {
-		return fmt.Errorf("invalid mode: %q", s.Mode)
-	}
-
+	
 	if s.Name != "" {
 		if len(s.Name) > 63 {
 			return fmt.Errorf("service name too long (max 63 characters): %q", s.Name)
 		}
 		if !dnsLabelRegexp.MatchString(s.Name) {
-			return fmt.Errorf("invalid service name: %q. Must comply with RFC 1123 label format", s.Name)
+			return fmt.Errorf("invalid service name: %q.must be 1-63 characters, lowercase letters, numbers, and dashes only; must start and end with a letter or number", s.Name)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/api/service.go
```diff
@@ -104,7 +104,15 @@ func (s *ServiceSpec) Validate() error {
 		return fmt.Errorf("invalid mode: %q", s.Mode)
 	}
 
-	// TODO: validate the service name is a valid DNS label.
+	if s.Name != "" {
+		if len(s.Name) > 253 {
+			return fmt.Errorf("service name too long (max 253 characters): %q", s.Name)
+		}
+		dnsSubdomainRegexp := regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(?:\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
```
psviderski: I'm sorry I perhaps confused with referring k8s docs on DNS subdomain names as I think a service name should actually be restricted to [RFC 1123 Label Names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) which don't allow dots in the names.

The reason is that allowing dots creates ambiguity about domain boundaries. At the moment, we use the format `<service-name>.internal` for internal DNS resolution. This ambiguity becomes problematic when we introduce additional domain structures like `machines.<service-name>.internal` in the future. When performing a lookup for `machines.foo.internal` it's unclear whether we want to look up machines for service `foo` or addresses for service `machines.foo`. Ever worse, one service (such as `machines.foo`) may accidentally or intentionally hijack someone else's subdomain.

Please adjust the implementation to validate for label names. Also, defining the regex at the beginning of the file as a var is more preferred as it would perform the regex complication only once when the module is loaded and not only every `Validate()` call.

psviderski (changes requested):
Thank you for the contribution, you're on the right track! Please see the comment that clarifies why we should likely restrict the validation to RFC 1123 Label Names only. I hope it makes sense

pkg/api/service.go
Tova-Rozovsky: I updated the service name validation to match RFC 1123 label names and moved the regex to a package-level variable for performance, as suggested.
Please let me know if there’s anything else to address.

psviderski: Are you a gpt bot?

Tova-Rozovsky: No, but I use gpt. I think like everybody...........

pkg/api/service.go
```diff
@@ -104,7 +105,19 @@ func (s *ServiceSpec) Validate() error {
 		return fmt.Errorf("invalid mode: %q", s.Mode)
 	}
 
-	// TODO: validate the service name is a valid DNS label.
+func (s *ServiceSpec) Validate() error {
```
psviderski: Please check that you've duplicated the `Validate` function so that there are syntax errors in the file and the code doesn't compile.

pkg/api/service.go
```diff
@@ -104,7 +105,19 @@ func (s *ServiceSpec) Validate() error {
...
+	if s.Mode != "someExpectedMode" {
```
psviderski: What is `someExpectedMode`?

pkg/api/service.go
```diff
@@ -104,7 +105,19 @@ func (s *ServiceSpec) Validate() error {
...
+		return fmt.Errorf("invalid mode: %q", s.Mode)
+	}
+
+	if s.Name != "" {
+		if len(s.Name) > 63 {
+			return fmt.Errorf("service name too long (max 63 characters): %q", s.Name)
+		}
+		if !dnsLabelRegexp.MatchString(s.Name) {
+			return fmt.Errorf("invalid service name: %q. Must comply with RFC 1123 label format", s.Name)
```
psviderski: I think a regular user shouldn't know about RFC 1123 and we can describe in simple words what a valid service name format should be

pkg/api/service.go
psviderski: It would also be great to add a few basic unit tests for `Validate` with valid and invalid service name. You can create a `service_test.go` file at `pkg/api` for that.

Please also make sure that all existing tests also pass by running at the root of the project: `go test ./...`
Please make sure you have Docker installed locally as it will also run e2e tests using a local uncloud cluster in Docker containers.

pkg/api/service.go
Tova-Rozovsky: I wrote the function and tried to include all the corrections. I wanted to run the tests to check everything, but I'm on Windows and the project uses Linux. I tried to fix this problem, but it seems like a big, system-wide change. It's probably too much for a first issue.

Hopefully, this is the last time and no further changes will be required.
Thank you for the guidance!!

pkg/api/service.go
```diff
@@ -103,8 +104,15 @@ func (s *ServiceSpec) Validate() error {
...
 	}
-
-	// TODO: validate the service name is a valid DNS label.
+	
+	if s.Name != "" {
+		if len(s.Name) > 63 {
+			return fmt.Errorf("service name too long (max 63 characters): %q", s.Name)
+		}
+		if !dnsLabelRegexp.MatchString(s.Name) {
+			return fmt.Errorf("invalid service name: %q.must be 1-63 characters, lowercase letters, numbers, and dashes only; must start and end with a letter or number", s.Name)
```
psviderski:
```suggestion
			return fmt.Errorf("invalid service name: %q. must be 1-63 characters, lowercase letters, numbers, and dashes only; must start and end with a letter or number", s.Name)
```

psviderski (approved):
Awesome work! I've run tests for you and all of them passed. Thank you for your first contribution! 👏

pkg/api/service.go
psviderski: Ah, I'm sorry I haven't really prioritised to make the project Windows friendly as I use macOS. Nevertheless, you nailed it! 💯

psviderski:
Aaand it has been released: [v0.7.2](https://github.com/psviderski/uncloud/releases/tag/v0.7.2). Congrats!

PR #62, opened 2025-05-12
rfc: Add a proposal for Compose-style source builds

tonyo:
An RFC doc for adding Compose-style deployments from local applications sources.

### Goal

Would be great to extend `uc deploy` so that it could build and deploy applications from source, for example the following Compose file:

```yaml
services:
  nginx:
    build:
      context: nginx/
      dockerfile: nginx/Dockerfile
      args:
        GIT_COMMIT: abc1234
```

My current preference would be using the option B-1 (remote build and using an internal registry for storage and distribution), but happy to discuss the alternatives and tradeoffs; please feel free to comment on the issue or inline.

After a feedback round I'd like to have a stab at implementing the proposal.

### DIFF
--- a/docs/rfc/app-builds-from-source.md
+++ b/docs/rfc/app-builds-from-source.md
@@ -1,0 +1,110 @@
+# Proposal: Application Builds from Source
+
+This proposal aims to support Compose-style builds and deployments for services from source, in addition to the existing functionality for deploying public container images. The proposed behavior should closely align with [Docker Compose](https://docs.docker.com/compose/) in terms of building and deploying local applications from source. This will ensure a smooth transition for current Compose users.
+
+### Example
+
+The `uc deploy` command should be able to build and deploy a Compose file like the following:
+
+```yaml
+services:
+  nginx:
+    build:
+      context: nginx/
+      dockerfile: nginx/Dockerfile
+      args:
+        GIT_COMMIT: abc1234
+```
+
+In this configuration, the nginx service does not specify an image to use (via the image: directive). Instead, it defines a local build context directory (nginx/) and a Dockerfile (nginx/Dockerfile) to build the image. Additionally, it includes a build argument (GIT_COMMIT) to use during the build process.
+
+The full Compose build specification is available [here](https://docs.docker.com/reference/compose-file/build/). While it is not necessary to implement the entire specification initially, supporting as many attributes as possible would enhance compatibility with Compose and provide a better user experience.
+
+To implement this feature, two major components need to be addressed, primarily concerning the lifecycle of the intermediate container images that are built and deployed:
+
+1. How and where the intermediate images are built
+2. Where the intermediate images are stored and how they are distributed
+
+Both problems are covered in the sections below.
+
+### Out of scope
+
+1. Supporting different image build systems (buildpack, nixpack, etc.)
+1. Supporting different build engines (buildkit, kaniko, buildah, etc.)
+
+## Implementation options: Building images
+
+Building images from application source typically involves a context folder and a Dockerfile. The key question is: where should the build process take place?
+
+### Option A: Local build (off-cluster)
+
+Idea: Perform the build locally on the same machine where the Compose file and application source are located.
+
+**Pros:**
+
+- Faster builds since the build context does not need to be transferred over the network.
+
+**Cons:**
+
+- Requires a local Docker engine (or equivalent), which introduces a dependency not previously needed.
+- The resulting image must be uploaded to the cluster, adding an extra step.
+
+### Option B: Remote build (on-cluster)
+
+Idea: Use a remote Docker engine to build the image. The build context is transferred over the network, and the image is built on one of the cluster nodes.
+
+**Pros:**
+
+- Keeps the client lightweight, avoiding the need for a local Docker daemon.
+- The built image is already stored on a cluster node, simplifying deployment.
+
+**Cons:**
+
+- Slower build times due to network transfer of the build context.
+- Resource-intensive builds could impact the performance of the cluster node.
+
+## Implementation options: Image storage and distribution
+
+After images are built from source, they need to be stored and made available on the node where the container will run.
+
+### Option 1: Internal registry
+
+Idea: Deploy an internal image registry (e.g., the official [distribution registry](https://hub.docker.com/_/registry)) to store build artifacts. When launching a container, the image is referenced using a URL pointing to the registry.
+
+**Pros:**
+
+- Provides a dedicated component for image storage and distribution.
+- Can act as a local cache for public images, improving performance.
+
+**Cons:**
+
+- Adds an additional component to the cluster, increasing overhead.
+- If the registry or its hosting node goes down, no new images can be built or retrieved.
+
+### Option 2: Peer-to-peer image storage
+
+Idea: Store the built image on one of the cluster nodes. When another node requires the image, its location is determined via distributed state (Corrosion), and the image is transferred over the cluster network.
+
+**Pros:**
+
+- No need for additional components.
+- Simplifies the architecture by leveraging existing cluster nodes.
+
+**Cons:**
+
+- Requires a custom implementation for image discovery and sharing.
+- Each cluster node must allocate sufficient disk space for image storage.
+
+### Option 3: External storage
+
+Idea: Upload built images to an external storage service, such as AWS S3, after the build process.
+
+**Pros:**
+
+- Avoids runtime overhead or new components inside the cluster.
+- External storage solutions are often highly reliable and scalable.
+
+**Cons:**
+
+- Likely requires authentication and integration with an external service.
+- Introduces dependency on a potentially proprietary external provider.
--- a/docs/rfc/app-builds-from-source.md
+++ b/docs/rfc/app-builds-from-source.md
@@ -80,10 +80,13 @@ Idea: Deploy an internal image registry (e.g., the official [distribution regist
 
 - Adds an additional component to the cluster, increasing overhead.
 - If the registry or its hosting node goes down, no new images can be built or retrieved.
+- Image storage is essentially duplicated (if using [filesystem storage driver](https://distribution.github.io/distribution/storage-drivers/filesystem/)) as some images will be stored in both the registry and Docker image store on machines.
+- The registry needs to be securely exposed to the internet (e.g. using basic auth -- not supported yet) or locally through port forwarding or WireGuard connection (not supported yet).
+- Images in registry need to be deleted and garbage collected, e.g. using [regclient](https://github.com/regclient/regclient) or automated policies, so that the occupied storage doesn't grow infinitely.
 
 ### Option 2: Peer-to-peer image storage
 
-Idea: Store the built image on one of the cluster nodes. When another node requires the image, its location is determined via distributed state (Corrosion), and the image is transferred over the cluster network.
+Idea: Store the built image on one or multiple cluster nodes. When another node requires the image, its location is determined via distributed state (Corrosion), and the image is transferred over the cluster network. Ultimately, each node may expose the Registry API backed by local image store so that regular push/pulls could be used to efficiently transfer only required layers.
 
 **Pros:**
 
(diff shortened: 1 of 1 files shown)


docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. How and where the intermediate images are built
+2. Where the intermediate images are stored and how they are distributed
+
+Both problems are covered in the sections below.
+
+### Out of scope
+
+1. Supporting different image build systems (buildpack, nixpack, etc.)
+1. Supporting different build engines (buildkit, kaniko, buildah, etc.)
```
psviderski: Isn't BuildKit the default Docker build engine we should use?

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+    build:
+      context: nginx/
+      dockerfile: nginx/Dockerfile
+      args:
+        GIT_COMMIT: abc1234
+```
+
+In this configuration, the nginx service does not specify an image to use (via the image: directive). Instead, it defines a local build context directory (nginx/) and a Dockerfile (nginx/Dockerfile) to build the image. Additionally, it includes a build argument (GIT_COMMIT) to use during the build process.
+
+The full Compose build specification is available [here](https://docs.docker.com/reference/compose-file/build/). While it is not necessary to implement the entire specification initially, supporting as many attributes as possible would enhance compatibility with Compose and provide a better user experience.
```
psviderski: Note, there is a Compose spec repo that defines the implementation agnostic specification: https://github.com/compose-spec/compose-spec/blob/main/build.md
Not sure if there are any discrepancies between them but would be good to check both when implementing

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Pros:**
+
+- Avoids runtime overhead or new components inside the cluster.
+- External storage solutions are often highly reliable and scalable.
+
+**Cons:**
+
+- Likely requires authentication and integration with an external service.
+- Introduces dependency on a potentially proprietary external provider.
```
psviderski, line 116: Option 4: External registry
Docker Hub, GHCR registry, AWS ECR, or custom one managed independently of Uncloud.
Or is this covered by Option 3?

This is the only option Kamal supports: https://kamal-deploy.org/docs/configuration/docker-registry/

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+**Pros:**
+
+- Keeps the client lightweight, avoiding the need for a local Docker daemon.
+- The built image is already stored on a cluster node, simplifying deployment.
+
+**Cons:**
+
+- Slower build times due to network transfer of the build context.
+- Resource-intensive builds could impact the performance of the cluster node.
+
```
psviderski, line 67: I have a feeling that both options could make sense in various circumstances.
Remote builds make it easy to start without an additional dependency but I'd most likely not like to use it on critical production clusters to not accidentally interfere with running workloads. Adding a dedicated build machine to a cluster could be a workaround for this.

Local builds could be useful, for example, when deploying from GitHub Actions. The action would build an image locally and then upload it to the cluster/registry.

It seems Kamal and Fly.io support both types:
* https://kamal-deploy.org/docs/configuration/builders/#remote
* https://fly.io/docs/flyctl/deploy/#options (see `--remote-only` , and `--local-only` options). By default fly creates a dedicated machine used exclusively for remote builds.

However, we can start with implementing only one of them. In terms of Uncloud-specific implementation details:
* A will only require to reuse the same logic `docker build` and `docker compose build` without doing anything on the uncloud machine side.
* B will require extending the Docker gRPC service with ?BuildX/Kit API endpoints to be able to send build requests (the same `docker compose build` uses) to the Docker API on the remote machine. Due to how uncloud proxies requests to machines through an SSH connection to one machine, we can't simply forward a remote Docker socket to our local machine to use the same logic the local `docker compose build` uses.

In my opinion, it would be much easier to start with the local build (A). I would probably not stress too much about the con "Requires a local Docker engine (or equivalent), which introduces a dependency not previously needed." because if a developer writes/uses a Dockerfile, they need to test it while they're writing it. How do they do it? Most likely using a local Docker.

psviderski (commented):
Nice work! I've left a few comments for the first part and will finish reviewing tomorrow

docs/rfc/app-builds-from-source.md
tonyo: Yes; guess I meant "all of them". Removed buildkit and rephrased a bit to reduce confusion

docs/rfc/app-builds-from-source.md
tonyo: Makes sense to link and refer to the general spec as the source of truth :+1: 
Updated.

docs/rfc/app-builds-from-source.md
tonyo: Agreed, "external registry" sounds more specific and overall better. Plus easier to implement :upside_down_face:

docs/rfc/app-builds-from-source.md
tonyo: Good points about CI, dedicated build machines, and likely having a local Docker.

Agree that we should start with something simple, and overall the "local build" + "external registry" seems like the simplest we can do now in terms of parts that we don't have to implement ourselves. The external registry is anyway a stepping stone towards the internal registry, and similar story with the local vs remote builds.

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+**Pros:**
+
+- Provides a dedicated component for image storage and distribution.
+- Can act as a local cache for public images, improving performance.
+
+**Cons:**
+
+- Adds an additional component to the cluster, increasing overhead.
+- If the registry or its hosting node goes down, no new images can be built or retrieved.
```
psviderski, line 84:
```suggestion
- If the registry or its hosting node goes down, no new images can be built or retrieved.
- Image storage is essentially duplicated (if using [filesystem storage driver](https://distribution.github.io/distribution/storage-drivers/filesystem/)) as some images will be stored in both the registry and Docker image store on machines.
- The registry needs to be securely exposed to the internet (e.g. using basic auth -- not supported yet) or locally through port forwarding or WireGuard connection (not supported yet).
- Images in registry need to be deleted and garbage collected, e.g. using [regclient](https://github.com/regclient/regclient) or automated policies, so that the occupied storage doesn't grow infinitely.
```

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Option 2: Peer-to-peer image storage
+
+Idea: Store the built image on one of the cluster nodes. When another node requires the image, its location is determined via distributed state (Corrosion), and the image is transferred over the cluster network.
```
psviderski:
```suggestion
Idea: Store the built image on one or multiple cluster nodes. When another node requires the image, its location is determined via distributed state (Corrosion), and the image is transferred over the cluster network. Ultimately, each node may expose the Registry API backed by local image store so that regular push/pulls could be used to efficiently transfer only required layers.
```

docs/rfc/app-builds-from-source.md
```diff
@@ -0,0 +1,110 @@
...
+
+**Pros:**
+
+- No need for additional components.
+- Simplifies the architecture by leveraging existing cluster nodes.
+
+**Cons:**
+
+- Requires a custom implementation for image discovery and sharing.
+- Each cluster node must allocate sufficient disk space for image storage.
```
psviderski, line 101:
```suggestion
- Each cluster node must allocate sufficient disk space for image storage.
- Each push/pull will require to transfer the entire image unless a more complex logic for transferring only new/missing layers or Registry API is implemented.
```

psviderski (commented):
I added a few more Cons to image storage options. It's a great write up to help decide what we should work on.
Overall, I believe we shouldn't be too opinionated about the image build and storage approach. But at the same time we should provide a reasonable default workflow for those who want as frictionless experience as possible -- one of the reasons we build uncloud.

There is always an option to build your image however you like it, push it to the external registry, and deploy from there. We're just getting out of their way and facilitating the deployment only.

To better support this use case, it would be great to provide a centralised mechanism for managing authentication data for external registries to not require the user to manually run `docker login` on each machine. But this may still be acceptable at this stage.

For other use cases, when we help build and store the image in a similar way `docker compose up [--build]` does, A and B look equally appealing to me. I've just realised that there is an option C that can be merged with A which is local or remote off-cluster Docker builder. The local Docker context or [`DOCKER_HOST`](https://docs.docker.com/reference/cli/docker/#host) env var could be configured to use a remote docker over ssh or tcp so the same `docker build` command can build the image on the remote host (some host, not a cluster machine). Depending on how the Docker API for building images work, by implementing the local build we may be able to automatically cover these remote builds as well.

A is also arguably easier to implement than B so my preference would be to start with A.

Regarding image store options, for my personal clusters I would definitely pick Option 2. I don't really want to manage a private registry for images I cannot push publicly (ghcr is free for public repos). I'm fine with transferring the full image every time to start with. There is also evidence that Kamal users are craving about getting rid of the registry requirement: https://github.com/basecamp/kamal/discussions/510. This would be a really slick user experience.

However, it seems all three options is essentially "push to a registry". The difference is only how to connect to that registry. So I guess it makes sense to start with Option 3. and then iterate on it to support self-managed internal registry and eventually the "cluster registry" (option 2).

tonyo:
>  I've just realised that there is an option C that can be merged with A which is local or remote off-cluster Docker builder.

Expanded option A :+1: 

Great to have a bunch of options; who knows, maybe in the future we'll support all of those and more :sweat_smile: Having a high degree of flexibility in how images are built and distributed could ultimately be one of those feature that would help to gain adoption with different audiences :eyes:  

Sweet, so looks like we're converging on options A (off-cluster builds) and 3 (external registries) for the initial implementation. Thanks for the comments! Any ~last words~ advice and/or things to keep in mind before I dive in the codebase?

psviderski:
Yeah, A3 sounds like a reasonable first step 👍
>Any last words advice and/or things to keep in mind before I dive in the codebase?

Let's have a chat about it as it could be more productive.
Something off the top of my head:
* When working on compose-related features, I always check the reference implementation of `docker compose`: https://github.com/docker/compose
* Same for Docker-related commands: https://github.com/docker/cli (Docker CLI), https://github.com/moby/moby (Docker engine), https://github.com/moby/moby/tree/master/client (Go client)
* `uc deploy` command: https://github.com/psviderski/uncloud/blob/main/cmd/uncloud/deploy.go
* Docker Go SDK examples: https://docs.docker.com/reference/api/engine/sdk/examples/
* Every machine serves gRPC API that contains Docker service: https://github.com/psviderski/uncloud/blob/main/internal/machine/api/pb/docker.proto. It's essentially Docker API proxied over gRPC. The API and the [client](https://github.com/psviderski/uncloud/blob/main/internal/machine/docker/client.go) aims to match the official Docker client interface so that it can potentially be used as a drop-in replacement in places where a regular local Docker client is used. At the moment, the method names could differ, e. g. `CreateContainer` instead of `ConstainerCreate`. But such discrepancies could be easily fixed. So when designing local builds and image uploads, try to use the Docker API in a standard way so that we can extend the design to be able to use any uncloud machine as a remote builder.
* This is likely not needed now but just keep in mind that you can easily extend the standard Compose spec with custom `x-` fields: https://github.com/compose-spec/compose-spec/blob/main/11-extension.md. We already introduced a custom `x-ports` field to support exposing ports with http(s) protocol and domain name: https://github.com/psviderski/uncloud/blob/main/test/e2e/fixtures/compose-basic.yaml#L9

tonyo:
Great, thanks for the hints :+1:

tonyo:
Basic support for local builds and pushing to external registries is now implemented in #68, and a separate issue is open for deploying without an external registry (#67), guess we can close this one.

psviderski:
Why don't we keep the doc in the repo for reference and as a decision record?

tonyo:
Didn't see too much value in it, given that we'll be implementing several of those, and the actual decisions were made here on the issue. Feel free to merge it if you find it useful.

PR #63, opened 2025-05-19
test: Add basic Github Actions config

tonyo:
Tests are currently failing (both in GHA and for me locally), will investigate separately.


scripts/docker/entrypoint.sh
```diff
@@ -6,7 +6,7 @@ cleanup() {
     echo "Terminating container processes..."
 
     # Terminate the main process if it has been started.
-    if [ -n "$MAIN_PID" ]; then
+    if [ -n "${MAIN_PID:-}" ]; then
```
tonyo, line 9: Since we `set -u`, this will fail if MAIN_PID is not defined (hitting it locally)

.github/workflows/go-tests.yml
```diff
@@ -0,0 +1,27 @@
...
+      - "**"
+  pull_request:
+    branches:
+      - main
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
psviderski: @cedws wrote a great post on the consequences of not pinning the action versions: https://cedwards.xyz/github-actions-supply-chain-attack/. There is no sensitive data in this workflow yet but it would be great to establish this practice from the very beginning to not redo it later.

.github/workflows/go-tests.yml
```diff
@@ -0,0 +1,27 @@
+name: Go Tests
+
+on:
+  push:
+    branches:
+      - "**"
+  pull_request:
+    branches:
+      - main
```
psviderski, line 11: What do you think about limiting test runs only when some files have been changed in Go paths, e.g. `cmd`, `internal`, `pkg`, `test`? This is to not unnecessarily run them every time we update docs or landing page. It seems we still have plenty free minutes (2,000 a month) so it's up to you.

psviderski (approved):
Thank you for automating Go and e2e tests in GHA! 🙌

psviderski:
I built multiarch `corrosion` and `ucind` images: https://github.com/psviderski/uncloud/commit/365a5fc0b54e6553d8ab2c941342067c099d6686. Before that we only supported `linux/arm64`. That's why e2e tests failed locally and on CI (amd64 arch). I suspect your local system is amd64, right?

I also pushed a few improvements for e2e tests to this branch as well.

So CI tests are now green, enjoy!

.github/workflows/go-tests.yml
tonyo: Let's do it

tonyo:
Cool, thx :+1: 
Yes, running amd64 now.

Issue #67, opened 2025-05-30
`uc deploy` to build and deploy without an external registry


psviderski:
A common development workflow is to build a Docker image from Dockerfile locally and run it on the remote machine(s). At the moment, we can only run images from external registries. It would be great to be able to skip this intermediary and push the built image directly to the cluster.

@tonyo wrote [RFC doc](https://github.com/psviderski/uncloud/pull/62) for adding Compose-style deployments from local applications sources. This approach is partially covered by the option A-2.

The user experience could look the following way:
1. A user defines `compose.yaml`:
   ```
   services:
     my-service:
       build: .
   ```
2. Runs `uc deploy` which will:
   1. Build the image using local Docker.
   2. Upload the built image to one or multiple machines in the cluster.
   3. Run the service in the cluster. On some machines where we need to create a container the image may be missing so we may need to pull it from other machines where it was uploaded to.

2ii and 2iii steps are worth exploring in more detail.

### Upload local image to the cluster

The simplest approach would be to essentially run `docker save IMAGE | docker load`. We can reuse the SSH transport from the cluster connection to create a Go Docker client connected to the remote machine. This way we can only connect to the entry point machine as we generally assume we don't have direct SSH connection to every machine. For example, this is true in a hybrid cluster when some machines are behind NAT and `uc` is executed outside of the private network, e.g. CI. We can workaround this by implementing a `Load` gRPC method in machine API that could be proxied to any machine.

In any case, the downside of this approach is we will need to transfer the entire image every time we upload even if only the last layer has changed. Furthermore, it may be tricky to implement a progress bar for data transfer.

A better approach could be to expose the machine's Docker/containerd image store as Registry API. This way we can reuse the standard `docker push` logic and get the layers deduplication and progress output for free. Spegel has already implemented such a lightweight API layer for containerd storage but only supports `pull` (read only): https://github.com/spegel-org/spegel/blob/main/pkg/registry/registry.go. We can reuse the code and implement the missing code for uploading images.

Each machine could run the registry API on its machine cluster IP `10.210.X.Y:5000`. In order to upload a local image to the remote machine, we can forward the `5000` port to localhost through the SSH transport and then `docker push` to it. We may also consider using https://github.com/regclient/regclient or https://github.com/google/go-containerregistry (crane) to copy image to the registry if there are issues with the standard `docker push`. For example, I'm pretty sure it doesn't support pushing multi-arch images. I instead us `docker build --push`.
[...]

PR #68, opened 2025-05-31
feat: Implement basic build & push capabilities for services

tonyo:
As a follow-up to #62, this PR implements the initial support for building images for services that contain the `build` section and pushing them to a specified (remote) registry. `uc deploy` is also updated to build and push the images before deploying the corresponding services to a cluster.

The builds are done on the local Docker daemon by default (meaning, on the host where `uc` commands are executed), but `DOCKER_HOST` environment variable should also be respected, meaning that users can use remote Docker-enabled hosts for building the images.

Summary of the new command semantics:
* `uc build` - builds the images for the services that have the non-empty `build` section in the Compose file
* `uc build --push` - builds the images and pushes them to the registry
* `uc deploy` - builds and deploys the services
* `uc deploy --no-build` - deploys the services without (re)building the images, i.e. simply using the (currently required) `image` attribute

TODOs:
* Add some basic tests

### DIFF
--- a/internal/cli/build.go
+++ b/internal/cli/build.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 
 	composetypes "github.com/compose-spec/compose-go/v2/types"
 	"github.com/distribution/reference"
@@ -137,13 +136,9 @@ func pushSingleServiceImage(ctx context.Context, dockerCli *dockerclient.Client,
 	if repoInfo.Index.Official {
 		registryKey = registry.IndexServer
 	}
-	dockerConfigDir := filepath.Join(os.Getenv("HOME"), ".docker")
 
-	// Load the Docker CLI config (config.json)
-	configFile, err := config.Load(dockerConfigDir)
-	if err != nil {
-		return fmt.Errorf("failed to load Docker CLI config: %w", err)
-	}
+	// Load the Docker config file with auth details, if available
+	configFile := config.LoadDefaultConfigFile(os.Stderr)
 
 	authConfig, err := configFile.GetAuthConfig(registryKey)
 	if err != nil {
@@ -158,8 +153,7 @@ func pushSingleServiceImage(ctx context.Context, dockerCli *dockerclient.Client,
 	authStr := base64.URLEncoding.EncodeToString(authJSON)
 
 	pushOptions := image.PushOptions{
-		RegistryAuth: authStr, // TODO: Handle authentication if needed
-		All:          true,    // Push all tags
+		RegistryAuth: authStr,
 	}
 	pushResponse, err := dockerCli.ImagePush(ctx, imageName, pushOptions)
 	if err != nil {


internal/cli/build.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	registryKey := repoInfo.Index.Name
+	if repoInfo.Index.Official {
+		registryKey = registry.IndexServer
+	}
+	dockerConfigDir := filepath.Join(os.Getenv("HOME"), ".docker")
+
+	// Load the Docker CLI config (config.json)
+	configFile, err := config.Load(dockerConfigDir)
```
psviderski: Did you see https://github.com/docker/cli/blob/master/cli/config/config.go#L156-L177 that handles the default `~/.docker` dir and also `DOCKER_CONFIG` override?

internal/cli/build.go
```diff
@@ -0,0 +1,203 @@
...
+
+	authJSON, err := json.Marshal(authConfig)
+	if err != nil {
+		return fmt.Errorf("failed to marshal auth config for registry %s: %w", registryKey, err)
+	}
+
+	authStr := base64.URLEncoding.EncodeToString(authJSON)
+
+	pushOptions := image.PushOptions{
+		RegistryAuth: authStr, // TODO: Handle authentication if needed
```
psviderski: Is this TODO still relevant?

internal/cli/build.go
```diff
@@ -0,0 +1,203 @@
...
+		All:          true,    // Push all tags
```
psviderski: Not sure I understand why we need to push all tags if we only build one specified in `service.image` in compose. Is there anything I'm missing?

psviderski (approved):
This is sick! ❤️  Please see a few minor comments but it's good to merge as is and iterate on it later.

Regarding the build approach, does it use the legacy builder, not the currently default buildkit? I suspected this because I saw `Step 1/2 : ...` in the builder output instead of `=> => ` which buildkit uses. I then tried running `DOCKER_BUILDKIT=0 docker build .` and indeed the output contained `Step ...` and the following warning:
```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            BuildKit is currently disabled; enable it by removing the DOCKER_BUILDKIT=0
            environment-variable.
```
Furthermore, when you run `docker compose build` it now prints:
```
Compose can now delegate builds to bake for better performance.
 To do so, set COMPOSE_BAKE=true.
```
Do we have 3 ways to build a docker image in Docker now? Why is the docker build such a mess? 😅 Apparently, they're pushing [Bake](https://docs.docker.com/build/bake/) to become the new default. They say the Bake's YAML format is an extension of a Docker Compose file. Maybe we should look into that as well.

I've just played with `docker compose up` command and your test compose files and the default compose behaviour slightly annoys me as it most of the time prioritises the built image and you have to manually pass `--build` to rebuild it.

Ignoring the Docker Compose behaviour, I can think of three main use cases for Uncloud `uc deploy`:

1. `build` defined, `image` NOT defined
Build an image every time `uc deploy` is run. The name of the image doesn't matter for me as a user. I just want my code to be built and run in my cluster. An image and the way to push it to the cluster machines is an implementation detail.

    E.g. it could be tagged as `uncloud.internal/<service-name>:<git-sha>` and transferred using `docker save | docker load` or any other more efficient way to the machines where it can be used to run a service.

2. `image` defined, `build` NOT defined
Deploy an existing image from the registry (public or private), there is no build process.

3. `build` defined and `image` defined
This is the confusing one. It seems this case tries to accommodate two scenarios:
    1. Build the image and push it to the external registry
    2. Deploy the image from the external registry

    To me, it's unclear when we need to build and push and when to just run the image from the registry similar to 2. And I don't see a reason to specify an image name if we don't intend to push it to a registry.

**What do you think about separating build and deploy when using a registry image?**
For example, `build` and `image` could still be defined in the Compose file but `uc deploy` will prioritise the `image` and consider it as case 2.

In order to build and image, push it to a registry, and deploy from the registry, a user will need to run two commands instead of one:
```
uc build --push  # or even use `docker compose build` or `docker bake` if they like
uc deploy
```

I could be wrong buy I have a feeling that if the case 1. doesn't work for the user and they want to use a registry, the user knows what they're doing and likely want to control how they build and push the image. In this case, getting out of their way seems reasonable to me.

What are your thoughts? This PR is probably not the best way to discuss this so happy to move the discussion to another place if you like.

internal/cli/build.go
tonyo: I looked for something of sort, but didn't find; thx!

internal/cli/build.go
tonyo: Nope, old stuff

internal/cli/build.go
tonyo: You're right, we don't need it, time for a cleanup :+1:

tonyo:
> To me, it's unclear when we need to build and push and when to just run the image from the registry similar to 2. And I don't see a reason to specify an image name if we don't intend to push it to a registry.

My assumption is, when the user specifies the `build` section, they want to make sure that they're running the latest image built from source. If not, and they want to optimize for the deploy times, and/or build the image in a more transparent/auditable way -- they'll probably need an external build/push pipeline anyway.

Splitting build and deploy in two commands sounds a bit complicated tbh. Always doing a build+push as part of `uc deploy` seems like a step closer to an ideal one-command build/deploy path I've personally wanted for myself, if that's any signal.

What do you say we revisit this question after we implement the registry-less approach in some form https://github.com/psviderski/uncloud/issues/67? Might be easier to reason having both workflows in front of us, and after some actual dogfooding.

tonyo:
> Regarding the build approach, does it use the legacy builder, not the currently default buildkit? 

For now -- yes; I saw the buildkit code branches in docker/compose codebase, but decided not to dive there for now. Would make sense to do it in a separate iteration, hopefully won't be hard to implement when we (or someone) actually needs that.

psviderski:
>My assumption is, when the user specifies the `build` section, they want to make sure that they're running the latest image built from source. If not, and they want to optimize for the deploy times, and/or build the image in a more transparent/auditable way -- they'll probably need an external build/push pipeline anyway.

Got it. I think this is where our assumptions don't fully match. For the case when I want to have an external build/push pipeline I still need to put the build configuration somewhere and the Compose file is one of the options. You specify the context, Dockerfile path, platforms, tags and use `docker compose build --push` in an external pipeline. Then I don't want `uc deploy` to do any builds. So this is where I guess the `--no-build` flag could be handy.

>What do you say we revisit this question after we implement the registry-less approach in some form https://github.com/psviderski/uncloud/issues/67? Might be easier to reason having both workflows in front of us, and after some actual dogfooding.

Yeah, sure that's a great idea!

Feel free to merge the PR whenever you're happy with it

PR #69, opened 2025-06-08
test: Add an e2e test for build&push

tonyo:
This adds a basic test for the Compose-based build/push functionality.

To test the push functionality, we start an actual registry inside the test cluster and expose its port.   
To make this work, a `PortMap` was added to `CreateClusterOptions` and `CreateMachineOptions`.

### DIFF
--- a/cmd/uncloud/deploy.go
+++ b/cmd/uncloud/deploy.go
@@ -92,11 +92,8 @@ func runDeploy(ctx context.Context, uncli *cli.CLI, opts deployOptions) error {
 			fmt.Println("Not building services as requested.")
 		} else {
 			buildOpts := cli.BuildOptions{
-				Files:    opts.files,
-				Profiles: opts.profiles,
-				Services: opts.services,
-				Push:     true,
-				NoCache:  false,
+				Push:    true,
+				NoCache: false,
 			}
 
 			if err := cli.BuildServices(ctx, servicesToBuild, buildOpts); err != nil {
--- a/test/e2e/compose_build_test.go
+++ b/test/e2e/compose_build_test.go
@@ -9,6 +9,7 @@ import (
 
 	composecli "github.com/compose-spec/compose-go/v2/cli"
 	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/docker/docker/api/types/image"
 	dockerclient "github.com/docker/docker/client"
 	"github.com/docker/go-connections/nat"
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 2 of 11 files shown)


cmd/uncloud/deploy.go
```diff
@@ -92,11 +92,8 @@ func runDeploy(ctx context.Context, uncli *cli.CLI, opts deployOptions) error {
 			fmt.Println("Not building services as requested.")
 		} else {
 			buildOpts := cli.BuildOptions{
-				Files:    opts.files,
-				Profiles: opts.profiles,
-				Services: opts.services,
-				Push:     true,
-				NoCache:  false,
+				Push:    true,
+				NoCache: false,
```
psviderski, line 96: I found it's generally a good idea to separate options structs for CLI flags and actual methods. They often start as 1:1 copy but very soon you want to add another flag (e.g. shortcut) that doesn't make sense as a method option or you want to do flag preprocessing first before passing a valid/normalized option to the method.

Like in this case, I guess `files`, `profiles`, `services` are already handled by the project and services loading logic.

test/e2e/compose_build_test.go
```diff
@@ -0,0 +1,163 @@
...
+		)
+		require.NoError(t, err)
+		servicesToBuild := cliInternal.GetServicesThatNeedBuild(project)
+		serviceImage1 := fmt.Sprintf("127.0.0.1:%d/service-first", registryHostPort)
+		serviceImage2 := fmt.Sprintf("127.0.0.1:%d/service-second:version2", registryHostPort)
+		t.Cleanup(func() {
+			// Remove the images after the test
+			removeOptions := image.RemoveOptions{Force: true, PruneChildren: true}
+			_, err := dockerCli.ImageRemove(ctx, serviceImage1, removeOptions)
+			fmt.Errorf("failed to remove image %s on test cleanup: %w; continuing", serviceImage1, err)
```
psviderski: `assert.NoError` is likely what you wanted to use here to print an error but continue execution

psviderski (approved):
Looks really good! I like how relatively simple the test looks when using ucind cluster.

Am I right that we need to expose the registry port inside the machine container as a port on the localhost of the host system to be able to push to `127.0.0.1:PORT/...` instead of `172.17.x.y:PORT/...`? The problem with pushing to non-`127.0.0.1` is that Docker will complain that the endpoint should serve HTTPS, right? And to workaround this you need to add `"insecure-registries" : [...]` to the local Docker daemon config.

test/e2e/compose_build_test.go
tonyo: Fair

tonyo:
> Am I right that we need to expose the registry port inside the machine container as a port on the localhost of the host system to be able to push to 127.0.0.1:PORT/... instead of 172.17.x.y:PORT/...? 

Tbh, just wanted to make it possible to use/test that compose file outside testing (with both `uncloud` and `docker compose`) to test different behaviors, hence localhost.

> The problem with pushing to non-127.0.0.1 is that Docker will complain that the endpoint should serve HTTPS

TIL :)

cmd/uncloud/deploy.go
tonyo: Agree, it's not the cleanest approach right now; happy to refactor after the build approach/interface is stabilized a bit more

PR #73, opened 2025-06-16
build: Install protobuf tools to subdirectory

tonyo:
We need a consistent/pinned set of versions for the protobuf toolchain.
WIP


tonyo:
Opened #77 instead

PR #74, opened 2025-06-19
Fumadocs is a modern and better alternative to Docusaurus

iRazvan2745:


psviderski:
Hey, thank you for the attempt to convert docs to fumadocs! Why did you close it? Was it just an experiment? I'm really curious to see how they looked if you have any screenshots

iRazvan2745:
image
image
image
image
Can be improoved with some banners or other components

iRazvan2745:
reopening it

psviderski:
Overall, I like the default design of fumadocs more than docusaurus. However, the documentation of fumadocs seems to target Next.js developers as it explains things in Next.js concepts in multiple places which doesn't help me at all. I'm pretty happy with docusaurus's as it required configuring a few properties in json/ts files and then worked out of the box without requiring me to write any .ts files.
How much more DIY approach does fumadocs require in that sense?

iRazvan2745:
Pretty much 0, you only need to write the markdown files and maybe use some of their components which work inside mdx

PR #77, opened 2025-06-20
build: Install protobuf tools via mise

tonyo:
This adds and pins protobuf-related tools to Mise config, so there'll be hopefully no more version (and PB-generated files) drift.
I enabled experimental mode for Mise to take advantage of the [lockfile](https://mise.jdx.dev/configuration/settings.html#lockfile) feature. One caveat there is that Mise will only add/update checksums for the platform where it runs (e.g. linux/amd64), meaning that after someone bumps versions in the Mise config, another person on a different platform will see a change in .mise.lock for their platform.

Also: add a GHA CI job to check if protobuf-generated files change.

Note: the pinned versions are not latest (to avoid massive changes in the generated files); I just matched the versions that I found in the headers of the generated files.


internal/machine/api/pb/machine.pb.go
tonyo, line 1: @psviderski let's check that these changes are ok

.mise.toml
```diff
@@ -1,5 +1,11 @@
+[settings]
+experimental = true
```
psviderski, line 2: Curious, what configuration requires `experimental`?

internal/machine/api/pb/machine.pb.go
psviderski: There is something low-level and cryptic there. As long as the e2e tests pass I think it should be alright

psviderski:
Looks awesome, ship it!

Issue #59, opened 2025-05-02
Ignore volumes not managed by uncloud


psviderski:
When using `uc volume`, it displays and manages all volumes from cluster machines which means it's possible to delete (accidentally or intentionally) volumes that weren't created by Uncloud.

Consider adding a label to Uncloud-created volumes and allow to list, inspect, and remove only these labeled volumes. This would be a safety measure to allow users to use Docker on the host with other tools.

PR #79, opened 2025-06-22
feat: Label new volumes as managed by uncloud

tonyo:
This is a first step to address https://github.com/psviderski/uncloud/issues/59; we'll be labeling all new volumes with "uncloud.managed" label (similar to how uncloud containers are now labeled).

Haven't updated ListVolumes and RemoveVolume methods yet since that will break the existing installations and needs some sort of upgrade path. A simple one could be: updating the deploy procedure to (force-)label any volume that is used by any uncloud service for now, lmk what you think @psviderski

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -345,6 +345,12 @@ func (s *Server) CreateVolume(ctx context.Context, req *pb.CreateVolumeRequest) 
 		return nil, status.Errorf(codes.InvalidArgument, "unmarshal options: %v", err)
 	}
 
+	// Always add the managed label to the volume to indicate that it is managed by Uncloud.
+	if opts.Labels == nil {
+		opts.Labels = make(map[string]string)
+	}
+	opts.Labels[api.LabelManaged] = ""
+
 	vol, err := s.client.VolumeCreate(ctx, opts)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
--- a/test/e2e/compose_deploy_test.go
+++ b/test/e2e/compose_deploy_test.go
@@ -121,6 +121,9 @@ func TestComposeDeployment(t *testing.T) {
 		})
 		require.NoError(t, err)
 		assert.Len(t, volumes, 2, "Expected 2 volumes to be created")
+		// Check that the volumes have the correct labels
+		assert.Equal(t, map[string]string{"uncloud.managed": ""}, volumes[0].Volume.Labels)
+		assert.Equal(t, map[string]string{"uncloud.managed": ""}, volumes[1].Volume.Labels)
 
 		data1Volume, data2Volume := volumes[0], volumes[1]
 		if volumes[0].Volume.Name == "test-compose-volumes-data2" {


psviderski (approved):
👍 
On the general note, do you think it'd be a good idea to change other Docker methods to allow uncloud to work with uncloud-managed resources exclusively? For example, to create/list/remove containers that are only labelled with `uncloud.managed`? And to not allow to inspect/remove a container that hasn't been created by uncloud.

tonyo:
Definitely a yes for labeling resources that uncloud creates, at least for informational purposes. 
To let list/remove only the labeled ones -- I'd also say yes, at least by default. If someone brings us a valid use case for managing non-uncloud resources with uncloud -- we could perhaps add a flag to allow that.

tonyo:
The tests will be fixed when the new ucind is pushed from main.

psviderski:
>A simple one could be: updating the deploy procedure to (force-)label any volume that is used by any uncloud service for now, lmk what you think @psviderski

Sorry I missed this one. There is one annoying problem -- Docker resources are immutable. In order to add a label to a volume/container you need to recreate it. This really sucks in this case, especially with volumes which is much more painful to recreate than containers.

It seems we can try hack labels for existing volumes by updating records in the corresponding boltdb: `/var/lib/docker/volumes/metadata.db`. However, I'm not sure if it's not locked from external writes when the docker daemon is running. Another problem is how to distinguish which existing volumes are uncloud.managed and which are not.

For containers, I think we can flip the switch to filter only `uncloud.managed` containers for all operations anytime because they have been labelled from the very beginning. However, you can see there is another set of RPC methods for managing containers with `ServiceContainer` suffix: https://github.com/psviderski/uncloud/blob/cf7e333579561316a8f41cca9d378c1e33e5c2aa/internal/machine/api/pb/docker.proto#L28-L31

These are the methods that include label filters already: https://github.com/psviderski/uncloud/blob/cf7e333579561316a8f41cca9d378c1e33e5c2aa/internal/machine/docker/server.go#L726-L728.

For the context, currently uncloud allows to create containers only as part of a service. So a **service** is the main compute primitive that consists of **containers**. A standalone container could not exist without a service. Also, a service is nothing else but just `uncloud.service.id` and `uncloud.service.name` labels assigned to containers. In order to get a service, we list all containers labeled with the given id or name: https://github.com/psviderski/uncloud/blob/cf7e333579561316a8f41cca9d378c1e33e5c2aa/pkg/client/service.go#L112-L117. In order to delete a service, we need to delete all its containers. There is no service entity in any database. Containers and labels on them is the source of truth.

You could also see `InspectServiceFromStore`: https://github.com/psviderski/uncloud/blob/fcf164dda1b003a2993506b1a704f8ad04b9e399/pkg/client/service.go#L197-L200
This does a similar container filtering but using the information about containers stored in Corrosion. I think we don't use it anywhere because it was frustratingly hard to implement reliable operations when working with eventually consistent data in Corrosion.

I still don't think we need to get rid of the non-Service set of methods:
https://github.com/psviderski/uncloud/blob/13dea92311f0e102deb456353bf1bfd6c5b24890/internal/machine/api/pb/docker.proto#L11-L16
I can think of other non-service containers we may want to run. For example, some kind of a temporary hook container that runs a DB migrations. It could be linked to a service but not necessarily use the same service spec and be included in the list of service containers.

To summarise, I think it's good to have a general purpose Docker container/volume management methods. But even for them, it seems it shouldn't hurt to add an `uncloud.managed` filter on the server side to reduce the API surface of uncloud and not accidentally interfere with other non-uncloud-managed resources.
The concern about immutable volumes described above is still valid.

psviderski:
>The tests will be fixed when the new ucind is pushed from main.

I rebuilt and pushed a new `ucind` image from main. Reran the tests and they're green now: https://github.com/psviderski/uncloud/actions/runs/15833968110

tonyo:
> To summarise, I think it's good to have a general purpose Docker container/volume management methods. But even for them, it seems it shouldn't hurt to add an uncloud.managed filter on the server side to reduce the API surface of uncloud and not accidentally interfere with other non-uncloud-managed resources.

Sounds good to me.

> The concern about immutable volumes described above is still valid.

Ah well. We could perhaps provide a migration script that creates a temporary volume, copies the data, and then recreates the original volume with the right labels. Manual intervention, but time to take advantage of the alpha status ;)

Issue #49, opened 2025-04-25
Set a reasonable logging driver default for a cluster


psviderski:
The current Docker default for logging driver is `json-file` without a limit on the size of the file. This is obviously not great as actively logging containers may fill up all the available disk space if the log driver is not explicitly configured with the limit.
Come up with a way to address this issue. Possible options:
1. Auto-configure the default in Docker daemon config.
2. Store a default for the cluster and patch service specs with it when creating services. Modifying the spec on the fly sounds more like k8s style which is not very obvious. I found Docker Swarm even has proto structure for doing this: https://github.com/moby/swarmkit/blob/447c51903463148a919770d631b0fc67839f68b3/api/types.proto#L851-L859 but I'm not sure if this has actually been implemented and available in CLI.
3. Implement own logs aggregator on each host with rotated storage and configure it as the default using 1. or 2. option. This is a rather the next step for improving logs collection and persistence (including deleted containers). IIUC Flynn had something similar: https://github.com/flynn/flynn/blob/master/docs/content/production.html.md#retrieving-logs

psviderski:
Just discovered that the `local` driver could safely be used by default: https://x.com/psviderski/status/1919550341986845058

PR #83, opened 2025-06-26
feat: set default log driver for service containers to 'local'

zasdaym:
Fix https://github.com/psviderski/uncloud/issues/49

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -203,7 +203,8 @@ func (s *ContainerSpec) SetDefaults() ContainerSpec {
 	spec := s.Clone()
 	if spec.LogDriver == nil {
 		spec.LogDriver = &LogDriver{
-			Name: "local",
+			Name:    "local",
+			Options: map[string]string{},
 		}
 	}
 	if spec.PullPolicy == "" {
--- a/test/e2e/assert.go
+++ b/test/e2e/assert.go
@@ -62,7 +62,6 @@ func assertContainerMatchesSpec(t *testing.T, ctr api.ServiceContainer, spec api
 	assert.Equal(t, spec.Container.Init, ctr.HostConfig.Init)
 	assert.True(t, strings.HasPrefix(ctr.Name, spec.Name+"-"))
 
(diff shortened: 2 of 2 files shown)


pkg/client/caddy.go
```diff
@@ -37,6 +37,9 @@ func (cli *Client) NewCaddyDeployment(image string, placement api.Placement) (*d
 		Container: api.ContainerSpec{
 			Command: []string{"caddy", "run", "-c", "/config/caddy.json", "--watch"},
 			Image:   image,
+			LogDriver: &api.LogDriver{
+				Name: "local",
+			},
```
psviderski: What do you think about setting this `local` driver by default in `CreateServiceContainer` in the daemon?
Here:
https://github.com/psviderski/uncloud/blob/main/internal/machine/docker/server.go#L530-L535

Basically allow the user to set it explicitly in the spec if they want to. But if it's not set, then set it to `local` instead of using the default from Docker. This way we won't depend on the Docker daemon configuration at all.

scripts/install.sh
```diff
@@ -55,6 +55,7 @@ install_docker() {
 
     log "⏳ Installing Docker..."
     curl -fsSL https://get.docker.com | sh
+    echo '{"log-driver": "local"}' > /etc/docker/daemon.json
```
psviderski: Please see my other comment about setting the default for each container.

This log-driver configuration will work fine for new setups. But there could be cases when a user wants to preinstall and configure Docker on their own and only then use the server as an uncloud machine. In this case, it's probably not a great idea to mangle their config (it could be owned by a configuration management tool such as Ansible).

zasdaym:
I think it's better to set it on `ContainerSpec.SetDefaults`, so the default config is not scattered.

psviderski:
>I think it's better to set it on `ContainerSpec.SetDefaults`, so the default config is not scattered.

Ah yeah agreed, this is even better!

psviderski:
Please see the failed tests, looks like we need to fix the nil vs empty map comparison in `assertContainerMatchesSpec`
https://github.com/psviderski/uncloud/actions/runs/15901842657/job/44896081108?pr=83

Issue #35, opened 2025-03-05
Fix occasional network uncloud not found when initialising a new cluster


psviderski:
`uc machine init` occasionally fails with:
<img width="1177" alt="Image" src="https://github.com/user-attachments/assets/d3d0bef6-a4ae-46b6-a487-b514645be8ba" />

pythoninthegrass:
I've run into this too on an ubuntu 24.04.2 on-prem server. You have a workaround @psviderski?

We're in the process of setting up a public IP address but for now it'll just be an internal app.

```bash
λ uc machine init ubuntu@<REDACTED> -c customer-survey -i ~/.ssh/id_rsa
Downloading Uncloud install script: https://raw.githubusercontent.com/psviderski/uncloud/refs/heads/main/scripts/install.sh
...
Updating cluster domain records in Uncloud DNS to point to machines running caddy service...
[+] Verifying internet access to caddy service 1/1
 ✘ Machine machine-wyno (<REDACTED>) Failed to send HTTP request: Get "http://<REDACTED>.uncloud-verify": context deadline exceeded (Client.Timeout exceeded while awaiting headers) Unreachable (probably behind NAT or firewall)6.4s 

DNS records for domain '<REDACTED>.cluster.uncloud.run' could not be updated as there are no internet-reachable machines running caddy containers.

Possible solutions:
- Ensure your machines have public IP addresses
- Use --public-ip flag when adding machines to override the automatically detected IPs
- Check firewall settings on your machines
- Configure port forwarding if behind NAT
- Retry Caddy deployment with 'uc caddy deploy' after resolving connectivity issues

Your services won't be accessible from the internet until at least one machine becomes reachable. If you aren't planning to expose any services publicly, you can release the domain by running 'uc dns release'.
Error: failed to update DNS records pointing to caddy service: no internet-reachable machines running service containers
```
ran `uncloud-uninstall`, then `uc machine init ubuntu@<REDACTED> -c customer-survey -i ~/.ssh/id_rsa --no-dns` but now getting

```bash
Cluster initialised with machine 'machine-pfro' and saved as context 'customer-survey' in your local config (/Users/lance/.config/uncloud/config.yaml)
Current cluster context is now 'customer-survey'.
Waiting for the machine to be ready...

[+] Deploying service caddy 0/1
 ⠸ Container caddy-0xh0 on machine-pfro  Starting                                                                0.3s
[...]

PR #89, opened 2025-07-03
Fix issue #66: prevent race condition with Docker network creation

### LINKED ISSUE #66: Do not serve API requests in Uncloud daemon before Docker network is configured
Tests on CI occasionally fail due to a race condition: https://github.com/psviderski/uncloud/actions/runs/15207844794/job/42774987227#step:5:682
```
start container: rpc error: code = NotFound desc = Error response from daemon: network uncloud not found
```

We create an `uncloud` Docker network when starting a daemon and it seems the API server can become ready and start responding before the network is created.

OrlovEvgeny:
Fixes race condition in CI tests where containers fail to start with "network uncloud not found" error. The issue occurred because the machine API server became ready before the Docker network was fully configured, allowing container operations to
  proceed before the network infrastructure was available.

  Problem

  - Tests occasionally failed on CI with error: 
  `start container: rpc error: code = NotFound desc = Error response from daemon: network uncloud not found`
  - Race condition between machine API readiness and Docker network creation
  - The machine signals readiness when API servers start, but network controller runs asynchronously
  - Containers could be started before the "uncloud" Docker network was created

  Solution

  Implemented a network readiness synchronization mechanism:

  1. Added networkReady channel to Machine struct to track network status
  2. Modified network controller to signal when Docker network is configured
  3. Updated Docker server to check network readiness before starting containers
  4. Applied Functional Options pattern for clean, extensible API design

  Key Changes

  - internal/machine/machine.go: Added network readiness tracking and IsNetworkReady() method
  - internal/machine/network.go: Signal network readiness after Docker network creation
  - internal/machine/docker/server.go: Check network readiness in StartContainer() with Functional Options pattern

  Behavior

  - Non-cluster machines: Network considered ready immediately (no cluster network needed)
  - Cluster machines: Network readiness checked before container operations
  - Cluster joining: Network readiness reset and re-signaled for new configuration

  Testing

  - Resolves intermittent failures in TestComposeDeployment tests
  - Maintains backward compatibility with existing API
  - No impact on normal operation flow

  Related

  Closes #66

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -52,6 +52,8 @@ type Server struct {
 	internalDNSIP func() netip.Addr
 	// networkReady is a function that returns true if the Docker network is ready for containers.
 	networkReady func() bool
+	// waitForNetworkReady is a function that waits for the Docker network to be ready for containers.
+	waitForNetworkReady func(ctx context.Context) error
 }
 
 // ServerOption configures the Docker server.
@@ -61,6 +63,13 @@ type ServerOption func(*Server)
 func WithNetworkReady(networkReady func() bool) ServerOption {
 	return func(s *Server) {
 		s.networkReady = networkReady
+	}
+}
+
+// WithWaitForNetworkReady sets the network readiness wait function.
+func WithWaitForNetworkReady(waitForNetworkReady func(ctx context.Context) error) ServerOption {
+	return func(s *Server) {
+		s.waitForNetworkReady = waitForNetworkReady
 	}
 }
 
@@ -136,8 +145,12 @@ func (s *Server) InspectContainer(ctx context.Context, req *pb.InspectContainerR
 
 // StartContainer starts a container with the given ID and options.
 func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerRequest) (*emptypb.Empty, error) {
-	// Check if Docker network is ready before starting the container
-	if s.networkReady != nil && !s.networkReady() {
+	// Wait for Docker network to be ready before starting the container
+	if s.waitForNetworkReady != nil {
+		if err := s.waitForNetworkReady(ctx); err != nil {
+			return nil, status.Errorf(codes.Unavailable, "Docker network not ready: %v", err)
+		}
+	} else if s.networkReady != nil && !s.networkReady() {
 		return nil, status.Errorf(codes.Unavailable, "Docker network not ready")
 	}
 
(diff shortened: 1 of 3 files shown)


psviderski:
This looks awesome! Thank you for your contribution. I'll take a closer look tomorrow at your both of your PRs 🙏

internal/machine/network.go
```diff
@@ -225,6 +230,13 @@ func (nc *networkController) prepareAndWatchDocker(ctx context.Context) error {
 	}
 	slog.Info("Docker network configured.")
 
+	// Signal that the Docker network is ready for containers
+	select {
+	case nc.networkReady <- struct{}{}:
```
psviderski: Every `StartContainer` tries to read from this channel but we write to it only once. This is likely why many tests fail because the network never gets ready. I guess you meant to close the channel instead.

internal/machine/docker/server.go
```diff
@@ -118,6 +136,11 @@ func (s *Server) InspectContainer(ctx context.Context, req *pb.InspectContainerR
 
 // StartContainer starts a container with the given ID and options.
 func (s *Server) StartContainer(ctx context.Context, req *pb.StartContainerRequest) (*emptypb.Empty, error) {
+	// Check if Docker network is ready before starting the container
+	if s.networkReady != nil && !s.networkReady() {
+		return nil, status.Errorf(codes.Unavailable, "Docker network not ready")
```
psviderski, line 154: I'm unsure what we fixed here. Previously the error was the following:
```
start container: rpc error: code = NotFound desc = Error response from daemon: network uncloud not found
```
now if the same race condition happens, the error is:
```
start container: rpc error: code = Unavailable desc = Docker network not ready
```
which should make more sense for the end user but this doesn't resolve the original issue that the race condition can happen.

psviderski:
Please also see https://github.com/psviderski/uncloud/issues/35 which looks like a duplicate of https://github.com/psviderski/uncloud/issues/66, but happens in a different scenario we'd like to fix as well

OrlovEvgeny:
I fixed the issues from the PR review:

 - Replaced channel write nc.networkReady <- struct{}{} with channel close close(nc.networkReady). This fixes the situation where the channel is read multiple times (every StartContainer call) but written to only once, causing deadlock after the first
   read.
- Instead of immediately returning "Docker network not ready" error, added actual waiting for network readiness. Created new method WaitForNetworkReady(ctx context.Context) error in machine.go and new option WithWaitForNetworkReady in
  docker/server.go. Now StartContainer waits for network readiness with context support for timeouts.
 -  Added thread-safe synchronization with sync.RWMutex to protect networkReady channel operations, especially important during cluster reinitialization. Updated channel reset logic when joining cluster to work correctly with the new architecture.

TestComposeDeployment tests now pass consistently, race condition eliminated for both tests (issue #66) and cluster initialization (issue #35). Containers now properly wait for Docker network readiness instead of failing with error.

PR #90, opened 2025-07-04
Support x-machines placement constraints in compose files

### LINKED ISSUE #88: Support placement constraints by machine names in Compose
Support custom placements on specific machines in Compose file. For example:
```
service:
  nginx:
    deploy:
      placement:
        x-machines: ["mach1", "mach3"]
```

OrlovEvgeny:
Implements support for `x-machines` placement constraints in Docker Compose files, allowing users to specify target machines for service deployment directly in their compose configuration

  ## Changes

  - Added x-machines parsing in `pkg/client/compose/service.go`, parse `x-machines` from `deploy.placement.extensions`
  - Support both `[]string` and `[]interface{}` formats from YAML
  - Added validation for machine names (non-empty, trimmed)

  ## Usage

  ```yaml
  services:
    nginx:
      image: nginx
      deploy:
        placement:
          x-machines: ["machine-1", "machine-2"]
```

  Related issues:
  
  Closes #88
  Related to #3

### DIFF
--- a/pkg/client/compose/service.go
+++ b/pkg/client/compose/service.go
@@ -55,7 +55,6 @@ func ServiceSpecFromCompose(project *types.Project, serviceName string) (api.Ser
 		},
 		Name: serviceName,
 		Mode: api.ServiceModeReplicated,
-		// TODO: implement and map x-machines to Placement.
 	}
 
 	if ports, ok := service.Extensions[PortsExtensionKey].([]api.PortSpec); ok {
@@ -84,6 +83,15 @@ func ServiceSpecFromCompose(project *types.Project, serviceName string) (api.Ser
 			}
 		default:
 			return spec, fmt.Errorf("unsupported deploy mode: '%s'", service.Deploy.Mode)
+		}
+
+		// Parse x-machines placement constraint
+		if xMachines, ok := service.Deploy.Placement.Extensions["x-machines"]; ok {
+			if machines, err := parseXMachines(xMachines); err != nil {
+				return spec, fmt.Errorf("parse x-machines placement: %w", err)
+			} else {
+				spec.Placement.Machines = machines
+			}
 		}
 	}
 
@@ -238,4 +246,37 @@ func tmpfsVolumeSpecFromCompose(serviceVolume types.ServiceVolumeConfig) api.Vol
 	}
 
 	return spec
+}
+
+// parseXMachines parses the x-machines placement constraint from Compose deploy.placement.extensions.
+// It supports both []string and []interface{} formats that may come from YAML parsing.
+func parseXMachines(xMachines interface{}) ([]string, error) {
+	switch v := xMachines.(type) {
+	case []string:
+		return validateMachineNames(v)
+	case []interface{}:
+		var machines []string
+		for i, machine := range v {
+			if str, ok := machine.(string); ok {
+				machines = append(machines, str)
+			} else {
+				return nil, fmt.Errorf("x-machines[%d] is not a string, got %T", i, machine)
+			}
+		}
+		return validateMachineNames(machines)
+	default:
+		return nil, fmt.Errorf("x-machines must be a list of strings, got %T", xMachines)
+	}
+}
+
+// validateMachineNames validates machine names to ensure they are not empty and contain valid characters.
+func validateMachineNames(machines []string) ([]string, error) {
+	for i, machine := range machines {
+		machine = strings.TrimSpace(machine)
+		if machine == "" {
+			return nil, fmt.Errorf("x-machines[%d] cannot be empty", i)
+		}
+		machines[i] = machine
+	}
+	return machines, nil
 }
--- a/pkg/client/compose/service_test.go
+++ b/pkg/client/compose/service_test.go
@@ -6,8 +6,9 @@ import (
 	"slices"
 	"strings"
 	"testing"
-
+	
 	"github.com/compose-spec/compose-go/v2/cli"
+	"github.com/compose-spec/compose-go/v2/loader"
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/docker/docker/api/types/mount"
 	"github.com/docker/go-units"
@@ -23,7 +24,7 @@ func loadProjectFromFile(t *testing.T, filename string) *types.Project {
 	t.Helper()
 	ctx := context.Background()
 	path := filepath.Join("testdata", filename)
-
+	
 	options, err := cli.NewProjectOptions(
 		[]string{path},
 		cli.WithName(FakeProjectName),
@@ -31,18 +32,18 @@ func loadProjectFromFile(t *testing.T, filename string) *types.Project {
 		cli.WithDotEnv,
 	)
 	require.NoError(t, err)
-
+	
 	project, err := options.LoadProject(ctx)
 	require.NoError(t, err)
-
+	
 	return project
 }
 
 func TestServiceSpecFromCompose(t *testing.T) {
 	t.Parallel()
-
+	
 	initTrue := true
-
+	
 	tests := []struct {
 		name     string
 		filename string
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: pkg/client/compose/service_test.go [hunk 1]
```diff
@@ -6,8 +6,9 @@ import (
 	"slices"
 	"strings"
 	"testing"
-
+
```
tonyo: I added [`.editorconfig` ](https://editorconfig.org/) to the repo earlier today; hope your IDE supports because it'll help to get rid of these trailing whitespaces and also ensure all files have trailing new lines :raised_hands:

test/e2e/compose_deploy_test.go
```diff
@@ -254,4 +254,67 @@ func TestComposeDeployment(t *testing.T) {
...
+			for _, machine := range c.Machines {
+				if machine.ID == container.MachineID {
+					machineNames = append(machineNames, machine.Name)
+					break
+				}
+			}
+		}
+		assert.ElementsMatch(t, []string{"machine-1", "machine-2"}, machineNames,
+			"Service containers should only be on specified machines")
+	})
```
tonyo, line 403: Curious: what's the current behavior if you specify a non-existing machine in the placement constraints? 
What do you think about adding another small test to assert it?

tonyo (commented):
:muscle:

test/e2e/compose_deploy_test.go
psviderski: I'd expect a non-existent machine name would just be ignored. IIRC, the final allowed set the scheduler uses is an intersection of the current set of machines with this placement set. But would be good to double check

pkg/client/compose/service.go
```diff
@@ -85,6 +84,15 @@ func ServiceSpecFromCompose(project *types.Project, serviceName string) (api.Ser
 		default:
 			return spec, fmt.Errorf("unsupported deploy mode: '%s'", service.Deploy.Mode)
 		}
+
+		// Parse x-machines placement constraint
```
psviderski: The compose-go library provides a way to register custom extensions and parsers for them in one place. Please see how we register `x-ports` here: https://github.com/psviderski/uncloud/blob/main/pkg/client/compose/project.go#L27

I think it would be nice to have a consistent approach with handling parsing and validation of all extensions on the project level only once when the compose file is loaded.

pkg/client/compose/service.go
```diff
@@ -239,3 +247,36 @@ func tmpfsVolumeSpecFromCompose(serviceVolume types.ServiceVolumeConfig) api.Vol
...
+		var machines []string
+		for i, machine := range v {
+			if str, ok := machine.(string); ok {
+				machines = append(machines, str)
+			} else {
+				return nil, fmt.Errorf("x-machines[%d] is not a string, got %T", i, machine)
+			}
+		}
+		return validateMachineNames(machines)
+	default:
```
psviderski: Do we also want to handle a string value that is just one machine? Can be converted to `["value"]`
```
x-machines: my-machine
```

pkg/client/compose/service.go
```diff
@@ -239,3 +247,36 @@ func tmpfsVolumeSpecFromCompose(serviceVolume types.ServiceVolumeConfig) api.Vol
 
 	return spec
 }
+
+// parseXMachines parses the x-machines placement constraint from Compose deploy.placement.extensions.
+// It supports both []string and []interface{} formats that may come from YAML parsing.
+func parseXMachines(xMachines interface{}) ([]string, error) {
+	switch v := xMachines.(type) {
+	case []string:
```
psviderski: I'm just curious, does the yaml parser check that all items in the slice are of the same type, e.g. `string` and cast it to `[]string`?

pkg/client/compose/service_test.go
```diff
@@ -213,26 +214,209 @@ func TestServiceSpecFromCompose(t *testing.T) {
...
+			configDetails := types.ConfigDetails{
+				ConfigFiles: []types.ConfigFile{
+					{
+						Filename: "docker-compose.yml",
+						Content:  []byte(tt.composeYAML),
+					},
+				},
+			}
+			
+			project, err := loader.LoadWithContext(t.Context(), configDetails, func(o *loader.Options) {
```
psviderski: I think we want to use the `compose.LoadProject` to test what is used everywhere in the real scenario: https://github.com/psviderski/uncloud/blob/main/pkg/client/compose/project.go#L16
It registers our custom extension parsers. You may want to extend the function to also accept ConfigFiles with content.

test/e2e/compose_deploy_test.go
```diff
@@ -254,4 +254,67 @@ func TestComposeDeployment(t *testing.T) {
...
+			Container: api.ContainerSpec{
+				Env: map[string]string{
+					"VAR":   "value",
+					"BOOL":  "true",
+					"EMPTY": "",
+				},
+				Image: "portainer/pause:3.9",
+			},
+			Placement: api.Placement{
+				Machines: []string{"machine-1", "machine-2"},
```
psviderski: The actual placement will likely be this two machines if you don't specify any placement. So a better test would be to specify only one machine or `machine-2` and `machine-3`.

test/e2e/compose_deploy_test.go
```diff
@@ -254,4 +254,67 @@ func TestComposeDeployment(t *testing.T) {
...
+		}
+		assertServiceMatchesSpec(t, svc, expectedSpec)
+
+		// Verify that containers are only deployed on specified machines
+		// Since we only specified 2 machines in x-machines and have 2 replicas,
+		// and the cluster has 3 machines, the third machine should have no containers
+		serviceMachines := serviceMachines(svc)
+		assert.Len(t, serviceMachines.ToSlice(), 2, "Service should only be on 2 machines")
+		
+		// Verify machines match the expected machine names
```
psviderski: JFYI: you can compare by IDs knowing that `c.Machines` contains `machine-1,2,3` in order: https://github.com/psviderski/uncloud/blob/main/test/e2e/service_test.go#L238-L239

pkg/client/compose/service.go
OrlovEvgeny: Good point thanks!
I initially tried to follow the same pattern as `x-ports` like `composecli.WithExtension(MachinesExtensionKey, MachinesSource{})`, but discovered a fundamental difference in requirements that led to a different architectural approach.

Why x-machines doesn't use extension registration:
Type flexibility requirement: Unlike `x-ports` which only accepts string arrays (["8080:80", "9000:9000"]), `x-machines` needs to support multiple input formats for better usability: e.g single string: `x-machines: "machine-1"`, array: `x-machines: ["machine-1", "machine-2"]`, comma-separated: `x-machines: "machine-1,machine-2"`

When you register an extension with a specific type (e.g., MachinesSource []string), compose-go strictly enforces that type during parsing. This causes errors like `source data must be an array or slice, got
  string"` when I have provide `x-machines: "machine-1"`. `x-ports` works differently: `x-ports` registration works because it only accepts arrays in YAML - Docker Compose spec naturally expects ports as arrays, so there's no type conflict.

The solution maintains architectural consistency: like x-ports, all x-machines processing happens once during project loading in LoadProject(), also we use the same project.WithServicesTransform() approach, and parsing logic is isolated in machines.go, validation is consistent

So the key insight is that extension registration is an optimization for cases where you have a fixed input type. For x-machines, we deliberately handle raw extension data to provide flexibility while maintaining the same architectural patterns for transformation and validation

OrlovEvgeny: I think I found a work around with the interface implementation decoder interface https://github.com/compose-spec/compose-go/blob/891fce532a5117121d5620e7dad5a009c6c5433e/loader/mapstructure.go#L27

OrlovEvgeny:
PR Changes Summary
Implemented professional x-machines extension support using compose-go built-in API through DecodeMapstructure interface, ensuring consistency with existing x-ports extension.
  - Added DecodeMapstructure method to MachinesSource type for custom decoding of multiple input types
  - Registered extension through composecli.WithExtension(MachinesExtensionKey, MachinesSource{})
  - Simplified transformation logic - removed manual type handling in favor of automatic compose-go decoding

  Supported Formats:
  - Single string: x-machines: "machine-1"
  - String array: x-machines: ["machine-1", "machine-2"]
  - Comma-separated: x-machines: "machine-1,machine-2"
  - Interface arrays: support for YAML parsing []interface{}
  
  jfyi @psviderski @tonyo

psviderski (approved):
I really like how clean and consistent this looks with `DecodeMapstructure` now 👍 
Awesome work!

Issue #86, opened 2025-07-01
Rename a machine


MegaGrindStone:
Hey there, first of all, thanks for this brilliant tools!

Is renaming a machine would be a feature in this tool? I've made a typo when adding the machine, and my OCD is ticking off 😆 . For my case, it would be no problem to remove and re-add the machine, then I'm seeing #14 . 

I believe I could make a PR about this renaming feature if you agree to it.

psviderski:
Hey, thank you for trying it out and taking your time to submit this issue!

Renaming a machine will definitely be a feature, we just didn't have time to implement it yet. I'm working on #14 right now and if I remember correctly @tonyo told me he wanted to take a look at renaming and/or updating machine metadata if he has time. @tonyo did you have a chance to look into this yet?

We could have either a dedicated `machine rename` command or a more versatile `machine update` command:
```
uc machine rename CURRENT_NAME NEW_NAME
uc machine update [--name NEW_NAME] [--public-ip NEW_IP] [--label FOO] CURRENT_NAME
```

### Workaround: rename a machine

For now, as a dirty workaround you can change the machine name manually in Corrosion (distributed SQLite db storing cluster state). On any machine run:
```
sudo uncloud-corrosion --admin-path /var/lib/uncloud/corrosion/admin.sock -c /var/lib/uncloud/corrosion/config.toml query 'select info from machines where name="NAME"'
```

Copy the returned JSON and change the `name` field to the new name, then replace the entire JSON object with (you would need to escape quotes inside the JSON correctly):
```
sudo uncloud-corrosion --admin-path /var/lib/uncloud/corrosion/admin.sock -c /var/lib/uncloud/corrosion/config.toml exec 'update machines set info="UPDATED-JSON" where name="NAME"'
```
This JSON object is a json-encoded [`MachineInfo`](https://github.com/psviderski/uncloud/blob/main/internal/machine/api/pb/machine.proto#L20-L25) protobuf message.

### Workaround: remove a machine
To remove a machine you can do a similar trick. Run on any cluster machine:
```
sudo uncloud-corrosion --admin-path /var/lib/uncloud/corrosion/admin.sock -c /var/lib/uncloud/corrosion/config.toml exec 'delete from machines where name="NAME"'
```
Then SSH into the removed from the cluster machine and run `sudo uncloud-uninstall`. It will clean everything up:
```
$ sudo uncloud-uninstall
[...]

Issue #30, opened 2025-02-26
Allow to configure public IP of a machine after adding to the cluster


psviderski:
`uc init` and `uc add` provide `--public-ip` flag to manually specify a public IP address of the machine for ingress configuration.
Implement a command to update the public IP for a machine in the cluster. For example:
```
uc machine update --public-ip 1.2.3.4
or
uc machine config --public-ip 1.2.3.4
```
The `update` command might be confusing if we introduce an `upgrade` command later to upgrade uncloud and the host system.

psviderski:
Could be related to https://github.com/psviderski/uncloud/issues/86 cc @tonyo

jabr:
I think this issue was resolved with #91 and can be closed, no?

psviderski:
Yeah, you're right, thanks!

PR #91, opened 2025-07-04
impl of functionality for renaming and updating machines

OrlovEvgeny:
CLI Commands
  - `uc machine rename OLD_NAME NEW_NAME`: Machine rename command
  - `uc machine update [flags]`: Universal update command
    - `--name NEW_NAME` - change machine name
    - `--public-ip IP` - update public IP address ('none' or '' to remove the public IP)

 Both support for `--context` flag for cluster selection
  
  
  Related issues: 
  #86, #30

### DIFF
--- a/test/e2e/machine_test.go
+++ b/test/e2e/machine_test.go
@@ -30,6 +30,7 @@ func TestMachineRename(t *testing.T) {
 		// Select the second machine to rename
 		originalMachine := machines[1]
 		originalName := originalMachine.Machine.Name
+		// nolint:goconst
 		newName := "renamed-machine-1"
 
 		// Rename the machine
@@ -222,6 +223,8 @@ func TestUpdateMachine(t *testing.T) {
 		// Select a machine to update
 		targetMachine := machines[1]
 		originalName := targetMachine.Machine.Name
+		//TODO: From my point of view, the most correct option is to use exclusions.rules list for `_test\.go$` files
+		// nolint:goconst
 		newName := "updated-machine-name"
 
 		// Update the machine name using UpdateMachine directly


OrlovEvgeny:
Hey @psviderski  I've been working on the machine rename/update feat and have a question about schema migrations in uncloud

Current Situation:
Looking at [/internal/machine/store/schema.sql](https://github.com/psviderski/uncloud/blob/main/internal/machine/store/schema.sql#L33), I see we have:
 ```sql
 CREATE INDEX idx_machines_name ON machines (name);
```
I believe we should make this a unique index:
```sql
  CREATE UNIQUE INDEX idx_machines_name_unique ON machines (name);
```
Logically, we shouldn't allow two machines with the same name in a cluster. This would prevent naming conflicts at the database level. And additionaly It would make the [machine Inspect](https://github.com/psviderski/uncloud/blob/main/pkg/client/machine.go#L18-L22) more reliable


 So, my question is:
 Have you already thought about how to handle SQLite schema migrations when users upgrade uncloud versions?
 
Migration Challenges I'm Considering:
1. Existing duplicate names: Current installations might already have machines with duplicate names
2. Safe migration path: How to upgrade without breaking existing deployments
3. Version compatibility: Ensuring smooth upgrades for production users

Potential Solutions:
- Schema versioning system
- Migration scripts to handle duplicates before applying UNIQUE constraint
- A "doctor" utility to fix conflicts (uc doctor --fix-duplicate-names) whatever 


For now, I've simplified the UpdateMachine logic to be atomic without uniqueness checks, but I think the database-level constraint would be the right long-term solution
But maybe you already have some ideas, and I could try to implement them in the current release?

psviderski:
>I believe we should make this a unique index:

> CREATE UNIQUE INDEX idx_machines_name_unique ON machines (name);

>Logically, we shouldn't allow two machines with the same name in a cluster. This would prevent naming conflicts at the database level. And additionaly It would make the [machine Inspect](https://github.com/psviderski/uncloud/blob/main/pkg/client/machine.go#L18-L22) more reliable

I hear you and I fully agree with you but we're dialing with a distributed CRDT-based SQLite database here managed by [Corrosion](https://github.com/superfly/corrosion). It has a lot of limitations and requires special handling of data, hence the overall complexity of the methods that deal with this distributed store. For example no unique indexes are allowed: https://superfly.github.io/corrosion/schema.html#constraints. I'd advice you to read these docs to better understand the database model and its behaviour that substantially differs from what we expect of a traditional db.

Regarding the uniqueness constraints, at the moment we handle it in many places very optimistically. For example, if there are two clients trying to create a service with the same name concurrently, there is a high chance they will succeed. Same for adding new machines with the same name. That's why we in addition to names we also have UUIDs for machines and services to distinguish them if a conflict occurred.

I think this is okay for now as long as we properly document such peculiarities. We're still at a quite early stage and trying to find out if the overall architecture and distributed approach are viable and what market we're targeting. 

I'm considering introducing distributed locks at some point to address these issues (for example by implementing [redlock](https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/) without redis or using something like [minio/dsync](https://blog.min.io/minio-dsync-a-distributed-locking-and-syncing-package-for-go/)). But again this will introduce more unnecessary complexity at this stage and will slow down our progress on developing core capabilities and validating the viability of our approach in various scenarios.

>So, my question is:
>Have you already thought about how to handle SQLite schema migrations when users upgrade uncloud versions?

I have and I asked Fly.io about their recommendation: https://github.com/superfly/corrosion/issues/261. This is really non-trivial as we need **ALL** machines in the cluster to be migrated to the new schema before we can use it on **ANY** machine. I haven't implemented anything yet but I guess we'd eventually need to design a proper migration system for this.

We didn't need to migrate anything for a while so didn't have to solve this problem. If we need to now, then let's try to think of the simplest possible approach for now. Instead of creating a complex system for distributing migrations, we can start guarding new features that require a new schema with some checks that verify the oldest schema version in the cluster. If there is a machine with an old schema, prompt the user to upgrade it and fail the new feature.

Note that there is no daemon upgrade command in CLI yet. I've been updating `uncloudd` on machines by simply downloading a new version from github releases to `/usr/local/bin/` and restarting `uncloud.service`. This is something we also need to implement. You can pick this up as well if interested (I don't think we created an issue for this).

I'm happy to chat on this topic further and share more thoughts if you like.

OrlovEvgeny:
@psviderski Thank you for the very detailed explanation! It became clearer

Makefile
```diff
@@ -46,6 +46,11 @@ proto:
 	protoc --go_out=. --go_opt=paths=source_relative --go-grpc_out=. --go-grpc_opt=paths=source_relative \
 		--proto_path=. --proto_path=internal/machine/api/vendor internal/machine/api/pb/*.proto
 
+.PHONY: proto-mise
+proto-mise:
```
psviderski, line 50: I wonder, did you set up mise in a way that `make proto` doesn't use the `protoc` that mise installs?

cmd/uncloud/machine/set.go
```diff
@@ -0,0 +1,123 @@
...
+type setOptions struct {
+	name     string
+	publicIP string
+	context  string
+}
+
+func NewSetCommand() *cobra.Command {
+	opts := setOptions{}
+	cmd := &cobra.Command{
+		Use:   "set",
```
psviderski: Naming is hard and I don't have a very strong opinion about this command but overall I'm trying to make uncloud UX to be more familiar to Docker users rather than k8s. Arguably docker has better CLI UX than kubectl.
Docker uses update in `docker container update` and `docker service update`, also Fly uses update in `fly machine update` for a similar purpose.
Maybe we should add both `set` and `update` as aliases, wdyt?

cmd/uncloud/machine/set.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+	}
+
+	cmd.Flags().StringVar(
+		&opts.name, "name", "",
+		"New name for the machine",
+	)
+	cmd.Flags().StringVar(
+		&opts.publicIP, "public-ip", "",
+		"Public IP address of the machine for ingress configuration",
```
psviderski: Do we want to accept `''` or `none` to remove the public IP in a similar way `init` and `add` allow?

cmd/uncloud/machine/set.go
```diff
@@ -0,0 +1,123 @@
...
+		newIP := "none"
+		if updatedMachine.PublicIp != nil {
+			if addr, err := updatedMachine.PublicIp.ToAddr(); err == nil {
+				newIP = addr.String()
+			}
+		}
+		changes = append(changes, fmt.Sprintf("public IP: %s -> %s", oldIP, newIP))
+	}
+
+	fmt.Printf("Machine %q (ID: %s) settings updated:\n", updatedMachine.Name, updatedMachine.Id)
```
psviderski: nit: to make it consistent with the command description
```suggestion
	fmt.Printf("Machine %q (ID: %s) configuration updated:\n", updatedMachine.Name, updatedMachine.Id)
```

internal/machine/api/pb/cluster.proto
```diff
@@ -76,3 +77,17 @@ message DNSRecord {
...
 }
+
+message SetMachineRequest {
+  // Machine to update
+  string machine_id = 1;
+  
+  // Updated machine information
+  optional string name = 2;
+  optional IP public_ip = 3;
+  optional NetworkConfig network = 4;
```
psviderski: I'm not sure if we want to support network config updates, at least at this stage, as many things need to be updated in the cluster to make it work correctly. If we want to support only endpoint updates, then lets specify only endpoints property here instead of the entire `NetworkConfig`.

For the context, IIRC, endpoints are currently not updated automatically but my idea was that each machine will periodically update its own endpoints in `MachineInfo` record in the db with its automatically discovered public and local network IPs. Maybe we should eventually support two different endpoint lists:
1. Automatically discovered by the machine and updated on schedule
2. Manually specified by the user

Then they are merged in one list which is used by the `WireGuardNetwork` to create WG peer configs.

But until we have the auto-update logic, I think it could be beneficial to allow a user to update the endpoints manually.

internal/machine/cluster/cluster.go
```diff
@@ -197,6 +198,89 @@ func (c *Cluster) AddMachine(ctx context.Context, req *pb.AddMachineRequest) (*p
...
+			return nil, err
+		}
+		// Only update specific network fields, preserving others
+		if req.Network.Endpoints != nil {
+			updatedMachine.Network.Endpoints = req.Network.Endpoints
+		}
+		// Note: We don't update subnet, management_ip, or public_key as these are critical cluster identifiers
+	}
+
+	// Update the machine in the store (atomic operation)
```
psviderski: It's not really an atomic operation if we consider all instances of the replicated db, so it's better to not confuse the reader
```suggestion
	// Update the machine in the store
```

internal/machine/cluster/cluster.go
```diff
@@ -197,6 +198,89 @@ func (c *Cluster) AddMachine(ctx context.Context, req *pb.AddMachineRequest) (*p
...
+			return nil, status.Error(codes.InvalidArgument, "invalid public IP")
+		}
+		updatedMachine.PublicIp = req.PublicIp
+	}
+	if req.Network != nil {
+		if err := req.Network.Validate(); err != nil {
+			return nil, err
+		}
+		// Only update specific network fields, preserving others
+		if req.Network.Endpoints != nil {
```
psviderski: Please see another comment that the design would be more straightforward if we only allow to specify the endpoints field in the gRPC request. For the store itself, as it's used internally only, I think it's fine to continue using `MachineInfo` to not introduce a new type for the update only.

internal/machine/cluster/cluster.go
```diff
@@ -197,6 +198,89 @@ func (c *Cluster) AddMachine(ctx context.Context, req *pb.AddMachineRequest) (*p
...
+
+	// Update the machine in the store (atomic operation)
+	if err = c.store.UpdateMachine(ctx, req.MachineId, updatedMachine); err != nil {
+		if errors.Is(err, store.ErrMachineNotFound) {
+			return nil, status.Errorf(codes.NotFound, "machine not found: %s", req.MachineId)
+		}
+		return nil, status.Errorf(codes.Internal, "update machine: %v", err)
+	}
+
+	slog.Info("Machine settings updated in the cluster.",
```
psviderski: nit to keep terminology consistent
```suggestion
	slog.Info("Machine configuration updated in the cluster.",
```

internal/machine/store/store.go
```diff
@@ -66,6 +67,79 @@ func (s *Store) CreateMachine(ctx context.Context, m *pb.MachineInfo) error {
 	return nil
 }
 
+func (s *Store) UpdateMachine(ctx context.Context, machineID string, m *pb.MachineInfo) error {
```
psviderski: Do we need an extra `machineID string` parameter?
If we update the entire `MachineInfo`, we should use `m.Id` to guarantee data consistency. Thus `machineID` seems redundant.

internal/machine/store/store.go
```diff
@@ -66,6 +67,79 @@ func (s *Store) CreateMachine(ctx context.Context, m *pb.MachineInfo) error {
...
+	}
+
+	result, err := s.corro.ExecContext(ctx, "UPDATE machines SET info = ? WHERE id = ?", string(mJSON), machineID)
+	if err != nil {
+		return fmt.Errorf("update machine: %w", err)
+	}
+
+	// Check if machine exists
+	if result.RowsAffected == 0 {
+		return fmt.Errorf("%w: %s", ErrMachineNotFound, machineID)
```
psviderski: Am I right that such wrapped error is correctly detected with `errors.Is(err, store.ErrMachineNotFound)`?

pkg/client/machine.go
```diff
@@ -46,6 +46,32 @@ func (cli *Client) ListMachines(ctx context.Context, filter *api.MachineFilter)
 	return machines, nil
 }
 
+// SetMachine sets machine configuration in the cluster.
+func (cli *Client) SetMachine(ctx context.Context, req *pb.SetMachineRequest) (*pb.MachineInfo, error) {
+	resp, err := cli.ClusterClient.SetMachine(ctx, req)
+	if err != nil {
+		return nil, err
```
psviderski, line 58: Let's catch gRPC's `codes.NotFound` and return `api.ErrNotFound` instead as this is the default not found error we return at the client layer.

psviderski (commented):
Amazing work! ❤️ Overall looks good but I left a bunch of comments to address minor issues. Feel free to let me know if something doesn't make sense

Makefile
OrlovEvgeny: In my case `make proto` calls my latest version of `protoc` which is installed on my machine, I use this latest for my other projects
direct call via `mise exec -- protoc` allows me to call exactly the version of protoc that is specified in the dot mise file

like this
```bash
➜  uncloud git:(feat/machine-rename-update-commands) ✗ protoc --version
libprotoc 29.3
➜  uncloud git:(feat/machine-rename-update-commands) ✗ mise exec -- protoc --version
libprotoc 27.3
```

I added the `proto-mise` command to the Makefile to have this shortcut for generating code

cmd/uncloud/machine/set.go
OrlovEvgeny: Yes, I'll probably just rename the command to `update`, but from my point of view without alias `set` would be beter. Duplicate interface, more entry points to the same action --> need to maintain documentation and tooltips for both..etc. Potential confusion: which alias is "correct"?

cmd/uncloud/machine/set.go
OrlovEvgeny: sounds good to me

internal/machine/store/store.go
OrlovEvgeny: yep its works, `%w` is a special marker for fmt.Errorf, and the inner `ErrMachineNotFound` will be caught in the wrapper chain, so `errors.Is` will find a match. 
`errors.Unwrap`, `errors.As` as well

https://go.dev/play/p/DKiBgT8228S

test/e2e/machine_test.go
```diff
@@ -0,0 +1,450 @@
...
+	t.Run("update machine name", func(t *testing.T) {
+		// Get initial machine state
+		machines, err := cli.ListMachines(ctx, nil)
+		require.NoError(t, err)
+		require.Len(t, machines, 3)
+
+		// Select a machine to update
+		targetMachine := machines[1]
+		originalName := targetMachine.Machine.Name
+		//TODO: From my point of view, the most correct option is to use exclusions.rules list for `_test\.go$` files
```
OrlovEvgeny, line 226: related: https://github.com/psviderski/uncloud/pull/97#issue-3245178560

Makefile
psviderski: This sounds like you didn't configure mise activation (deliberately?) https://mise.jdx.dev/getting-started.html#activate-mise. The whole idea of using mise in a project is to provide a reproducible environment for all developers of the project. In our case to make sure we use the same versions of Go, golangci-lint, and protoc tools.

I suspect that you may also run versions of Go and golangci-lint that differ from the ones specified in .mise.toml which in most cases won't cause any issues. But overall tools like mise or nix just make the setup of the local dev environment so much easier so you don't have to think about it. Just `cd` to the project `mise trust` and you have everything you need.

I think all other workflows operate under the assumption that the mise environment is activated. That's why let's not add additional Makefile targets that seem redundant.

psviderski (approved):
Awesome work! ❤️

Issue #85, opened 2025-06-29
Is uncloud.run DNS required?


reorx:
While reading the explanation of how Uncloud works under the hood, I discovered that each deployed service has a `<service-name>.<cluster-id>.cluster.uncloud.run` domain. I'm wondering if this step is required in the deployment workflow. Since Uncloud is used in a self-hosted scenario, I would prefer it to have zero external service dependencies (though having a public free registry is a nice option for small hobby projects). If this step is required, deploying a Uncloud.run identical DNS registration should be considered part of the Uncloud components.

psviderski:
The Managed DNS service providing `*.cluster.uncloud.run` domains is **optional**. It's enabled by default but you can disable it when initialising your first machine `uc machine init --no-dns`. Also, you can reserve or release this record at any time with the following commands:
```
uc dns reserve
uc dns release
```

This feature was added to provide a nice UX to new users who may not have their own DNS domain yet but would like to try out self-hosting with proper domain names and HTTPS. Also, when you add more machines to the cluster or change the deployment of the reverse proxy (Caddy), this cluster record is automatically updated to point to the public IP addresses of the machines running Caddy. So it could be a bit more convenient to use it to not manually manage IP addresses for your own domain, but just create a CNAME instead. Please see https://docs.uncloud.run/getting-started/deploy-demo-app#use-your-own-domain for details.

You can also run the Managed DNS component on your own: https://github.com/psviderski/uncloud-dns
You can provide the endpoint to it when initialising the machine `uc machine init --dns-endpoint https://...` or when reservice a cluster domain `uc dns reserve --endpoint https://...`.

Thank you @reorx for the sponsorship once again! ❤

reorx:
Understood. It's a great idea to separate this Managed DNS component as an
independent service. I remember that in Coolify and other platforms, they
are all coupled together. Thanks for the detailed explanation, and looking
forward to seeing Uncloud release a stable version!

Sincerely,
Xiao Meng

reorx:
While mentioning Caddy, I came up with another question: is it possible to
control how Caddy's config file is generated? For exmaple when deploying a
chat app, some times we need manually control the connection timeout and
file upload max size.

Sincerely,
Xiao Meng

psviderski:
>is it possible to control how Caddy's config file is generated?

[...]

PR #93, opened 2025-07-11
Add documentation regarding `--no-dns` flag

matiasgarciaisaia:
See #85

### DIFF
--- a/docs/docs/2-getting-started/2-deploy-demo-app.md
+++ b/docs/docs/2-getting-started/2-deploy-demo-app.md
@@ -271,6 +271,14 @@ Want to use your own domain, for example, `excalidraw.example.com` instead of `e
 Add a CNAME record `excalidraw.example.com` in your DNS provider (Cloudflare, Namecheap, etc.) pointing to
 `excalidraw.7za6s7.cluster.uncloud.run`. Alternatively, you can add an A record pointing to your server's IP.
 
+:::info note
+
+These instructions set up your own domain _in addition to_ uncloud's managed DNS service.
+
+If you want to avoid the managed service altogether, add `--no-dns` to your `uc machine init` command, and point an `A`-type DNS record to your server(s)'s IP(s).
+
+:::
+
 Then update the published port `80/https` in `compose.yaml` to use your domain:
 
 ```yaml title="compose.yaml"
@@ -297,8 +305,8 @@ Choose [y/N]: y
 Chose: Yes!
 
 [+] Deploying services 2/2
- ✔ Container excalidraw-0z12 on machine-dc3c  Started                       3.5s 
- ✔ Container excalidraw-azpc on machine-dc3c  Removed                       3.4s 
+ ✔ Container excalidraw-0z12 on machine-dc3c  Started                       3.5s
+ ✔ Container excalidraw-azpc on machine-dc3c  Removed                       3.4s
 ```
 
 Notice how Uncloud performed a **zero-downtime deployment** — it started the new container with the updated
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -20,7 +20,7 @@ Uncloud stores its configuration in `~/.config/uncloud/config.yaml`. If you wish
 
 ### Initialisation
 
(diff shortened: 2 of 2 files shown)


matiasgarciaisaia:
I think these are all the relevant parts of the docs. Not sure if I've overdone it a bit, though.

psviderski (approved):
Thank you so much for clarifying the docs! ❤️ 

I can see that several people have already asked questions about this DNS feature and it seems to be confusing. I'm thinking about changing the default and to not reserve the cluster domain automatically on cluster init, essentially make the `--no-dns` the default and opt-in.

If a user doesn't have their own domain, then they can reserve it via the `--managed-dns` flag for `uc machine init` or by running `uc dns reserve/release` later.

What do you think about such a change and the Uncloud DNS feature in general? I was hoping it could lower the barrier of entry for people new to self-hosting but started to doubt if it's more confusing than helpful.

matiasgarciaisaia:
I love how the managed DNS lowers the entry barrier, so I think it's a good default for now.

I'd probably change it for an stable 1.0 release, assuming the project gets traction - I think it will become a headache by itself due to Let's Encrypt per-domain certificate limits.

Also, "professional" use will probably want to be independent of your infra - so it's a must to allow to avoid it.

But I wouldn't change it now - it's great to be able to give the project a try _this easily_ 👍

I've yet to take a closer look at uncloud-dns to see how to self-host it.

psviderski:
Thank you for sharing your thoughts, yeah I agree regarding the professional use. 
>I think it will become a headache by itself due to Let's Encrypt per-domain certificate limits.

Can you please elaborate on why the limit will become a headache and in what circumstances?

>I've yet to take a closer look at uncloud-dns to see how to self-host it.

I don't really think it provides that much value if you have your own domain you can create DNS records for. The only advantage is that when you change the set of machines where the `caddy` service runs, it updates the `A` records pointing to the machines running `caddy`. I don't think you'd do this often so shouldn't be a big deal to manually update DNS.

There is another caveat that DNS load balancing (having multiple A records for the same domain name) doesn't provide any failover mechanism. If one of you machine goes down, the corresponding A record is not automatically removed. You'd need an additional mechanism to handle this. For example, use Cloudflare *proxy* CNAME record which seems to correctly handle failed destinations by switching traffic to the alive ones.

matiasgarciaisaia:
> Can you please elaborate on why the limit will become a headache and in what circumstances?

In [Rate Limits - New Certificates per Registered Domain](https://letsencrypt.org/docs/rate-limits/#limit-1):

> Up to 50 certificates can be issued per registered domain (or IPv4 address, or IPv6 /64 range) every 7 days

By the time it'll become a problem, we may want to include the service into the [Public Suffix List](https://publicsuffix.org/) - but I don't know how involved that is.

----

Having some kind of guarantee that the cluster auto-publishes a list of its ingresses' IPs on a DNS record is really handy. I love how Rancher 1.6 had the Route53 plugin that managed this. I usually set my app's records as `CNAME production.manas.tech`, and let the cluster figure out in which IP addresses are those ingress reverse-proxies running. One less thing to worry about is a good thing 👍

psviderski:
>Up to 50 certificates can be issued per registered domain

Ah, I didn't know that the registered domain is the suffix, not the FQDN. Thank you for clarifying this 🙏 

I think adding a domain to Publis Suffix List is doable. I could find the following domain there:
```
// Acorn Labs : https://acorn.io
// Submitted by Craig Jellick <domains@acorn.io>
*.on-acorn.io
```
Which is the domain [Acorn](https://github.com/acorn-io/runtime) used (the last infra project from Darren, the former Rancher CTO). uncloud-dns I use is the fork of their [`acorn-dns`](https://github.com/acorn-io/acorn-dns).
I believe this DNS feature was inspired by their Route53 plugin from Rancher 1.6.

>I usually set my app's records as CNAME production.manas.tech, and let the cluster figure out in which IP addresses are those ingress reverse-proxies running.

That's really handy, I agree! Here is the old Rancher 1.6 docs for external DNS service for reference: https://www.rancher.com/docs/rancher/v1.1/en/cattle/external-dns-service/

Issue #81, opened 2025-06-23
Add support for Compose "ports" directive


tonyo:
Currently Uncloud supports "x-ports" directive which is mainly aiming at forwarding/exposing ports to the outside via a DNS name.

The "regular" ["ports" directive](https://docs.docker.com/compose/how-tos/networking/) should also be supported, even like an alias/wrapper around "x-ports".

PR #95, opened 2025-07-11
feat: add support for standard compose ports directive

OrlovEvgeny:
The implementation works as an alias/wrapper around the existing `x-ports` functionality, allowing users to use familiar compose syntax while leveraging Uncloud's port forwarding capabilities.

 Changes
  -  Services can specify either `ports` or `x-ports`, but not both
  -  Standard compose ports are converted to x-ports string format
  -  Existing x-ports functionality remains unchanged, but mutual exclusivity

Examples:
```yaml
  services:
    web:
      image: nginx
      ports:
        - "80:8080"
        - "443:8443/tcp"

  Long Syntax with Host Mode

  services:
    api:
      image: api
      ports:
        - target: 8080
          published: 80
          protocol: tcp
          mode: host
          host_ip: "127.0.0.1"

  IPv6 Support

  services:
    service:
      image: app
      ports:
        - "[::1]:6001:6001"

  Services cannot specify both ports and x-ports directives
This will result in an error
  services:
    web:
      image: nginx
      ports:
        - "80:8080"
      x-ports:
        - "443:8443/https"

```
Related issue: #81

### DIFF
--- a/pkg/client/compose/port.go
+++ b/pkg/client/compose/port.go
@@ -2,6 +2,10 @@ package compose
 
 import (
 	"fmt"
+	"net/netip"
+	"strconv"
+	"strings"
+
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/psviderski/uncloud/pkg/api"
 )
@@ -10,20 +14,49 @@ const PortsExtensionKey = "x-ports"
 
 type PortsSource []string
 
-// TransformServicesPortsExtension transforms the ports extension of all services in the project by replacing a string
-// representation of each port with a parsed PortSpec.
+// transformServicesPortsExtension transforms both standard 'ports' and 'x-ports' to PortSpecs.
+// Standard 'ports' directive is converted to x-ports string format first, then parsed.
 func transformServicesPortsExtension(project *types.Project) (*types.Project, error) {
 	return project.WithServicesTransform(func(name string, service types.ServiceConfig) (types.ServiceConfig, error) {
-		ports, ok := service.Extensions[PortsExtensionKey].(PortsSource)
-		if !ok {
+		// Check for mutual exclusivity
+		hasStandardPorts := len(service.Ports) > 0
+		hasXPorts := service.Extensions[PortsExtensionKey] != nil
+
+		if hasStandardPorts && hasXPorts {
+			return service, fmt.Errorf("service %q cannot specify both 'ports' and 'x-ports' directives, use only one", name)
+		}
+
+		var portsSource PortsSource
+		var err error
+
+		if hasStandardPorts {
+			// Convert standard ports to x-ports string format
+			portsSource, err = convertStandardPortsToXPorts(service.Ports)
+			if err != nil {
+				return service, fmt.Errorf("convert standard ports for service %q: %w", name, err)
+			}
+		} else if hasXPorts {
+			// Use existing x-ports
+			var ok bool
+			portsSource, ok = service.Extensions[PortsExtensionKey].(PortsSource)
+			if !ok {
+				return service, nil
+			}
+		} else {
+			// No ports specified
 			return service, nil
 		}
 
-		specs, err := transformPortsExtension(ports)
+		// Parse the port strings using existing logic
+		specs, err := transformPortsExtension(portsSource)
 		if err != nil {
 			return service, err
 		}
 
+		// Ensure extensions map exists
+		if service.Extensions == nil {
+			service.Extensions = make(types.Extensions)
+		}
 		service.Extensions[PortsExtensionKey] = specs
 		return service, nil
 	})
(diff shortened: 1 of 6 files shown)


psviderski:
Just before getting into the details of the implementation, what is there a reason to go "structured ports" -> "serialised x-ports" -> "structured x-ports" and doing all the string manipulation instead of going straight "structured ports" -> "structured x-ports"?

I believe "structured ports" are already validated by `compose-go` and it seems pretty straightforward to create `api.PortSpec` from `types.ServicePortConfig` without creating the entire serialisation logic to a string in order to parse it right away to a structured representation.

Also, why do you think we shouldn't allow to specify both `ports` and `x-ports`?

OrlovEvgeny:
> Just before getting into the details of the implementation, what is there a reason to go "structured ports" -> "serialised x-ports" -> "structured x-ports" and doing all the string manipulation instead of going straight "structured ports" -> "structured x-ports"?
> 
> I believe "structured ports" are already validated by `compose-go` and it seems pretty straightforward to create `api.PortSpec` from `types.ServicePortConfig` without creating the entire serialisation logic to a string in order to parse it right away to a structured representation.

Yes, that makes sense
I've refactored 


> Also, why do you think we shouldn't allow to specify both `ports` and `x-ports`?

The main reason is to avoid ambiguity when combining these sections. In the current version, the logic is as follows: if both fields are specified, it is not obvious what the final configuration should look like, especially if there are conflicts. To avoid potential errors and unpredictable behavior, I proposed leaving only one source of truth. If in the future a real case for combining appears, this decision can always be revised.

pkg/client/compose/port.go
```diff
@@ -10,20 +14,51 @@ const PortsExtensionKey = "x-ports"
...
+			specs, err = transformPortsExtension(portsSource)
+			if err != nil {
+				return service, err
+			}
+		} else {
+			// No ports specified
+			return service, nil
 		}
 
+		// Ensure extensions map exists
```
psviderski: nit: I believe this block is redundant as we can assume `compose-go` initialises extensions maps. We also wouldn't get here because `hasXPorts := service.Extensions[PortsExtensionKey] != nil` already asserted it's initialised.

pkg/client/compose/port.go
```diff
@@ -41,3 +76,119 @@ func transformPortsExtension(ports PortsSource) ([]api.PortSpec, error) {
...
+
+		// Set published port
+		publishedPort := startPort + i
+		expandedPort.Published = strconv.FormatUint(publishedPort, 10)
+
+		// Set target port (aka container port)
+		// If target port range matches published port range, expand it too
+		// Otherwise, keep the same target port for all
+		if basePort.Target != 0 {
+			// For ranges, Docker Compose maps ports 1:1 by default
```
psviderski: From what I checked, `a-b:x` is handle the following way:
* Docker Compose: maps only one port (first available?) from `a-b` range to `x`
* Docker Swarm: maps every port from `a-b` to `x` (not `x+i`)

`a-b:x-y` is expanded by `compose-go` parser to individual entries so it seems this case is already handled.

For example, create `compose.yaml`:
```
services:
  nginx:
    image: nginx:latest
    ports:
      - 80-85:180-185
```
then run `docker compose config`:
```
name: data
services:
  nginx:
    image: nginx:latest
    networks:
      default: null
    ports:
      - mode: ingress
        target: 180
        published: "80"
        protocol: tcp
      - mode: ingress
        target: 181
        published: "81"
        protocol: tcp
      - mode: ingress
        target: 182
        published: "82"
        protocol: tcp
      - mode: ingress
        target: 183
        published: "83"
        protocol: tcp
      - mode: ingress
        target: 184
        published: "84"
        protocol: tcp
      - mode: ingress
        target: 185
        published: "85"
        protocol: tcp
networks:
  default:
    name: data_default
```

I believe we only need to handle `a-b:x` case and we shouldn't change the `Target` if we want to be compatible with compose/swarm. Can you please double-check? I couldn't find any docs clarifying the expected behaviour.

Note also that `compose-go` uses [`docker/go-connections`](https://github.com/docker/go-connections/blob/main/nat/parse.go) to parse and validate the ports: https://github.com/compose-spec/compose-go/blob/main/types/types.go#L504. So you can assume the data is valid.

pkg/client/compose/port.go
```diff
@@ -41,3 +76,119 @@ func transformPortsExtension(ports PortsSource) ([]api.PortSpec, error) {
...
+		Protocol:      port.Protocol,
+		Mode:          port.Mode,
+	}
+
+	// Apply defaults according to Docker Compose specification
+	if spec.Protocol == "" {
+		spec.Protocol = "tcp"
+	}
+	if spec.Mode == "" {
+		spec.Mode = "ingress"
```
psviderski: Oh, I just realised that the meaning of the `host` [mode](https://github.com/compose-spec/compose-spec/blob/main/05-services.md#long-syntax-3) in Compose spec differs from the ones in Uncloud. Uncloud publishes the host port only on the machines that the service runs on, *not* on all the machines.

Also, the compose default mode is [ingress](https://github.com/compose-spec/compose-go/blob/main/types/types.go#L531) and `ports` only supports `tcp` and `udp` ports. But Uncloud supports ingress (expose via the reverse proxy) only for the `http(s)` ports at the moment. This means that any `ports` using short syntax will result in unsupported service spec for uncloud which undermines the entire effort to support `ports` spec (at least for now) 🤔

pkg/client/compose/port.go
OrlovEvgeny: > I believe we only need to handle a-b:x case and we shouldn't change the Target if we want to be compatible with compose/swarm. Can you please double-check? I couldn't find any docs clarifying the expected behaviour.

Yes, long story short - you're right, I found some old issues which on the contrary state that swarm doesnt support a-b:x-y
It seems that in our case it is better to make the same behavior as swarm
display a parsing error in case of use a-b:x-y


btw just checked how it works in compose-go
https://gist.github.com/OrlovEvgeny/d7cab80672a956fb9b5c0d421316699a 

```text
compose file:
services:
  web:
    image: nginx:latest
    ports:
      - "80-82:8080-8082"

Port 1: Published="80", Target=8080, Protocol="tcp", Mode="ingress"
Port 2: Published="81", Target=8081, Protocol="tcp", Mode="ingress"
Port 3: Published="82", Target=8082, Protocol="tcp", Mode="ingress"

compose file:
services:
  web:
    image: nginx:latest
    ports:
      - "80-82:8080"

Port 1: Published="80-82", Target=8080, Protocol="tcp", Mode="ingress"
```

Well `Published="80-82"` is valid for container run,  https://github.com/docker/compose/issues/9306#issuecomment-1212185763

but we're having `PublishedPort` as `uint16` https://github.com/OrlovEvgeny/uncloud/blob/main/pkg/api/port.go#L28 

which we then cast as a string into `HostPort`
https://github.com/OrlovEvgeny/uncloud/blob/main/internal/machine/docker/server.go#L531 

Perhaps the best option would be if I didn't implement `expandPortRange` at all, since compose-go will indeed do everything for us

in my opinion this looks like a big refactoring, and I think it would be better to leave expandPortRange, but use a call to `ParsePortSpec(rawPort string) ([]PortMapping, error)` in it https://github.com/docker/go-connections/blob/main/nat/nat.go#L173
This way we will get the desired result with minimal effort.

anyway, I'll take a closer look in the next couple of days and make some fixes to this PR
Thanks for the detailed explanation!

psviderski: Oh it's such a mess!

I quite like the Docker Compose behaviour which feels intuitive to me, except the `a-b:x` one. But the behaviour follows the spec:
>It is defined as a string and can be set as a range using syntax start-end. It means the actual port is assigned a remaining available port, within the set range.

As I mentioned earlier both `docker run` and `compose` publish only one available port if the range is specified: `docker run -it --rm -p 8000-8003:80 nginx`

As our `PortSpec` doesn't support ranges, I'm happy to just not support `a-b:x` case but support `a:x` and `a-b:x-y` that are automatically handled/expanded by `compose-go`. TBH, after more than 10 years of using docker, I have never used port ranges 🤷 

>in my opinion this looks like a big refactoring, and I think it would be better to leave expandPortRange, but use a call to ParsePortSpec(rawPort string) ([]PortMapping, error) in it https://github.com/docker/go-connections/blob/main/nat/nat.go#L173
This way we will get the desired result with minimal effort.

I'm more than happy to reuse code from docker-owned libraries to minimise the amount of our own code and tests that we need to maintain.

If you're ok to not support `a-b:x` and fail if there is one for now, then it looks like we don't even need to parse anything, only map `a:x` from already parsed `ServicePortConfig` to `api.PortSpec`. It's up to you to decide how you'd like to approach this though.

psviderski (approved):
I see the reason for the last change but I think this is not what I meant by this comment: https://github.com/psviderski/uncloud/pull/95#discussion_r2207339752
Let me provide some more context on how I wanted to design the ports publishing.

The docker `-p`/`ports` always binds the port to the host. In the cluster context (uncloud) IMO binding to the host is used rarely than routing traffic through some kind of an ingress. That's why I think the default in uncloud for published ports should be through an ingress (Caddy reverse proxy at the moment).

Here is the help message for `uc run` specifying the format:
```
  -p, --publish strings     Publish a service port to make it accessible outside the cluster. Can be specified multiple times.
                            Format: [hostname:][load_balancer_port:]container_port[/protocol] or [host_ip:]:host_port:container_port[/protocol]@host
                            Supported protocols: tcp, udp, http, https (default is tcp). If a hostname for http(s) port is not specified
                            and a cluster domain is reserved, service-name.cluster-domain will be used as the hostname.
                            Examples:
                              -p 8080/https                  Publish port 8080 as HTTPS via reverse proxy with default service-name.cluster-domain hostname
                              -p app.example.com:8080/https  Publish port 8080 as HTTPS via reverse proxy with custom hostname
                              -p 53:5353/udp@host            Bind UDP port 5353 to host port 53
```

If `@host` is not specified then it means via ingress. Changing the default to `@host` based on the protocol, e.g. `8000:8000/tcp` might be confusing. What I think `8000:8000/tcp` should do is route `8000` tcp port through Caddy to 8000 tcp container port.

We have a TODO to implement L4 ingress routing for Caddy: https://github.com/psviderski/uncloud/blob/fc61e2fc07fd8ee8cf22f58f43350d567677c813/internal/machine/caddyconfig/config.go#L57-L58 but just haven't implemented it yet. We'd need to use a custom bundle for caddy because https://caddyserver.com/docs/modules/layer4 module doesn't come with the default caddy build.

Just to save round-trips I'm happy to merge this as is and make required changes in a follow-up.
Please let me know if this doesn't make sense or you have any questions.

PR #97, opened 2025-07-19
E2E tests use repeated strings for test data where constants add no value

OrlovEvgeny:
Why this change:
 E2E tests frequently use repeated string literals for test names, container names, and test data. Use constants instead of string literals  reduces test readability without improving maintainability
This is a common practice many projects exclude style linters from test directories

Technical details:
Added exclude-rules section to .golangci.yaml. Excludes goconst linter specifically for path test/e2e. No impact on production code so goconst remains active for all other directories

additionaly, I've enable goimports auto‑sorts and prunes imports, while gci enforces grouping rules together they keep every file consistently formatted and `merge‑conflict‑friendly` without manual effort.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,4 +1,10 @@
 version: "2"
+
+run:
+  concurrency: 4
+  tests: true
+  timeout: 5m
+
 linters:
   default: none
   enable:
@@ -26,5 +32,12 @@ linters:
 
 formatters:
   enable:
-    - gofmt
-    # - goimports
+    - gofumpt
+    - goimports
+    - gci
+
+issues:
+  exclude-rules:
+    - path: "test/e2e"
+      linters:
+        - goconst  # constants here add no value, so we skip goconst only for test/e2e.
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -39,5 +39,4 @@ formatters:
   enable:
     - gofumpt
     - goimports
-    - gci
 
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -29,6 +29,11 @@ linters:
     # - unparam
     # - unused
     - whitespace
+  exclusions:
+    rules:
+      - path: ^test/e2e
+        linters:
+          - goconst  # constants here add no value, so we skip goconst only for test/e2e.
 
 formatters:
   enable:
@@ -36,8 +41,3 @@ formatters:
     - goimports
     - gci
 
-issues:
-  exclude-rules:
-    - path: "test/e2e"
-      linters:
-        - goconst  # constants here add no value, so we skip goconst only for test/e2e.
--- a/Makefile
+++ b/Makefile
@@ -103,10 +103,6 @@ _lint:
 	# Uncloud daemon won't likely support OS other than Linux anytime soon, so for now we can rely on that.
 	GOOS=linux golangci-lint run $(ARGS)
 
-# .PHONY: lint-and-fix
-# lint-and-fix: lint
-# 	ARGS="--fix"
-
 .PHONY: docs-image-push
 docs-image:
 	docker buildx build --push --platform linux/amd64,linux/arm64 -t "$(DOCS_IMAGE)" ./docs
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,8 @@ $(LINT_TARGETS): _lint
 lint: ARGS=
 lint-and-fix: ARGS=--fix
 _lint:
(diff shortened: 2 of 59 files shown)


Makefile
```diff
@@ -99,7 +99,7 @@ $(LINT_TARGETS): _lint
 lint: ARGS=
 lint-and-fix: ARGS=--fix
 _lint:
-	golangci-lint run $(ARGS)
+	GOOS=linux golangci-lint run $(ARGS)
```
tonyo, line 104: Curious, why is this needed?

.golangci.yaml
```diff
@@ -23,8 +29,15 @@ linters:
...
+        linters:
+          - goconst  # constants here add no value, so we skip goconst only for test/e2e.
 
 formatters:
   enable:
-    - gofmt
-    # - goimports
+    - gofumpt
+    - goimports
+    - gci
```
tonyo: Do we need both goimports and gci? Since they both do import management, they could start conflicting, so wondering if having both adds any value.

Makefile
```diff
@@ -99,7 +99,7 @@ $(LINT_TARGETS): _lint
 lint: ARGS=
```
tonyo, line 99: Please update the "format" target above, so it also uses golangci-lint now.

tonyo (approved):
lgtm, letting @psviderski to have a look before we sell our souls to golangci-lint ;)

.golangci.yaml
OrlovEvgeny: it won't cause conflicts since lint will call them sequentially. But in general you are right, it will be enough for us to use only one, perhaps I will fix it

Makefile
OrlovEvgeny: the reasons is *_linux.go files are skipped unless the target OS is Linux
So when you run any Go tool on macOS or Windows the entire *_linux.go slice of the code base is invisible. golangci-lint relies on the same build rules
The linter loads packages via go/packages that loader honours GOOS/GOARCH. 
I don't think Uncloud will support OS other than Linux anytime soon, so for now I think we can rely on that

OrlovEvgeny: the file i'm talking about https://github.com/OrlovEvgeny/uncloud/blob/refactor/linter-config/internal/machine/network/wireguard_linux.go

Makefile
```diff
@@ -89,17 +89,17 @@ test-clean:
...
+	GOOS=linux golangci-lint fmt
 
 LINT_TARGETS := lint lint-and-fix
 .PHONY: $(LINT_TARGETS) _lint
 $(LINT_TARGETS): _lint
 lint: ARGS=
 lint-and-fix: ARGS=--fix
 _lint:
-	golangci-lint run $(ARGS)
+	GOOS=linux golangci-lint run $(ARGS)
```
psviderski, line 104:
```suggestion
	# Explicitly set OS to Linux to not skip *_linux.go files when running on macOS.
	# Uncloud daemon won't likely support OS other than Linux anytime soon, so for now we can rely on that.
	GOOS=linux golangci-lint run $(ARGS)
```

psviderski (approved):
>lgtm, letting @psviderski to have a look before we sell our souls to golangci-lint ;)

I love linters and formatters. I used to work in a project with a very aggressive configuration and it's great to enforce consistency to not have any debates about the style.

FYI, we used this golden config as a reference and then adjusted for our needs: https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322

OrlovEvgeny:
@psviderski 
It seems that everything is fine with the tests, but there is an error not related to the case, I think you just need to restart the job (I don't have permissions for this)

https://github.com/psviderski/uncloud/actions/runs/16411584531/job/46367644033?pr=97

```bash

=== NAME  TestClusterLifecycle
    cluster_test.go:46: 
        	Error Trace:	/home/runner/work/uncloud/uncloud/test/e2e/cluster_test.go:46
        	            				/home/runner/work/uncloud/uncloud/test/e2e/cluster_test.go:69
        	Error:      	Received unexpected error:
        	            	create machine 'machine-1': pull Docker image: Error response from daemon: Get "https://ghcr.io/v2/psviderski/ucind/manifests/sha256:fea05a90e3119158503df11d36b4327705d950bcb590a5e937d7eb297f41eebf": dial tcp 140.82.114.34:443: i/o timeout
        	Test:       	TestClusterLifecycle
--- FAIL: TestClusterLifecycle (30.89s)
```

Issue #12, opened 2025-02-06
Implement service logs command


psviderski:
`uc service logs` or `uc logs` — view service logs similar to `docker logs`

PR #99, opened 2025-07-20
feat: implement service logs command with colored output and strict ordering

OrlovEvgeny:
This PR implements the `uc logs` command for viewing aggregated logs from all containers of a service across the Uncloud cluster

## Part 0
  - New `uc logs <service>` command that aggregates logs from all service containers
  - Real-time streaming via gRPC from each machine in the cluster
  - Support for standard Docker log options: `--follow`, `--tail`, `--timestamps`, `--since`, `--until`

 Two Merge Models:
  - Fast mode (default): Logs are printed immediately as received
    - Minimal latency and resource usage
    - Best-effort ordering between machines
    - Suitable for common debugging
  - Strict ordering mode (`--strict-order` flag):
    - Guaranteed chronological order across all machines
    - Uses min-heap based k-way merge algorithm

### Commands
```bash
uncloud git:(feat/uncloud-logs) ./uc log --help

View logs from all replicas of a service.

The logs command retrieves and displays logs from all running replicas
of the specified service across all machines in the cluster.

Usage:
  uncloud logs SERVICE [flags]

Aliases:
  logs, log

Flags:
  -c, --context string   Name of the cluster context. (default is the current context)
  -f, --follow           Follow log output
  -h, --help             help for logs
      --since string     Show logs since timestamp or relative (42m for 42 minutes)
      --strict-order     Merge logs in strict chronological order (slower but accurate)
      --tail int         Number of lines to show from the end of the logs (default -1)
  -t, --timestamps       Show timestamps
      --until string     Show logs before a timestamp or relative (42m for 42 minutes)

Global Flags:
      --connect string          Connect to a remote cluster machine without using the Uncloud configuration file.
                                Format: [ssh://]user@host[:port] or tcp://host:port
      --uncloud-config string   Path to the Uncloud configuration file. (default "~/.config/uncloud/config.yaml")
```

### Examples
<details>
  <summary>Show details...</summary>
-  Show the last 5 lines of each replica
```bash
./uc log excalidraw --tail 5
```

> image

- Show logs for the last 6 minutes
```bash
./uc log excalidraw --since 6m
```

> image

- Show the last 10 lines of each replica, and merge logs in strict chronological order (slower but accurate)
```bash
./uc log excalidraw -t --tail 10 --strict-order
```

> image

</details>

## Part 1 WIP
Implement a persistent log storage that saves container logs to disk on each machine, enabling access to historical logs even after containers are removed

related issue [#12 ](https://github.com/psviderski/uncloud/issues/12)

### DIFF
--- a/cmd/uncloud/logs/root.go
+++ b/cmd/uncloud/logs/root.go
@@ -1,0 +1,416 @@
+package logs
+
+import (
+	"container/heap"
+	"context"
+	"fmt"
+	"hash/fnv"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/psviderski/uncloud/internal/cli"
+	"github.com/psviderski/uncloud/internal/machine/api/pb"
+	"github.com/psviderski/uncloud/internal/machine/docker"
+	"github.com/psviderski/uncloud/pkg/api"
+	"github.com/psviderski/uncloud/pkg/client"
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc/metadata"
+)
+
+func NewRootCommand() *cobra.Command {
+	var options logsOptions
+
+	cmd := &cobra.Command{
+		Use:     "logs SERVICE",
+		Aliases: []string{"log"},
+		Short:   "View service logs",
+		Long: `View logs from all replicas of a service.
+
+The logs command retrieves and displays logs from all running replicas
+of the specified service across all machines in the cluster.`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			uncli := cmd.Context().Value("cli").(*cli.CLI)
+			return streamLogs(cmd.Context(), uncli, args[0], options)
+		},
+	}
+
+	cmd.Flags().StringVarP(&options.context, "context", "c", "",
+		"Name of the cluster context. (default is the current context)")
+	cmd.Flags().BoolVarP(&options.follow, "follow", "f", false,
+		"Follow log output")
+	cmd.Flags().Int64Var(&options.tail, "tail", -1,
+		"Number of lines to show from the end of the logs")
+	cmd.Flags().BoolVarP(&options.timestamps, "timestamps", "t", false,
+		"Show timestamps")
+	cmd.Flags().StringVar(&options.since, "since", "",
+		"Show logs since timestamp or duration (42m for 42 minutes)")
+	cmd.Flags().StringVar(&options.until, "until", "",
+		"Show logs before a timestamp or duration (42m for 42 minutes)")
+	cmd.Flags().BoolVar(&options.strictOrder, "strict-order", false,
+		"Merge logs in strict chronological order (slower but accurate)")
+
+	return cmd
+}
+
+type logsOptions struct {
+	context     string
+	follow      bool
+	tail        int64
+	timestamps  bool
+	since       string
+	until       string
+	strictOrder bool
+}
+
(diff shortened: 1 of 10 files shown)


OrlovEvgeny:
Hey @tonyo  @psviderski !
As this PR is still wip, feel free to drop any suggestions for improving the logger
happy to hear your thoughts

tonyo:
Great stuff :muscle: 

Haven't looked too deep yet, but big +1 for following the docker / docker-compose semantics.

> Implement a persistent log storage that saves container logs to disk on each machine, enabling access to historical logs even after containers are removed or restarted

That might blow up the PR size, I'd suggest to leave it for a follow-up PR and focus on basic functionality and user interface here.

cmd/uncloud/logs/root.go
```diff
@@ -0,0 +1,416 @@
...
+		close(errChan)
+	}()
+
+	h := &logEntryHeap{}
+	heap.Init(h)
+
+	// Read initial entries from each channel
+	for _, ch := range machineChannels {
+		select {
+		case entry, ok := <-ch:
```
psviderski, line 226: Is there a risk that a particular machine won't have any logs for a while so we will block here?

cmd/uncloud/logs/root.go
```diff
@@ -0,0 +1,416 @@
...
+			select {
+			case entry, ok := <-item.channel:
+				if ok {
+					heap.Push(h, &logEntryWithChannel{
+						entry:   entry,
+						channel: item.channel,
+					})
+				}
+			default:
+				// Channel is truly empty, don't re-add
```
psviderski, line 274: Could it become non empty later if we're using the `--follow` mode?
This one could be tricky. I have a feeling that we might need to send empty "checkpoint" entries on a regular interval to streams when there are no container logs. They may help distinguish the cases when there are issues with communicating with the machine and there are no logs from this machine.

internal/machine/api/pb/docker.proto
```diff
@@ -195,3 +198,23 @@ message MachineServiceContainers {
...
+  int64 tail = 3; // -1 means all
+  bool timestamps = 4;
+  string since = 5; // https://www.rfc-editor.org/rfc/rfc3339.html timestamp or duration string
+  string until = 6; // https://www.rfc-editor.org/rfc/rfc3339.html timestamp or duration string
+  bool details = 7; // TODO: reserve for show extra details (container labels or smth like this)
+}
+
+message ContainerLogsResponse {
+  // Stream type: 1 = stdout, 2 = stderr
+  int32 stream_type = 1;
```
psviderski, line 215: FYI protobuf support enums: https://github.com/psviderski/uncloud/blob/2c3bea64e52915d3257e82b58fa587f5717e928b/internal/machine/api/pb/cluster.proto#L72-L76

psviderski (commented):
I like the direction!

To give you some food for thoughts, here is a talk about grpc-proxy from one of the Sidero (Talos linux) developers who forked grpc-proxy and add support for 1-Many requests, including for streaming methods. We use this project.

I wonder, if we can encapsulate some of the complexity currently implemented on the client in a server side gRPC call `ServiceLogs`. Note that this call could be broadcasted to N machines at the same time and it will return a stream that would combine all the streams from the machines. So basically all the multiplexing will be done by grpc-proxy.

What we'd need to do on the client is to reorder log entries if strict order is required. I believe this could reliably be done by using some buffer and periodic checkpoint entries when containers don't send logs. This way we can flush the next log entry once we received at least one entry from each machine.

This idea might not work though but feel free to investigate if you like.

psviderski:
Closing this as forked and finished in https://github.com/psviderski/uncloud/pull/196

PR #100, opened 2025-07-20
feat(deploy): Recreate container when pull_policy set to Always

tonyo:
Currently, when `pullPolicy` is set to `always` for a service, we don't redeploy the service on `uc deploy`.
This is problematic for e.g. situations when a floating reference (e.g. "xxx:latest") is used for `image`.

`docker compose up` respects `pullPolicy: always`: it pulls the image and recreates the container every time; I'd suggest we follow the same behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,10 +101,6 @@ lint-and-fix: ARGS=--fix
 _lint:
 	golangci-lint run $(ARGS)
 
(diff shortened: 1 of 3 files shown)


psviderski (approved):
This is so good! 👍

Issue #43, opened 2025-04-04
Add support for Compose configs


psviderski:
https://docs.docker.com/reference/compose-file/configs/

tonyo:
Some thoughts before I jump into implementation.


**Goal**: Allow users to inject non-sensitive information, such as configuration files, into running services (containers) without the need to rebuild the containers images.

The [`configs`](https://docs.docker.com/reference/compose-file/configs/) feature (not to be confused with [`docker compose config`](https://docs.docker.com/reference/cli/docker/compose/config/) command) implemented in Docker Compose and Docker Swarm seems like a decent interface to implement, which is in line with our current goal of maintaining compatibility with Compose.

### Configs in Docker

A few hard facts about “configs” feature in Docker first:

- `config` is a first-class object in `docker`, but proper `docker config` management is only available in Swarm mode ([docs](https://docs.docker.com/engine/swarm/configs/) about configs for Swarm).
    
    ```
    ✘ $ docker config 
    Usage:  docker config COMMAND
    
    Manage Swarm configs
    
    Commands:
      create      Create a config from a file or STDIN
      inspect     Display detailed information on one or more configs
      ls          List configs
      rm          Remove one or more configs
    
    Run 'docker config COMMAND --help' for more information on a command.
    
    ✘ $ docker config ls
    Error response from daemon: This node is not a swarm manager. Use "docker swarm init" or "docker swarm join" to connect this node to swarm and try again.
     
    ```
    
- It’s still possible to use the `config` feature with Compose. On standalone Docker Compose, configs are not centrally managed nor versioned as in Swarm – instead, the file (or inline content, or environment variable contents) is simply mounted when you run `docker compose up`, basically behaving like a read-only file bind mount.
    - Configs top-level field docs: https://docs.docker.com/reference/compose-file/configs/
    - Service-level field docs: https://docs.docker.com/reference/compose-file/services/#configs
- Configs content is not encrypted

### Our options

* **Option 1**: Implement “config” as a first-class entity (similar to volumes), including its own subcommands (`uc config create/inspect/ls/rm`) and gRPC methods. UX-wise, this will be similar to how Swarm lets you manage configs. When creating a config — either via CLI or via the Compose file — a separate (decoupled) config object is created in the cluster that can later be mounted by services. 

[...]

PR #116, opened 2025-09-04
feat: Initial support for Compose configs

tonyo:
Resolves https://github.com/psviderski/uncloud/issues/43.

This adds initial support for [Compose configs](https://docs.docker.com/reference/compose-file/configs/). 
Configs are basically arbitrary data objects that are copied to files inside the uncloud-managed containers on startup. 

Supported:
* `file` source: using content from a file on the local (control node) filesystem.
* `content` source: using inline content from the Compose file

Not supported:
* External configs -- and probably won't be, unless we pivot towards configs being first-class citizens as in Swarm.
* `environment` source -- should be straightforward to add in a follow-up.

TODOs for this PR:
* Adding basic docs.


tonyo:
@psviderski ready for an initial review

internal/machine/docker/server.go
```diff
@@ -672,6 +683,103 @@ func ToDockerMounts(volumes []api.VolumeSpec, mounts []api.VolumeMount) ([]mount
...
+	configMap := make(map[string]api.ConfigSpec)
+	for _, config := range configs {
+		configMap[config.Name] = config
+	}
+
+	// Process each config mount
+	for _, mount := range mounts {
+		config, exists := configMap[mount.Source]
+		if !exists {
+			return fmt.Errorf("config mount references a config that doesn't exist: '%s'", mount.Source)
```
psviderski: It feels like this should be part of spec validation. We should still perhaps handle this case in this function but it would be better to not even get to this stage and fail on the deployment planning stage.

See volumes validation for example: https://github.com/psviderski/uncloud/pull/116/files#diff-52e41636f29e1f855e51f1ad1144dd1cc18e411574f06628ddc711f2564130d5L171-L178

pkg/api/config.go
```diff
@@ -0,0 +1,51 @@
+// Implementation of Config feature from the Compose spec
+package api
+
+// ConfigSpec defines a configuration object that can be mounted into containers
+type ConfigSpec struct {
+	// Do we need this, if external is not supported yet?
```
psviderski: It seems we do need this as `ConfigMount` references this spec by this name.

pkg/api/config.go
```diff
@@ -0,0 +1,51 @@
...
+	// TODO: add support for "environment"
+}
+
+// ConfigMount defines how a config is mounted into a container
+type ConfigMount struct {
+	// Source is the name of the config
+	Source string `json:"source"`
+
+	// Target path inside the container (defaults to /<source> if not specified)
+	Target string `json:"target,omitempty"`
```
psviderski: What do you think about naming these fields in a similar style we do in `VolumeMount`? https://github.com/psviderski/uncloud/blob/b605a64fbe02d338bdd1b28d91d230e56fd444e1/pkg/api/volume.go#L197-L200

I think source and target terms are used in Docker but for our own volumes I decided to call them explicitly what they actually are.

>(defaults to /\<source> if not specified)

is this true? Please also review other comments for fields to not accidentally mislead

pkg/api/config.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// UID for the mounted config file
+	UID string `json:"uid,omitempty"`
+
+	// GID for the mounted config file
+	GID string `json:"gid,omitempty"`
+
+	// Mode (file permissions) for the mounted config file
+	Mode *uint32 `json:"mode,omitempty"`
```
psviderski: BTW, we use the standard Go CamelCase for all API structs similar to how Docker serialises its structs. Should we not change the case here for consistency?

internal/machine/docker/server.go
```diff
@@ -672,6 +683,103 @@ func ToDockerMounts(volumes []api.VolumeSpec, mounts []api.VolumeMount) ([]mount
...
+		}
+
+		// Determine target path in container
+		targetPath := mount.Target
+		if targetPath == "" {
+			targetPath = "/" + mount.Source
```
psviderski: Does this make sense if `Source` is always a name of the `ConfigSpec`?

pkg/api/config.go
psviderski: Can we use `os.FileMode` here like in `mount.TmpfsOptions`? There is no strong reason, I'm just concerned about consistency in the API.

pkg/api/config.go
```diff
@@ -0,0 +1,51 @@
+// Implementation of Config feature from the Compose spec
+package api
+
+// ConfigSpec defines a configuration object that can be mounted into containers
+type ConfigSpec struct {
+	// Do we need this, if external is not supported yet?
+	Name string `json:"name"`
+
+	// File path (when External is false)
+	File string `json:"file,omitempty"`
```
psviderski: Do you think we still need `File` in the uncloud spec that we don't use? I think we have something similar for the Caddy config: https://github.com/psviderski/uncloud/blob/dd7bc6c982be2a321ce7d14b029341ba2925ab58/pkg/api/caddy.go#L7-L9

In Compose, you can define a Caddy config either inline or from file but then our spec contains only the content.

internal/machine/docker/server.go
```diff
@@ -672,6 +683,103 @@ func ToDockerMounts(volumes []api.VolumeSpec, mounts []api.VolumeMount) ([]mount
...
+		Name:     filepath.Base(targetPath),
+		Size:     int64(len(content)),
+		Mode:     int64(fileMode),
+		ModTime:  time.Now(),
+		Typeflag: tar.TypeReg,
+	}
+
+	// Set ownership if specified
+	if uid != "" {
+		if parsedUID, parseErr := strconv.Atoi(uid); parseErr == nil {
```
psviderski: Should UID/GID be ints or uint16 in the first place (API spec) if we only accept ints here? Again, we should likely validate this when validating the `ServiceSpec` and work with only valid data here.

internal/machine/docker/server.go
```diff
@@ -672,6 +683,103 @@ func ToDockerMounts(volumes []api.VolumeSpec, mounts []api.VolumeMount) ([]mount
...
+	}
+	if err := tw.Close(); err != nil {
+		return fmt.Errorf("close tar writer: %w", err)
+	}
+
+	// Copy the tar archive to the container
+	targetDir := filepath.Dir(targetPath)
+	if targetDir == "." {
+		targetDir = "/"
+	}
```
psviderski, line 785: What is the behaviour if `targetPath` is a relative path such as `./relative/path/in/container`?

internal/machine/docker/server.go
```diff
@@ -672,6 +683,103 @@ func ToDockerMounts(volumes []api.VolumeSpec, mounts []api.VolumeMount) ([]mount
...
+
+	if err := s.client.CopyToContainer(ctx, containerID, targetDir, &buf, container.CopyToContainerOptions{}); err != nil {
```
psviderski: Do we need to set the option `CopyUIDGID` in `CopyToContainerOptions` to correctly set the uid/gid from the tar archive? or it means something else?

pkg/client/compose/config.go
```diff
@@ -0,0 +1,79 @@
...
+				// TODO: handle this in a separate function?
+				if !filepath.IsAbs(configPath) {
+					configPath = filepath.Join(workingDir, configPath)
+				}
+
+				fileContent, err := os.ReadFile(configPath)
+				if err != nil {
+					return nil, nil, fmt.Errorf("read config from file '%s': %w", spec.File, err)
+				}
+				spec.Content = string(fileContent)
```
psviderski: I've just thought, is there a valid case for a config to be a binary file? If so, I'm not sure what `string(binary non-string content)` will do with it and whether it will correctly write it later to the tarball.
Maybe we just need to use `[]byte` for `spec.Content`?

test/e2e/compose_configs_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"testing"
+
+	"github.com/psviderski/uncloud/internal/ucind"
+	"github.com/psviderski/uncloud/pkg/api"
+	"github.com/psviderski/uncloud/pkg/client/compose"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestComposeConfigs(t *testing.T) {
```
psviderski, line 15: Just a note that I usually try to add e2e tests to existing test suits to reduce the number of ucind clusters we create in parallel. I'm not sure how far we are from hitting the memory limit on GHA agents as one ucind cluster allocates non-negligible amount of ram. Also, I suspect that adding one more cluster may slow down the all tests but I haven't analysed that.

It's fine to keep this test but just keep this in mind for future work.

Ah, just saw it uses 1 machine so the overhead should be much smaller compared to 3-machine cluster.

test/e2e/compose_configs_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		project, err := compose.LoadProject(ctx, []string{"fixtures/compose-configs.yaml"})
+		require.NoError(t, err)
+
+		deploy, err := compose.NewDeployment(ctx, cli, project)
+		require.NoError(t, err)
+
+		plan, err := deploy.Plan(ctx)
+		require.NoError(t, err)
+		// Should plan to deploy 1 service (configs are parsed but not yet deployed by uncloud)
```
psviderski: what does this mean 🤔

psviderski (approved):
Awesome work! Really love how configs are copied to a created container before starting it.
I left a bunch of minor comments mainly focused on the consistency of the API. It's mainly used internally at this stage (Compose is the main user interface for now). But it would still be hard to make breaking changes in it later.

Btw, did you have any plans to add configs support to `uc run` command? For example, using a format similar to volumes: `--config ./local/path:/container/path`

pkg/api/config.go
tonyo: Since "Mode" is already camel case: what attribute do you mean exactly?

pkg/api/config.go
tonyo: Renamed :+1: 

>   (defaults to /<source> if not specified)

This is the default behavior as described by the [Compose spec](https://github.com/compose-spec/compose-spec/blob/main/08-configs.md); updated the comment but think it makes sense to keep the logic.

internal/machine/docker/server.go
tonyo: This is the default behavior as per the [Compose spec](https://docs.docker.com/reference/compose-file/configs/).

pkg/api/config.go
tonyo: The underlying Compose spec structure ([FileReferenceConfig](https://github.com/compose-spec/compose-go/blob/e16f794feeddb329ca80d4918a780dcc01705994/types/types.go#L658)) used to be uint32; I've added the conversion for now, but it won't be required after we upgrade the version of compose-spec/compose-go.

test/e2e/compose_configs_test.go
tonyo: Some LLM iteration leftovers :upside_down_face:

pkg/api/config.go
psviderski: Sorry I wasn't clear enough. I was referring to the json tags using snake? case while we don't specify any serialised JSON keys for any other `api` structs and rely on the Go JSON serialisation defaults (like Docker does).

internal/machine/docker/server.go
psviderski: Ah cool, agreed that it's good to follow the spec 👍

pkg/client/compose/config.go
tonyo: Makes sense, see no reason not to support binary configs

internal/machine/docker/server.go
tonyo: Agreed

internal/machine/docker/server.go
tonyo: Added a check to only allow absolute paths :+1:

pkg/api/config.go
tonyo: Makes sense, removed the File attribute

internal/machine/docker/server.go
tonyo: Decided against converting it to int on the client, mostly to preserve the compatibility of the internal representation with Compose and potentially support the "resolve-username-to-id" logic in the future. 

Added some validation logic on client and server side instead.

internal/machine/docker/server.go
tonyo: Interesting, it worked even without CopyUIDGID, but it seems that it's platform-specific and might be "fixed" in the next Docker versions; added the option to make it explicit.

tonyo:
> Btw, did you have any plans to add configs support to uc run command? For example, using a format similar to volumes: --config ./local/path:/container/path.

Not in the initial implementation for sure; I don't mind supporting it, but TBH wouldn't use this myself.

PR #132, opened 2025-09-27
fix: Create host directories explicitly

### LINKED ISSUE #118: [BUG] Error on `uc caddy deploy`: bind source path does not exist
**Describe the bug**

Getting the following error when running `uc caddy deploy`:

**Error: deploy caddy: create container: rpc error: code = Internal desc = Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /run/uncloud/caddy**

Full log:
```
✓ $ uc caddy deploy
Service: caddy (global mode)
Current image: caddy:2.10.0

Preparing a deployment plan...
Target image: caddy:2.10.2 (latest stable)

Deployment plan:
- hetzner: Run container [image=caddy:2.10.2]
- hetzner: Remove container [name=caddy-bqbr]

Do you want to continue?

Choose [y/N]: y
Chose: Yes!

[+] Deploying service caddy (global mode) 0/1
 ⠹ Container caddy-yvnt on hetzner  Creating                                                   0.1s 
Error: deploy caddy: create container: rpc error: code = Internal desc = Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /run/uncloud/caddy
```

**Expected behavior**

Caddy is redeployed without errors.

**Environment:**

- Uncloud versions:
  - Control (client) node (`uc --version`): `0.12.0`
  - Uncloud daemon (from the server) (`uncloudd --version`): `0.12.0`
- OS versions:
  - Server (`uname -a`):  `Linux 6.1.0-25-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.106-3 (2024-08-26) x86_64
`

tonyo:
This fixes bind mount directory creation when CreateHostPath is set to `true`.

`CreateMountpoint` is only responsible for creating the missing directories inside the container, so this was probably not intentional:
https://github.com/psviderski/uncloud/blob/aebd2ac232eb5d4eff249c677183b61f66f4c779/internal/machine/docker/server.go#L807

Instead we create the missing directories manually when CreateHostPath is `true`.

Fixes #118.

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -535,6 +535,9 @@ func (s *Server) CreateServiceContainer(
 	if err = s.verifyDockerVolumesExist(ctx, mounts); err != nil {
 		return nil, err
 	}
+	if err = ensureHostPaths(spec.Volumes); err != nil {
+		return nil, status.Errorf(codes.Internal, "ensure host directories: %v", err)
+	}
 
 	portBindings := make(nat.PortMap)
 	for _, p := range spec.Ports {
@@ -804,7 +807,7 @@ func toDockerBindOptions(opts *api.BindOptions) *mount.BindOptions {
 
 	dockerOpts := &mount.BindOptions{
 		Propagation:      opts.Propagation,
-		CreateMountpoint: opts.CreateHostPath,
+		CreateMountpoint: true, // Always create container mountpoint
 	}
 
 	switch opts.Recursive {
@@ -817,6 +820,23 @@ func toDockerBindOptions(opts *api.BindOptions) *mount.BindOptions {
 	}
 
 	return dockerOpts
+}
+
+// ensureHostPaths creates host directories for bind mounts that have CreateHostPath=true.
+func ensureHostPaths(volumes []api.VolumeSpec) error {
+	for _, vol := range volumes {
+		bindOptions := vol.BindOptions
+		if vol.Type != api.VolumeTypeBind || bindOptions == nil {
+			continue
+		}
+
+		if bindOptions.CreateHostPath && bindOptions.HostPath != "" {
+			if err := os.MkdirAll(bindOptions.HostPath, os.ModePerm); err != nil {
+				return fmt.Errorf("create host path for bind mount '%s': %w", bindOptions.HostPath, err)
+			}
+		}
+	}
+	return nil
 }
 
 // verifyDockerVolumesExist checks if the Docker named volumes referenced in the mounts exist on the machine.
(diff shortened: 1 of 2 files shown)


internal/machine/docker/server.go
```diff
@@ -804,7 +807,7 @@ func toDockerBindOptions(opts *api.BindOptions) *mount.BindOptions {
 
 	dockerOpts := &mount.BindOptions{
 		Propagation:      opts.Propagation,
-		CreateMountpoint: opts.CreateHostPath,
+		CreateMountpoint: true, // Always create container mountpoint
```
psviderski, line 810: According to the docs and code, this should actually create the host path, not the container one.

Image

`CreateMountpoint` option was introduced in https://github.com/moby/moby/pull/43484 to achieve parity between the old `Binds` and new `Mounts` (supersedes `Binds`property). So I guess it makes it possible to optionally *not* create the host path but as long as I remember Docker, it always created the host path if it didn't exist. 

Some related code after checking `SkipMountpointCreation` https://github.com/moby/moby/blob/28.x/volume/mounts/mounts.go#L159-L161:

https://github.com/moby/moby/blob/28.x/volume/mounts/mounts.go#L250-L264

This looks more like a bug or an edge case in Docker. What Docker version do you use?
Also, I remember you mentioned you use rootless mode somewhere, could this be the reason it's not able to create the dir? I'd be curious to check the docker logs after you run `uc caddy deploy`.

This change is good but it would be great to understand what specific edge case we're covering here and also double-check that we don't change the default Docker's behaviour with regards to the file mode, user and group (`user.MkdirAllAndChown`).

tonyo: You're right; a good example of how I was wrongly convinced by an LLM :disappointed_relieved: 

I'm not using rootless docker on the remote server, docker version there is:

```
Client: Docker Engine - Community
 Version:           24.0.9
 API version:       1.43
 Go version:        go1.20.13
 Git commit:        2936816
 Built:             Thu Feb  1 00:47:17 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          24.0.9
  API version:      1.43 (minimum version 1.12)
  Go version:       go1.20.13
  Git commit:       fca702d
  Built:            Thu Feb  1 00:47:17 2024
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.22
  GitCommit:        7f7fdf5fed64eb6a7caf99b3e12efcf9d60e311c
 runc:
  Version:          1.1.14
  GitCommit:        v1.1.14-0-g2c9f560
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```
(which doesn't seem that old)

Nothing interesting in the docker logs for now.

Will try to reproduce it on the same server with a simpler use case.

tonyo: https://github.com/psviderski/uncloud/issues/118#issuecomment-3368890350

So: let's not do this for now.

PR #135, opened 2025-10-05
Draft: Set machine id in service container ENV

jabr:
Just a draft of the idea, but I was thinking about how to expose info like the current machine ID to the services running on it. This is passing it in via the ENV.

Some thoughts:
1. I'm not entirely positive that it needs the callback approach to get the id here (like it does with the IP), but it _seemed_ like it was possible the machine ID would not be available yet when it was new a new machine still be initialized.

2. Are there other bits of info like machine ID that would be useful, and if so, would the ENV approach work for them to?
- service id?
- public IP (maybe wireguard IP -- not sure if useful?), but not sure if that is too dynamic/could change during lifetime of the instance

3. Is there a better way to do this than ENV? I currently solve this by writing a file with the machine id to a host file on each machine that is mounted into the service for it to read. Maybe something similar, but built-in to uncloud, would be a better approach.

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -57,6 +57,9 @@ type Server struct {
 	// internalDNSIP is a function that returns the IP address of the internal DNS server. It may return an empty
 	// address if the address is unknown (e.g. when the machine is not initialised yet).
 	internalDNSIP func() netip.Addr
+	// machineID is a function that returns the machine ID. It may return an empty string if the machine
+	// is not initialised yet.
+	machineID func() string
 	// networkReady is a function that returns true if the Docker network is ready for containers.
 	networkReady func() bool
 	// waitForNetworkReady is a function that waits for the Docker network to be ready for containers.
@@ -72,12 +75,13 @@ type ServerOptions struct {
 }
 
 // NewServer creates a new Docker gRPC server with the provided Docker service.
-func NewServer(service *Service, db *sqlx.DB, internalDNSIP func() netip.Addr, opts ServerOptions) *Server {
+func NewServer(service *Service, db *sqlx.DB, internalDNSIP func() netip.Addr, machineID func() string, opts ServerOptions) *Server {
 	s := &Server{
 		client:        service.Client,
 		service:       service,
 		db:            db,
 		internalDNSIP: internalDNSIP,
+		machineID:     machineID,
 	}
 
 	s.networkReady = opts.NetworkReady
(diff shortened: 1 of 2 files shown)


internal/machine/docker/server.go
```diff
@@ -529,9 +533,19 @@ func (s *Server) CreateServiceContainer(
 		containerName = fmt.Sprintf("%s-%s", spec.Name, suffix)
 	}
 
+	// Start with the environment from the service spec
+	envVars := spec.Container.Env.ToSlice()
+
+	// Inject the machine ID if available
+	if s.machineID != nil {
+		if machineID := s.machineID(); machineID != "" {
+			envVars = append(envVars, "UNCLOUD_MACHINE_ID="+machineID)
```
psviderski: Can you please not depend on the serialisation format (`key=value`) of the `Env` map for the Docker config here but instead just assign a new key-value in the map. I also think we don't want to modify the `Env` map in the user's spec so what we can do is to clone it (`maps.Clone`), assign a new `UNCLOUD_MACHINE_ID` value, and then use the `.ToSlice()` below in the config.

psviderski:
We should also create a new reference doc page for this and the DNS names. Any advice on how to structure it and which section to put?

internal/machine/docker/server.go
jabr: Updated.

jabr:
> Just one little comment on setting a new env var in the docker config.

Updated. Let me know if that change looks okay.

> Fly uses exactly this approach as well

The Fly env variables look pretty similar to other cases I'd thought of, but most either aren't relevant, already available, or implicit or easily embedded in the service spec (e.g. `FLY_APP_NAME`).

The public IP and service ID are still the only ones I can imagine coming up, but I agree waiting to add them until it does is reasonable.

I do already need the private IP in a service, but `hostname -i` works fine for me there. Having it in the env might be a bit more convenient for other cases, though. 

> We should also create a new reference doc page for this and the DNS names.

Definitely. Maybe a new section under "Concepts"? Maybe a "Services" or "Errata"? Start with useful details that don't have an obvious place for now and reorganize as it becomes more clear?

I'll try to think about what other bits weren't obvious to me as I got started with it as well.

internal/machine/docker/server.go
```diff
@@ -529,9 +534,18 @@ func (s *Server) CreateServiceContainer(
 		containerName = fmt.Sprintf("%s-%s", spec.Name, suffix)
 	}
 
+	envVars := maps.Clone(spec.Container.Env)
```
psviderski, line 537: The tests are failing for a reason but error messages are not informative because the the `uncloudd` daemon panics when trying to set a key in `envVars` when it's `nil`.
```suggestion
	envVars := maps.Clone(spec.Container.Env)
	if envVars == nil {
		envVars = make(api.EnvVars)
	}
```

psviderski:
Thanks for the update! I've fixed one edge case with a `nil` map and merged 🚀 

>I do already need the private IP in a service, but hostname -i works fine for me there. Having it in the env might be a bit more convenient for other cases, though.

This is a bit tricky as we rely on Docker for IP address management so we don't know the private IP before starting a container. In order to be able to inject it as an env var, we would need to do the IP management for the `uncloud` Docker network in the uncloud daemon and statically assign IPs to containers.
Having an env var is definitely more convenient but honestly I'd like to avoid dealing with IP management, race conditions, etc. if possible.

>Maybe a new section under "Concepts"? Maybe a "Services" or "Errata"? Start with useful details that don't have an obvious place for now and reorganize as it becomes more clear?

The "Services" section sounds good to me 👍 

>I'll try to think about what other bits weren't obvious to me as I got started with it as well.

That would be great! Feel free to share them in any way you like. Or you can even submit a PR with a markdown file in docs: https://github.com/psviderski/uncloud/tree/main/website. Any non-perfect info is better than missing docs as it's now 🙏

jabr:
Thanks for the fix!

I opened a PR with some quick notes on the internal DNS names and container ENV here: https://github.com/psviderski/uncloud/pull/136/files

PR #136, opened 2025-10-08
docs: Initial notes for "services" docs

jabr:
Examples of internal DNS lookups and container ENV.


psviderski (approved):
Awesome! Thank you so much for putting this together 🙏 
BTW, do you run nats? What are you hosting with uncloud now?

PR #139, opened 2025-10-10
feat: "exec" command to start processes inside remote containers

tonyo:
The PR adds initial implementation for "uc service exec".
In the current implementation, a random container from the service is picked if the service has multiple replicas.

Done:
- [x] Docs
- [x] Tests

Left a few TODOs:
* Handling Ctrl-C and other signals; currently we don't forward the signals via the stream
* We can be smarter about TTY detection; currently the code mirrors "docker compose" behavior
* Merging `docker/client_exec.go` and `pkg/client'
* Implementation should be better split between `service.go` and `server.go`

### DIFF
--- a/cmd/uncloud/service/exec.go
+++ b/cmd/uncloud/service/exec.go
@@ -41,7 +41,7 @@ If the service has multiple replicas, the command will be executed in a random c
   uc exec --container d792ea7347e5 web-service ls -la
 
   # Pipe input to a command inside the service container
-  cat /var/log/app.log | uc exec -T web-service grep "ERROR"
+  cat backup.sql | uc exec -T db-service psql -U postgres mydb
 
   # Run a task in the background (detached mode)
   uc exec -d web-service /scripts/cleanup.sh`,


cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+	tty         bool
+	context     string
+}
+
+func NewExecCommand() *cobra.Command {
+	opts := execOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "exec [OPTIONS] SERVICE COMMAND [ARGS...]",
+		Short: "Execute a command in a running container",
```
psviderski: nit: I prefer to not use the container term on its own as the main compute primitive in uncloud is a service, not container. So "Execute a command in a running service" or "Execute a command in a running service container" would be more consistent with other commands.
I think I used "service container" in multiple flag descriptions already.

cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+		Long: `Execute a command in a running container within a service.
+
+If the service has multiple replicas, the command will be executed in the first container.
```
psviderski: "the first container" is vague as it's unclear what the sorting criterion is. In fact, the current `InspectService` is non-deterministic as the results from multiple machines may arrive in random?/based on latency? order and we don't sort them automatically.

I think we should either say that it's an arbitrary/random container or implement some deterministic ordering and describe what it is here.

cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+
+Examples:
```
psviderski: You can use the `Example:` key for examples which formats as a separate section in auto-generated reference docs: https://github.com/psviderski/uncloud/blob/main/cmd/uncloud/image/ls.go#L36

cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+			return runExec(cmd.Context(), uncli, args[0], args[1:], opts)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&opts.detach, "detach", "d", false,
+		"Detached mode: run command in the background")
+	cmd.Flags().BoolVarP(&opts.interactive, "interactive", "i", false,
+		"Keep STDIN open even if not attached")
+	cmd.Flags().BoolVarP(&opts.tty, "tty", "t", false,
+		"Allocate a pseudo-TTY")
```
psviderski: As I mentioned earlier, I think we should consider enabling `-it` as the default behaviour:
1. It's a much friendlier default for users who want to run an interactive shell in a service container (for me this is the most common `docker exec` use case)
2. I can't think of major downsides of allocating a TTY even for non-shell commands `uc exec /bin/foo args`

We can still have these `-i` and `-t` flags but set the defaults for them to `true` and hide them from the help output. And maybe have a flag to disable the interactive mode if absolutely necessary. This will accept passing `-it` (no-op) to not annoy seasoned docker users who have this in their muscle memory.

Ah, I've checked that this is exactly the behaviour of `docker compose exec` (removed `-it` and added `-T`):
```
  -T, --no-TTY docker compose exec   Disable pseudo-TTY allocation. By default docker compose exec allocates a TTY.
```
and hidden `-it` flags: https://github.com/docker/compose/blob/main/cmd/compose/exec.go#L88-L91

The TTY allocation can also be configured based on the execution environment. E.g. if there is no tty where the `uc exec` command runs (CI for example) then do not enable it for `docker exec`. Some more thinking and modeling common use cases is needed to find out if this helps and not complicates things.
Just discovered that there is `docker compose run` command that has the `-T` flag:
```
  -T, --no-TTY                      Disable pseudo-TTY allocation (default: auto-detected)
```
here it states `auto-detected`. I haven't checked how it actually works.

What do you also think of setting the default command to `/bin/bash` or `/bin/sh` if none is provided?
This will allow opening an interactive shell in a service container using a nice short command:
```
uc exec app
```

Ideally, we should automatically detect the most advanced shell available in a container and use it. E.g. implement a cheap gRPC endpoint to check for paths in the image/container fs. But we can leave this for future improvements.

Later, `uc debug app` will complement this `exec` command to run a shell in a service container that doesn't have a shell binary.

cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+
+  # Start an interactive shell
+  uc exec -it web-service bash
+
+  # Run a background task
+  uc exec -d web-service /scripts/cleanup.sh`,
+		Args: cobra.MinimumNArgs(2),
+		// DisableFlagParsing would disable all flag parsing, but we want to parse our own flags.
+		// Instead, we'll use FParseErrWhitelist to ignore unknown flags which will be passed to the command.
+		FParseErrWhitelist: cobra.FParseErrWhitelist{UnknownFlags: true},
```
psviderski: I'm not familiar with this configuration but is it needed even when `SetInterspersed(false)` set?

cmd/uncloud/service/exec.go
```diff
@@ -0,0 +1,95 @@
...
+		return fmt.Errorf("connect to cluster: %w", err)
+	}
+	defer c.Close()
+
+	// Build the exec configuration
+	execConfig := container.ExecOptions{
+		Cmd:          command,
+		AttachStdin:  opts.interactive,
+		AttachStdout: true,
+		AttachStderr: true,
```
psviderski: Should these depend on `opts.detach`?
https://github.com/docker/cli/blob/master/cli/command/container/exec.go#L236-L243

cmd/uncloud/main.go
```diff
@@ -87,6 +87,7 @@ func main() {
 		image.NewRootCommand(),
 		machine.NewRootCommand(),
 		service.NewRootCommand(),
+		service.NewExecCommand(),
```
psviderski, line 93: I guess this is a shortcut `uc exec` but should we also add it under `uc service exec` for consistency with other `service` commands?

internal/machine/api/pb/docker.proto
```diff
@@ -30,6 +30,8 @@ service Docker {
   rpc InspectServiceContainer(InspectContainerRequest) returns (ServiceContainer);
   rpc ListServiceContainers(ListServiceContainersRequest) returns (ListServiceContainersResponse);
   rpc RemoveServiceContainer(RemoveContainerRequest) returns (google.protobuf.Empty);
+
+  rpc ContainerExec(stream ContainerExecRequest) returns (stream ContainerExecResponse);
```
psviderski: nit: looking at other methods, `ExecContainer`, `ExecContainerRequest,` and `ExecContainerResponse` would be more preferred names

internal/machine/api/pb/docker.proto
```diff
@@ -214,3 +216,41 @@ message MachineServiceContainers {
...
+    bytes stdin = 2;
+    // TTY resize event (only used when TTY is enabled).
+    ResizeEvent resize = 3;
+  }
+}
+
+message ExecConfig {
+  // Container ID to execute the command in.
+  string container_id = 1;
+  // JSON serialised types.ExecConfig (includes Cmd, AttachStdin, AttachStdout, AttachStderr, Tty, etc).
```
psviderski: `container.ExecOptions` to be precise?

internal/machine/docker/server.go
```diff
@@ -1011,3 +1039,176 @@ func (s *Server) RemoveServiceContainer(ctx context.Context, req *pb.RemoveConta
...
+		return status.Errorf(codes.Internal, "create exec: %v", err)
+	}
+
+	// Send the exec ID back to the client
+	if err := stream.Send(&pb.ContainerExecResponse{
+		Payload: &pb.ContainerExecResponse_ExecId{ExecId: execResp.ID},
+	}); err != nil {
+		return status.Errorf(codes.Internal, "send exec ID: %v", err)
+	}
+	slog.Info("Sent exec ID", "exec_id", execResp.ID)
```
psviderski: Feels more like a Debug log rather than Info. If you intend to keep it, please make the message more descriptive

internal/machine/docker/server.go
```diff
@@ -1011,3 +1039,176 @@ func (s *Server) RemoveServiceContainer(ctx context.Context, req *pb.RemoveConta
...
+		// If not attaching stdin, close the write side immediately
+		attachResp.CloseWrite()
+	}
+
+	// Goroutine to read from Docker stdout/stderr and write to gRPC stream
+	go func() {
+		slog.Info("Output goroutine started", "exec_id", execResp.ID, "tty", config.Tty)
+		if config.Tty {
+			// In TTY mode, all output is stdout - just copy directly
+			buf := make([]byte, 32*1024)
```
psviderski: Is there any reasoning behind the size of the buffer or just a reasonable size that doesn't matter much?

internal/machine/docker/server.go
```diff
@@ -1011,3 +1039,176 @@ func (s *Server) RemoveServiceContainer(ctx context.Context, req *pb.RemoveConta
...
+			}
+			return
+		}
+	}()
+
+	// Wait for the output goroutine to complete (it signals on errCh when done)
+	// We only wait for 1 error from the output goroutine, not from stdin goroutine.
+	// The stdin goroutine may still be blocked in stream.Recv() waiting for client data,
+	// and that's OK - it will be cleaned up when the RPC returns.
+	if err := <-errCh; err != nil {
```
psviderski: It doesn't seem we check that it was the output goroutine which completed and sent a message to `errCh`. An error could occur in the stdin one or it may just finish earlier but the output one may still be running

internal/machine/docker/server.go
```diff
@@ -1011,3 +1039,176 @@ func (s *Server) RemoveServiceContainer(ctx context.Context, req *pb.RemoveConta
 
 	return resp, nil
 }
+
+// ContainerExec executes a command in a running container with bidirectional streaming for stdin/stdout/stderr.
+func (s *Server) ContainerExec(stream pb.Docker_ContainerExecServer) error {
```
psviderski: I started consolidating all the business logic in a separate Docker service: https://github.com/psviderski/uncloud/blob/719109812f7cb7a4f3a6c69d9eeec38184879557/internal/machine/docker/service.go so that the server is just a thin gRPC layer for the service. So let's follow this pattern for the new code.
The service is also used internally by the Docker controller.

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+	// Exec configuration.
+	Config container.ExecOptions
+	// Stdin, Stdout, Stderr streams. If nil, os.Stdin/Stdout/Stderr will be used.
+	Stdin  io.Reader
+	Stdout io.Writer
+	Stderr io.Writer
+}
+
+// ExecContainer executes a command in a running container with bidirectional streaming.
+func (c *Client) ExecContainer(ctx context.Context, opts ExecOptions) (int, error) {
```
psviderski: This is not critical but I'm planning to merge this separate docker client into a single `pkg/client` as failed to define a clear boundary between them and there is not much value in having this intermediate layer. All the client methods anyway make sense only in the context of working with uncloud resources.

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+									Height: uint32(size.Height),
+									Width:  uint32(size.Width),
+								},
+							},
+						})
+					}
+				}
+			}()
+
+			// Send initial window size
```
psviderski: It's probably doesn't matter but it feels more natural for me to send the initial window size first and only then start the goroutine to resize in response to signals (subscribing to signals early is perhaps good). Otherwise, there is a minor chance they both may send resize message concurrently and I'm not sure if this may result in reverting the correct last size.

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+				})
+			}
+		}
+	}
+
+	var wg sync.WaitGroup
+
+	// Goroutine to read from stdin and send to stream
+	if opts.Config.AttachStdin {
+		wg.Add(1)
```
psviderski: FYI Go 1.25 introduced `wg.Go` syntax sugar: https://antonz.org/go-1-25/#go-wait-group-go
and we migrated to 1.25 recently.

You may also consider using `errgroup.WithContext` that will automatically handle `errCh` for you and context cancellation. We use it in a couple places

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// Wait for all goroutines to complete
+	// The output goroutine will complete when it receives the exit code from the server
+	// The stdin goroutine (if running) may be blocked on Read(), but cancel() signals it to stop
+	wg.Wait()
+
+	// Check if any goroutine reported an error
+	select {
+	case err := <-errCh:
+		return exitCode, err
```
psviderski: Do you want to set a non-zero initial value to make sure we only return 0 when it's explicitly set from the payload? We may introduce bugs in the code above later so this may help troubleshoot.

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+					errCh <- fmt.Errorf("write stderr: %w", err)
+					return
+				}
+			case *pb.ContainerExecResponse_ExitCode:
+				exitCode = int(payload.ExitCode)
+				// Cancel context to stop stdin reader if it's still running
+				cancel()
+				// Return immediately after getting exit code to avoid waiting
+				return
+			case *pb.ContainerExecResponse_Error:
```
psviderski: It doesn't seem we send any Error payload, do we need it?

internal/machine/docker/client_exec.go
```diff
@@ -0,0 +1,252 @@
...
+						errCh <- fmt.Errorf("read stdin: %w", err)
+					}
+					return
+				}
+			}
+		}()
+	}
+	// Note: We don't call CloseSend() for non-interactive mode.
+	// The stream will be closed when this function returns.
+	// Calling CloseSend() early can cause the server's context to be canceled prematurely.
```
psviderski: Did you investigate why this happens? It doesn't sound like it should behave like that. Maybe it's because https://github.com/psviderski/uncloud/pull/139/files#r2427945079 has to be fixed to prevent early return while the output gorouting is still running?

cmd/uncloud/service/exec.go
tonyo: This part (how to pick the container) I haven't finalized yet, we can discuss. 
By default it can be random, but we can also let people provide the container ID, and also let specify a machine where it runs.

So e.g. we have 

```
✓ $ uc inspect test
ID:    6133c696ea9003e49620465261be4aff
Name:  test
Mode:  replicated

CONTAINER ID   IMAGE         CREATED          STATUS          MACHINE
2b250daefc24   alpine:3.20   22 minutes ago   Up 22 minutes   machine-1
```

So it'll be possible to do
`uc exec test -c  2b250daefc24` (this will uniquely identify a container)
or 
`uc exec test -m machine-1` (this might be ambiguous, so we'll have to document it clearly).
Wdyt?

cmd/uncloud/service/exec.go
tonyo: TIL, nice

cmd/uncloud/service/exec.go
tonyo: cool cool, `-T` sounds good, the default to /bin/bash or /bin/sh makes sense as well.

cmd/uncloud/service/exec.go
psviderski: I think we should come up with a universal language for selecting a service container and use it consistently for all commands where it's applicable (exec, logs, rm, restart, etc.). For example, 
* `<service-name-or-id> --container <container-name-or-id>` — `-c` is already used by `--context`
* `<service-name-or-id>/<container-name-or-id>` — can support specifying multiple service containers
* `<service-name-or-id> <container-name-or-id>` — positional arguments (can specify only one service container)
* `<container-name-or-id>` — by default, a container name is `service-name-RANDOM_ASCII{4}`, there is minor chance of collision if many containers on multiple machines.

I like `<service-name-or-id>/<container-name-or-id>` which is concise. But it's a bit harder to discover than the explicit `--container` flag. This format can potentially support `project-or-namespace/service/container` in the future.

I also like how we already consistently use `-m|--machines` filter in multiple commands. So I think it make sense for this command as well.

tonyo: Implemented "--container" for now; alternative options with slashes will need a bit more thinking and better understanding of how we want to move forward with the project (namespace) support.

cmd/uncloud/service/exec.go
tonyo: Looks like it's not needed indeed, removed

internal/machine/docker/server.go
tonyo: More of a "reasonable size" approach (e.g. it's also used by `io.CopyBuffer`). It's a bit sad that this read loop is manual; we could use `io.Copy` here, but we should also handle `ctx.Done()` at least somewhat occasionally, and `io.Copy` might block for longer than necessary when reading from the input stream.

internal/machine/docker/server.go
tonyo: Yesss, should be fixed now

internal/machine/docker/client_exec.go
tonyo: Currently we're sending the errors directly via grpc streams, so doesn't look like we need it after all

internal/machine/docker/server.go
```diff
@@ -1012,3 +1014,250 @@ func (s *Server) RemoveServiceContainer(ctx context.Context, req *pb.RemoveConta
...
+	}()
+
+	// Wait for the output goroutine to complete (it signals when done)
+	// We only wait for output, not for stdin goroutine.
+	if err := <-outputDone; err != nil {
+		slog.Warn("Error in exec output handler", "err", err, "exec_id", execResp.ID)
+	}
+	// The stdin goroutine may still be blocked in stream.Recv() waiting for client data,
+	// so cancel it explicitly.
+	cancel()
```
psviderski, line 1245: I see the intention but it doesn't seem it works this way. The passed `handlerCtx` is not handled in any way in `handleServerExecInput`. I'm not sure if it's possible to interrupt `.Recv` but maybe it's not even needed as it will unblock when the client disconnects anyway.
But this cancelable context is a bit misleading

tonyo: Ah yes, missed the context cancellation handling, added it here and updated the comments: https://github.com/psviderski/uncloud/commit/30bb33f1a06720207706fc8ecbc64dc3d6a5cf1c
You're right that `.Recv` is blocking and in some cases we won't be able to cancel the input-handling goroutine properly, but I think it still makes sense to attempt the best-effort cancellation to reduce potentially cryptic error messages in the logs.

PR #152, opened 2025-10-19
Connect to remote SSH nodes using SSH CLI

### LINKED ISSUE #131: Possible for Uncloud to use SSH configuration? (or provide an specific identity?)
I've machines provisioned with SSH keys that are stored in 1Password SSH agent:

https://developer.1password.com/docs/ssh/agent/

The agent is running and is the default SSH_AUTH_SOCK, so doing `ssh-add -L` list all the SSH keys available in all the vaults I have access to.

```
$ ssh-add -L | wc -l
      73
```

Due that, when connecting to SSH, I normally receive the following error:

```
Received disconnect from REDACTED-IPV6::1 port 22:2: Too many authentication failures
Disconnected from REDACTED-IPV6::1 port 22
```

To workaround that, I've explicitly defined a SSH configuration in my ~/.ssh/config file for that server:

```
Host my-machine
    HostName my-machine
    User core
    IdentityAgent ~/.1password/agent.sock
    IdentitiesOnly yes
    IdentityFile ~/.ssh/keys/my-user-ssh-key.pub
```

This works great for `ssh`:

```
$ ssh my-machine "whoami"
core

$ ssh core@my-machine "whoami"
core
```

And also work for `docker`:

```
$ DOCKER_HOST=ssh://core@my-machine docker network ls
NETWORK ID     NAME      DRIVER    SCOPE
2e677f8e8266   bridge    bridge    local
bc4366252ae3   host      host      local
872afa4e11ba   none      null      local
```

However, the same approach fails for uncloud (`~/bin/uncloud` is the latest build from `main`):

```
$ UNCLOUD_CONNECT=ssh://core@my-machine ~/bin/uncloud machine ls
Error: connect to cluster: connect to machine: SSH login to core@my-machine:22: connect using SSH agent: ssh: handshake failed: ssh: disconnect, reason 2: Too many authentication failures
```

It seems Docker relies on standard `ssh` command and has a _hidden_ `dial-stdio` plugin under their system subcommand:

https://github.com/docker/cli/blob/75f3c08257ea0f04ddb57ec09afb7af182c36134/cli/connhelper/connhelper.go#L43-L66

Not submitting this as a feature request or a bug, but putting out there to gather feedback if I'm the only one on this scenario in order to determine if is worth to invest time on replace current SSH implementation with something similar.

Right now I've a workaround locally by commenting out all the other vaults and leave only the one that has the SSH keys for the active cluster. Also this is not a general issue as SSH to those machines is managed via Tailscale SSH.

Thank you.
❤️ ❤️ ❤️

luislavena:
Replace Go-native SSH implementation with SSH CLI execution to support diverse SSH configurations and agents. Implements 'uncloudd dial-stdio' subcommand that proxies gRPC connections over stdin/stdout, similar to Docker's approach.

This change addresses compatibility issues with:
- SSH agents exposing many keys (1Password, causing "too many authentication failures")
- Tailscale SSH (which doesn't support advanced SSH channel types like direct-streamlocal)
- Custom SSH configurations in ~/.ssh/config

The dial-stdio approach reduces SSH feature requirements by streaming the unix socket connection over stdin/stdout instead of using SSH channel forwarding.

Changes:
- Add 'uncloudd dial-stdio' hidden subcommand for socket proxy
- Add SSHCLIConnector using ssh command + dial-stdio
- Update connection logic to use SSH CLI connector
- Maintain backward compatibility with SSHKeyFile config

Resolves #131

### DIFF
--- a/cmd/uncloudd/dialstdio.go
+++ b/cmd/uncloudd/dialstdio.go
@@ -118,6 +118,8 @@ func runDialStdio(ctx context.Context, socketPath string, stdin io.Reader, stdou
 		if err != nil {
 			return err
 		}
+		// wait for stdout
+		err = <-socket2stdout
 	case err = <-socket2stdout:
 		// return immediately, matching Docker's approach
 		// (stdin is never closed when TTY)


luislavena:
Hola!

Took a quick stab at this as it was a bit painful for me having to unlock 1Password every single time when Tailscale manages it transparently and keyless 😅 

I didn't remove the current SSH connector, but I replaced it with SSHCLI one for the CLI to use it.

This is a very naive implementation. Blindly stole the logic from Docker's dial stdio implementation, but is not exactly the same as I didn't fully understand the code (I'm not versed in all the dark incantations of Go) 😬 

Since there was no SSH test, I wasn't sure if we should be testing it or not, so I added a kinda pointless test (actually was Claude Code who did it, but I take responsibility for that) 😁

I've tested it against my fleet of machines connected to my tailnet and my finger is much better that I don't have to unlock 1Password on every deployment 🤣 

Let me know what do you think!

Cheers
❤️ ❤️ ❤️

cmd/uncloudd/dialstdio.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	defer conn.Close()
+
+	// Copy data bidirectionally between stdin/stdout and the socket.
+	errCh := make(chan error, 2)
+
+	// Copy from stdin to socket.
+	go func() {
+		_, err := io.Copy(conn, stdin)
+		errCh <- err
```
psviderski: Not sure if half-closing the connection is critical here but looks like `docker dial-stdio` implements this. Check out my proxy implementation for I hope a simpler implementation: https://github.com/psviderski/uncloud/blob/main/internal/proxy/proxy.go#L100-L116.

We basically check if the connection satisfies the `halfCloser` interface (has `CloseWrite()` method). If so, call it.

cmd/uncloudd/dialstdio.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Wait for either direction to complete or context to be canceled.
+	select {
+	case err := <-errCh:
+		// One direction finished (possibly with an error).
+		// This is normal when the connection closes.
+		if err != nil && err != io.EOF {
+			return fmt.Errorf("copy error: %w", err)
+		}
+		return nil
```
psviderski: I'm not sure if returning right after any of the direction is complete won't be problematic in some edge cases. If the stdin is complete, shouldn't we still wait until the stdout is fully written?
If think this is what docker does: https://github.com/docker/cli/blob/171a9b70b2736c719019393259ecebdcac47bead/cli/command/system/dial_stdio.go#L62-L71

pkg/client/connector/sshcli.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Dialer returns a proxy dialer for establishing connections within the cluster through the SSH tunnel.
+func (c *SSHCLIConnector) Dialer() (proxy.ContextDialer, error) {
+	if c.conn == nil {
+		return nil, fmt.Errorf("SSH connection must be established first")
+	}
+	// For SSH CLI connector, we can't provide a generic dialer since we only have
+	// a single connection to the dial-stdio process.
+	// This matches the limitation of the old SSH connector.
```
psviderski: We unfortunately can't switch to this connector by default unless this `Dialer` method is implemented. `uc image push` depends on the dialer to proxy to the `unregistry` server on the remote machine over the ssh connection. Currently, the Go SSH client provides the required method: https://github.com/luislavena/uncloud/blob/main/pkg/client/connector/ssh.go#L90

With the `ssh` CLI approach, I think the dialer implementation would be similar to `Connect` but instead of calling the `uncloud dial-stdio` remote command, we need to connect to the requested dial_host:dial_port.
An example SSH command to dial to `unregistry` listening on `5000/tcp`:
```
ssh -W 10.210.1.1:5000 user@host
```

```
     -W host:port
             Requests that standard input and output on the client be forwarded to host on
             port over the secure channel.  Implies -N, -T, ExitOnForwardFailure and
             ClearAllForwardings, though these can be overridden in the configuration file
             or using -o command line options.
```

I suspect that `-W` requires an advanced SSH channel (the same that is required for dialing the uncloud socket over the current ssh connection?). Can you please test this command over your machines/ssh servers?

Alternatively, we can also change the `uc image push` implementation to use the local port forwarding (`ssh -L` flag) instead of using the dialer interface. I'm not sure but assume that local port forwarding is yet another ssh channel that if I'm not mistaken supported by Tailscale SSH.

internal/cli/connect.go
```diff
@@ -64,13 +64,13 @@ func connectCluster(ctx context.Context, conn config.MachineConnection) (*client
...
 
-		sshConfig := &connector.SSHConnectorConfig{
+		sshCliConfig := &connector.SSHCLIConnectorConfig{
 			User:    user,
 			Host:    host,
 			Port:    port,
 			KeyPath: keyPath,
 		}
-		return client.New(ctx, connector.NewSSHConnector(sshConfig))
+		return client.New(ctx, connector.NewSSHCLIConnector(sshCliConfig))
```
psviderski: We can't switch to `SSHCLIConnector` by default until `SSHCLIConnector.Dialer` is fully supported. Please see another comment https://github.com/psviderski/uncloud/pull/152/files#r2443816679

What do you think about optionally enabling this connector using the connection schema `--connect ssh+cli://` in additiona to the existing `ssh://` and `tcp://`?
```
      --connect string          Connect to a remote cluster machine without using the Uncloud configuration file. [$UNCLOUD_CONNECT]
                                Format: [ssh://]user@host[:port] or tcp://host:port
```
something similar could also be supported in the uncloud yaml config.

This way we can easily iterate on it until it reaches feature parity with the existing `SSHConnector`. Once it's well tested and supports everything we need, we can switch the default.

internal/cli/cli.go
psviderski, line 440: Am I right that the new connector is only used for already provisioned machines? So for provisioning new machines you will still need to change your ssh agent setup, right?

This is out of scope of this PR, but I guess we'd need to also implement the [`Executor`](https://github.com/psviderski/uncloud/blob/d735d568b36759bb1c3f91ed00fd20c8d9cbdce7/internal/sshexec/executor.go#L10-L14) using the `ssh` CLI to be able to fully migrate to this connector.

psviderski (commented):
Great work on the new SSH connector! 👏 

Looks much simpler than what I expected to see. While it's on the right track, please see a few comments and questions we need to resolve before we can merge it.

luislavena:
Hello @psviderski, thank you for your patience and looking into this!

In the case of Tailscale, it does support both port forwarding and IO forwarding transiently, so the Dialer should work.

> Am I right that the new connector is only used for already provisioned machines? So for provisioning new machines you will still need to change your ssh agent setup, right?

Ideally the connector should replace all the flows, including that to avoid discrepancies.

> What do you think about optionally enabling this connector using the connection schema `--connect ssh+cli://` in additional to the existing `ssh://` and `tcp://`?

That is a great idea, just for testing I did `ssh_cli:` for the configuration file, so we can test these side-by-side.

I went ahead and completed the Dialer so `uc image push` works.

Adding responses do your other comments now 😊 

Important note: I run out of brain for some stuff so decided to use Claude Code to assist in some cleanup and address some of the feedback. Note that it was heavily guided as the generated code was not verbatim as it generated it. Anyhow, wanted to disclose that as I'm not sure what would be your position towards AI-assisted contributions.

Thank you.
❤️ ❤️ ❤️

cmd/uncloudd/dialstdio.go
luislavena: I haven't addressed this as I didn't fully understand it. Now the issue seems clearer and I will bring the same approach to the implementation. Thank you!

pkg/client/connector/sshcli.go
luislavena: Responded about `-W` and `-L` in the main comment, but Tailscale SSH does support both transparently, so it will work.

With the change done in latest commits, `image push` should work (encountered another bug, #153) but should be fine.

internal/cli/connect.go
luislavena: I kinda like it. Right now I've only added it to the config part as `ssh_cli` vs `ssh` types of entries. Done that for testing between the two, but didn't address `--connect` support 😓

internal/cli/cli.go
luislavena: > This is out of scope of this PR, but I guess we'd need to also implement the [`Executor`](https://github.com/psviderski/uncloud/blob/d735d568b36759bb1c3f91ed00fd20c8d9cbdce7/internal/sshexec/executor.go#L10-L14) using the `ssh` CLI to be able to fully migrate to this connector.

I kinda would like to have a single connector instead of having to maintain two, so Executor will also need to be implemented with SSH CLI too.

internal/cli/config/connection_test.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := tt.conn.Validate()
+			if tt.wantErr {
+				if err == nil {
+					t.Errorf("Validate() expected error containing %q, got nil", tt.errMsg)
```
psviderski: Please use the https://github.com/stretchr/testify library as in all other tests. I think one AI prompt asking to refactor using testify should help.

internal/cli/config/connection_test.go
```diff
@@ -0,0 +1,193 @@
...
+			} else {
+				if err != nil {
+					t.Errorf("Validate() unexpected error = %v", err)
+				}
+			}
+		})
+	}
+}
+
+func contains(s, substr string) bool {
```
psviderski: I think you can just use `assert.Contains`

pkg/client/client.go
```diff
@@ -57,6 +58,14 @@ func New(ctx context.Context, connector Connector) (*Client, error) {
...
 	c.Caddy = pb.NewCaddyClient(c.conn)
 	c.Docker = docker.NewClient(c.conn)
+
+	// Validate the connection by making a simple RPC call.
+	// This ensures we fail fast if the connection is broken, allowing failover to work.
+	if _, err = c.MachineClient.Inspect(ctx, &emptypb.Empty{}); err != nil {
+		c.Close()
+		return nil, fmt.Errorf("validate connection: %w", err)
+	}
+
```
psviderski: I see the reason of this call but usually prefer to not do that in a constructor method but do it explicitly where needed. There could be valid cases where we might want to create a client before the connection can be established and do not use it until then. With this change, we won't be able to do that.

Calling `.Inspect()` or adding a special `.Ping()` method and calling it where we need to check connectivity is usually a more flexible approach.

internal/cli/connect.go
```diff
@@ -56,26 +56,42 @@ func connectClusterWithProgress(ctx context.Context, conn config.MachineConnecti
...
+		sshDest = conn.SSH
+		useSSHCLI = false
+	} else if conn.SSHCLI != "" {
+		sshDest = conn.SSHCLI
+		useSSHCLI = true
 	} else if conn.TCP != nil && conn.TCP.IsValid() {
 		return client.New(ctx, connector.NewTCPConnector(*conn.TCP))
+	} else {
+		return nil, errors.New("connection configuration is invalid")
 	}
```
psviderski, line 78: Did you want to use the `.Validate()` method somewhere here?

internal/cli/cli.go
psviderski: Are you planning to do it in this PR as well?

internal/cli/config/connection_test.go
```diff
@@ -0,0 +1,193 @@
...
+	return len(s) >= len(substr) && (s == substr || len(s) > len(substr) && findSubstring(s, substr))
+}
+
+func findSubstring(s, substr string) bool {
```
pieroshka: LLMs love to add this; modern Go uses https://pkg.go.dev/strings#Contains

luislavena:
Hello @psviderski, thank you for your patience.

Finally I was able to take a look to the feedback and address it.

- Kept both ssh and ssh+cli as connection schema, and made it work with `--connect`
- Dealt similar to Docker with stdio/socket connection (introduced halfCloser interface)
- Got rid of the inspect during connection, we can deal with that later
- Corrected the AI-slop associated with tests. I take full responsibility about that 😅 

re: sshexec, interestingly, I didn't change anything and it works out of the box with Tailscale SSH (no need for SSH keys or anything) but then either `init` or `add` fail as it tries to use the SSH connector to initialize the node:

```
$ ./uncloud --uncloud-config ./blatta.yaml machine init root@blatta11 --no-dns --public-ip none
Downloading Uncloud install script: https://raw.githubusercontent.com/psviderski/uncloud/refs/heads/main/scripts/install.sh
⏳ Running Uncloud install script...
✓ Docker is already installed.
Client: Docker Engine - Community
 Version:           28.5.1
 API version:       1.51
 Go version:        go1.24.8
 Git commit:        e180ab8
 Built:             Wed Oct  8 12:17:24 2025
 OS/Arch:           linux/amd64
 Context:           default
...
✓ Linux user 'uncloud' already exists.
✓ uncloudd binary is already installed.
✓ Systemd unit file created: /etc/systemd/system/uncloud.service
✓ uncloud-corrosion binary is already installed.
✓ Systemd unit file created: /etc/systemd/system/uncloud-corrosion.service
⏳ Starting Uncloud machine daemon (uncloud.service)...
✓ Uncloud machine daemon started.
✓ Uncloud installed on the machine successfully! 🎉
Error: inspect machine: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: connect to machine API socket '/run/uncloud/uncloud.sock' through SSH tunnel (is uncloud.service running on the remote machine and does the SSH user 'root' have permissions to access the socket?): ssh: rejected: unknown channel type (unsupported channel type)"
```

Note that `blatta.yaml` does not exist, this is the first machine to be initialized but is not possible with this scenario.

I would love to tackle the sshexec/Executor aspect in another PR, for the time being, I can rely on 1Password SSH Agent socket and the connection for the time being.

Apologies for the force push, I wanted to ensure I was testing against the latest changes from `main` but I didn't wanted to merge main inside (I prefer to avoid the merges-on-merges scenarios).

Thank you in advance for looking into this and I hope this looks much better now 😊 

Cheers,
❤️ ❤️ ❤️

cmd/uncloudd/dialstdio.go
```diff
@@ -0,0 +1,126 @@
...
+	}()
+
+	select {
+	case err = <-stdin2socket:
+		if err != nil {
+			return err
+		}
+	case err = <-socket2stdout:
+		// return immediately, matching Docker's approach
+		// (stdin is never closed when TTY)
```
psviderski, line 125: Does this comment mean that `_, err := io.Copy(conn, stdin)` may never finish if stdin is a TTY, hence `err = <-stdin2socket` never succeeds?

cmd/uncloudd/dialstdio.go
```diff
@@ -0,0 +1,126 @@
...
+		if stdoutCloser != nil {
+			stdoutCloser.CloseWrite()
+		}
+	}()
+
+	select {
+	case err = <-stdin2socket:
+		if err != nil {
+			return err
+		}
```
psviderski, line 120: Do I understand the intended behaviour correctly?

If we finished reading from stdin and there is an error, return it and complete the `dial-stdio` command.
If there is no error, complete the `dial-stdio` command, potentially not finishing reading everything from socket to stdout. Note that cases in the `select` statements don't fallback to the next one.

Ah, I checked the docker implementation, and it indeed contains the missing logic to wait for another goroutine before returning in this case:
https://github.com/docker/cli/blob/7414d73fc119a6b97ff240b7c7df91f1f642e323/cli/command/system/dial_stdio.go#L66-L67

luislavena:
Hello @psviderski, addressed the comment about stdio, I completely missed it! 🤦 

Ready for last review! I've already worked a bit on a SSH CLI connector, but I will explain that more in detail in a separate PR.

Thank you!
❤️ ❤️ ❤️

PR #156, opened 2025-10-26
feat: support DeviceRequests (service.gpus and reservations.devices)

jabr:
Update: I'm successfully running an uncloud service with GPU access using a custom `uc`/`uncloudd` of this branch.

New tests pass and it compiles for me ~, but I have not yet tried this on a real deploy~ .

### DIFF
--- a/pkg/client/deploy/container.go
+++ b/pkg/client/deploy/container.go
@@ -83,11 +83,17 @@ func EvalContainerSpecChange(current api.ServiceSpec, new api.ServiceSpec) Conta
 		}
 	}
 
+	// Device reservations are immutable, so we'll need to recreate if any have changed
+	if !reflect.DeepEqual(current.Container.Resources.DeviceReservations, newResources.DeviceReservations) {
+		return ContainerNeedsRecreate
+	}
+
 	// Check if any mutable properties changed.
 	if !current.Caddy.Equals(new.Caddy) {
 		return ContainerNeedsRecreate
 	}
 
+	// Remaining resources are mutable.
 	if !reflect.DeepEqual(current.Container.Resources, newResources) {
 		return ContainerNeedsUpdate
 	}
--- a/pkg/client/deploy/container_test.go
+++ b/pkg/client/deploy/container_test.go
@@ -3,6 +3,7 @@ package deploy
 import (
 	"testing"
 
+	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/mount"
 	"github.com/psviderski/uncloud/pkg/api"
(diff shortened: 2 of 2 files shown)


jabr:
I'm now running this build of `uncloudd` on my GPU server and successfully deployed a test service with this `uc` build:
```
services:
  nvidia:
    image: nvidia/cuda:12.8.1-base-ubuntu24.04
    command: tail -f /dev/null
    x-machines: [ gpu-server ]
    gpus: all
```

I then verified it has access to the GPU with (run on `gpu-server`):
```
docker exec <container-id-of-nvidia-service> nvidia-smi
```

pkg/api/resources.go
```diff
@@ -13,4 +17,6 @@ type ContainerResources struct {
 	// MemoryReservation is the minimum amount of memory (in bytes) the container needs to run efficiently.
 	// TODO: implement a placement constraint that checks available memory on machines.
 	MemoryReservation int64
+	// DeviceRequests for access to things like GPUs
+	DeviceRequests []container.DeviceRequest
```
psviderski: Just a tiny nit, regardless of the name of the docker internal struct, what do you think of calling it `DeviceReservations` to be consistent with the compose terminology and config (limits and reservations) and the `MemoryReservation` above?

jabr: I’m happy to change it if you want. I just used that because that’s what the docker container api used (it’s not consistent with the `MemoryReservation` naming pattern): https://pkg.go.dev/github.com/docker/docker/api/types/container#Resources

jabr: Oh, I see. You want to be consistent with the compose pattern, not the docker struct.

I’ll update it in a bit.

jabr: Updated.

jabr:
I renamed `DeviceRequests` to `DeviceReservations` and updated `EvalContainerSpecChange` to return a `ContainerNeedsRecreate` instead of `ContainerNeedsUpdate` when they are changed.

jabr:
Just a heads up that I apparently broke something in the last cleanup commit or two on this. The DeviceRequests aren't getting sent in the spec to the servers anymore so a subsequent redeploy of a service didn't have GPU access.

I'll take a look shortly. Apologies for not testing it again after the initial commit. 😢

psviderski:
No worries at all and thank you for raising this! Do you need any help with troubleshooting?

jabr:
Ah, nevermind, false alarm.

I still had the first version of the daemon running on my server. It was still expecting the `DeviceRequests` name that we changed to `DeviceReservations`.

I updated both it and my local `uc` to a fresh build from `main` and it's working fine.

Issue #154, opened 2025-10-22
Sort container IP in embedded DNS starting from local containers


psviderski:
When accessing a service with multiple-replicas by a DNS name, we should prioritise the IPs of the replicas running on the service host first to minimise latency. I guess sorting IPs in the DNS response should achieve this.

jabr:
A thought I had about this: I wonder if some applications might still need the option for a random (or a round-robin) behavior. (Or perhaps a "not local", or "local last"?)

I was specifically thinking of a case like the NATS cluster peer discovery with all nodes configured to `nats://cluster.internal`. While it is not a problem for NATS itself (it gets all IPs, excludes its own, and then randomly picks one: https://github.com/nats-io/nats-server/blob/main/server/route.go#L2913), it might be an issue for other applications that aren't as smart and always use the first IP which would always be itself.

Maybe there should be a `rr.{service}.internal` that does round-robin (or random) as well?

Or, while I do expect local first is a better default for most uses, maybe it would be least surprising if `{service}.internal` wasn't smart and instead there was a new `nearest.{service}.internal`? It could just do local first initially, but that could later be updated with some lowest latency logic, too.

jabr:
Initial draft of the "nearest.service.internal" mode putting machine-local (same subnet) IPs at the beginning of the response: https://github.com/psviderski/uncloud/pull/159

psviderski:
I like the addition of `nearest.{service}.internal` and also having an explicit `rr.{service}.internal`.

But I still struggle to figure out what makes the most sense for the default `{service}.internal`. I believe peer discovery for things like a NATS cluster is rather an edge case. I guess the most common use case is a regular service A accesses another regular web service B (replicated).

My intuition tells that minimising latency by prioritising local replicas makes sense. Currently, if there is no DNS resolution caching is implemented in service A, it will distribute requests among all available replicas by resolving the name to a randomised list of IPs. This could be beneficial in the case where the local replica is overloaded or misbehaving but doesn't have a health check configured.

While researching round-robin DNS, I found some interesting posts:
* https://medium.com/@lherrera/poor-mans-load-balancing-with-docker-2be014983e5
  * Describes default address selection from RFC3484 that prevents a normal round robin using DNS. However, I tried their lab https://github.com/lherrera/lb and couldn't reproduce the described behaviour. All clients requests were evenly distributed among all available servers. I'm not sure what has changed (Linux?, musl?, glibc?) since the post was written (9 years ago).
[...]

Issue #64, opened 2025-05-23
Create CLI command to inspect and troubleshoot WireGuard network


psviderski:
Adding a new machine to the cluster may fail due to Corrosion not being able to establish a connection to other peers. The main reason for this is usually WireGuard not being able to establish a tunnel between machines.

```
...
✓ Uncloud machine daemon started.
✓ Uncloud installed on the machine successfully! 🎉
Machine 'uc-prod-ap1' added to the cluster (context 'uc-prod').
Waiting for the machine to be ready...
Error: wait for cluster to be initialised on machine: rpc error: code = FailedPrecondition desc = cluster is not initialised
```

`uncloud-corrosion` service logs on `uc-prod-ap1`:
```
May 23 04:06:29 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:29.029051Z ERROR corro_agent::transport: error=deadline has elapsed
May 23 04:06:29 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:29.029077Z ERROR corro_agent::transport: error=deadline has elapsed
May 23 04:06:29 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:29.029081Z ERROR corro_agent::agent::handlers: could not write datagram [fdcc:259e:b5d0:d11b:b9d:27f:fdc3:3a17]:51001:>
May 23 04:06:34 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:34.030117Z ERROR corro_agent::transport: error=deadline has elapsed
May 23 04:06:34 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:34.030147Z ERROR corro_agent::transport: error=deadline has elapsed
May 23 04:06:34 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:34.030151Z ERROR corro_agent::agent::handlers: could not write datagram [fdcc:259e:b5d0:d11b:b9d:27f:fdc3:3a17]:51001:>
May 23 04:06:46 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:46.131217Z ERROR corro_agent::transport: error=deadline has elapsed
May 23 04:06:46 vps1 uncloud-corrosion[1016597]: 2025-05-23T04:06:46.131257Z ERROR corro_agent::transport: error=deadline has elapsed
```

[...]

Issue #32, opened 2025-02-27
Command to list caddy servers and their upstreams with health status


psviderski:
It's a nice idea to list upstreams for served hosts and their health status that should be helpful for troubleshooting. The idea is borrowed from [stellar](https://www.youtube.com/watch?v=diRmI45Yhg4):
<img width="651" alt="Image" src="https://github.com/user-attachments/assets/a260c0b7-4789-4f4b-812f-4a1b490907ac" />

miekg:
wouldn't metrics (also) solve this and be more generic?

cmdline interface is still nice idea as well, but pushing that idea further; the above can then also be shown, but retrieved by asking the in-cluster promtheus server (if we want to have that)

psviderski:
Maybe if the caddy `/metrics` endpoint can provide all the required data to list all hosts and their upstreams. Then the rpc call will query `/metrics` instead of the `/reverse_proxy/upstreams` endpoint as it's done in https://github.com/psviderski/uncloud/pull/218. Not a big difference tbh.

But I wouldn't artificially add a dependency on the in-cluster Prometheus to serve this data for the CLI if we can easily get them without Prometheus.

I think we shouldn't require every setup to deploy Prometheus in the cluster because it's too heavy. It's also quite often the monitoring stack is deployed separately and shared by multiple clusters/environments so it feels uncloud shouldn't have an opinion here.

PR #159, opened 2025-10-27
feat: Adding a "nearest.service.local" mode to internal DNS

jabr:
Related issue: https://github.com/psviderski/uncloud/issues/154

Added "nearest" internal DNS mode the puts machine-local IPs at the start of responses. Also added a "rr" mode that has the same, random shuffle behavior of the current default.

I've tested this build on my cluster.

### DIFF
--- a/website/docs/3-concepts/6-services/1-internal-dns.md
+++ b/website/docs/3-concepts/6-services/1-internal-dns.md
@@ -59,3 +59,75 @@ Address: 10.210.1.3
 Name:   worker.internal
 Address: 10.210.1.4
 ```
+
+## IP Ordering Mode
+
+Additionally, the IP ordering preference can be specified with a `rr` (round-robin) or `nearest` subdomain prefix.
+
+### `rr` (round-robin) *current default*
+Randomly shuffled order on each lookup.
+
+```
+$ nslookup rr.worker.internal
+Server:         127.0.0.11
+Address:        127.0.0.11#53
+
+Name:   rr.worker.internal
+Address: 10.210.0.3
+Name:   rr.worker.internal
+Address: 10.210.0.4
+Name:   rr.worker.internal
+Address: 10.210.1.3
+Name:   rr.worker.internal
+Address: 10.210.1.4
+
+$ nslookup rr.worker.internal
+Server:         127.0.0.11
+Address:        127.0.0.11#53
+
+Name:   rr.worker.internal
+Address: 10.210.0.4
+Name:   rr.worker.internal
+Address: 10.210.1.3
+Name:   rr.worker.internal
+Address: 10.210.1.4
+Name:   rr.worker.internal
+Address: 10.210.0.3
+```
+
+## Nearest scope
+Returns machine-local instances first.
+
+`machine-a`:
+```
+$ nslookup nearest.worker.internal
+Server:         127.0.0.11
+Address:        127.0.0.11#53
+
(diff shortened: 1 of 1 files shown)


internal/machine/dns/server.go
```diff
@@ -53,7 +57,8 @@ type Server struct {
 // NewServer creates a new DNS server with the given configuration.
 // If upstreams is nil, nameservers from /etc/resolv.conf will be used. An empty upstreams list means to only resolve
 // internal DNS queries and not forward any external queries.
-func NewServer(listenAddr netip.Addr, resolver Resolver, upstreams []netip.AddrPort) (*Server, error) {
+func NewServer(localSubnet netip.Prefix, resolver Resolver, upstreams []netip.AddrPort) (*Server, error) {
+	var listenAddr = network.MachineIP(localSubnet)
```
psviderski: I think we're better off passing both `listenAddr` and `localSubnet` to the DNS server to avoid coupling it with the internal implementation of uncloud address assignment. The decision to listen on the machine IP (first address) in a subnet shouldn't be the responsibility of the DNS server, IMO.

This won't require a dependency on the internal network module and will keep the server independent. Later, instead of `localSubnet` we might want to pass a comparator function or an interface that would handle the IP sorting for the server, again keeping it independent of the internal details of uncloud networking.

jabr: Good point. Updated.

I thought about how this might be extended to rank IPs based on proximity (ping latency between machines?), but I decided to just keep it simple for now and start with the "same machine" test based on the subnet.

I think we'd need to figure out what exactly we want other info we'd sort on (and how we'd get it) for proximity first, and then figure out how the code interface would best work.

jabr:
Just trying out an idea, but I also added a "rr" mode with an explicit round-robin (rather than random) based on a count of total lookups the server has resolved so far (so not _really_ round-robin, I guess, as it's not specific to that exact name...). The default behavior is still just random shuffle. I'll follow up in #154.

internal/machine/dns/server.go
psviderski: Improving this iteratively sounds perfect.

I also have an idea (stole from abandoned project Stellar) to show the latency for Caddy upstreams: https://github.com/psviderski/uncloud/issues/32. I think their implementation was very naive (ping essentially).

It doesn't seem that the WireGuard Peer status struct contains any latency-like data, only transmitted/received bytes and latest handshake. If we were using Surf instead of Corrosion, we could use its built-in coordinates capability to sort machines based on RTT: https://github.com/hashicorp/serf/blob/master/docs/internals/coordinates.html.markdown.

I think we can later introduce a component in an uncloud daemon that would periodically ping each machine and store last N data points, or maybe only a sliding average/mean. Then we can display this data in a network troubleshooting command: https://github.com/psviderski/uncloud/issues/64 and also use for sorting the IP addresses for `nearest.*.internal` records.

psviderski:
I'm not sure if the approach based on `resolvedCount` is any different to the random one in the case where there are many local services making many requests. I think random should statistically provide even distribution which IMO is good enough. Note that Caddy's default LB policy is also random.

jabr:
1. Removed the count based rotation mode for `rr` and just made it use the (current) default of random shuffle alone
2. Added a test on the nearest ordering
3. Added docs on the `rr` and `nearest` prefixes

I think it's good to go now, but let me know if you have any notes.

PR #160, opened 2025-10-27
feat: Allow to match containers against truncated ID

tonyo:
@psviderski thoughts on this? Lmk if there was no intention to do any "smarter" kind of matching on this API level.

This would slightly simplify it for https://github.com/psviderski/uncloud/pull/139/ (since people will only have access to the short ID and that's what they'll be pasting to `exec --container`, but no big deal to write a small wrapper there. 

Alternatively: I can add another function like InspectContainerFuzzy or similar.

### DIFF
--- a/pkg/client/container.go
+++ b/pkg/client/container.go
@@ -209,6 +209,7 @@ func (cli *Client) InspectContainer(
 		return api.MachineServiceContainer{}, fmt.Errorf("inspect service: %w", err)
 	}
 
+	// TODO: support matching by any prefix of the container ID
 	for _, c := range svc.Containers {
 		if c.Container.ID == containerNameOrID ||
 			c.Container.Name == containerNameOrID ||


pkg/client/container.go
```diff
@@ -208,7 +210,9 @@ func (cli *Client) InspectContainer(
 	}
 
 	for _, c := range svc.Containers {
-		if c.Container.ID == containerNameOrID || c.Container.Name == containerNameOrID {
+		if c.Container.ID == containerNameOrID ||
+			c.Container.Name == containerNameOrID ||
+			stringid.TruncateID(c.Container.ID) == containerNameOrID {
```
psviderski, line 216: This looks good but we can go even further and support any ID prefix as docker does. Something like
```go
strings.HasPrefix(c.Container.ID, containerNameOrID)
```
But in this case we shouldn't return right after the first match but instead compare the prefix against all containers. And return only if the prefix matches only one unique container.

Then a user will be able to type just a few first chars of the container id `uc exec app --container ea0`
It's up to you if you want to implement this right away. Otherwise supporting only 12-chars short IDs is fine as well.

tonyo: Sounds good, added a TODO for that. Will need to think a bit more about edge cases, like a container's name being a prefix for some container id, or similar.

Issue #119, opened 2025-09-10
[BUG] Unable to connect to cluster if first machine is down


luislavena:
**Describe the bug**

When attempting to `machine ls` or `service ls` fails when the first node in the cluster is down:

```
$ uc --uncloud-config ./blatta.yaml service ls
Error: connect to cluster: connect to cluster (context 'default'): connect to machine: SSH login to provision@192.168.1.203:22: connect using SSH agent: dial tcp 192.168.1.203:22: i/o timeout
```

**How to reproduce**

Local configuration:

```yaml
current_context: default
contexts:
  default:
    connections:
      # blatta1
      - ssh: provision@192.168.1.203
      # blatta2
      - ssh: provision@192.168.1.194
      # blatta3
      - ssh: provision@192.168.1.246
```

1. Setup 2/3 nodes cluster and ensure all nodes are listed in `connections` of uncloud configuration
2. Verify services and machines are listed and running
3. Turn off/poweroff the first node of the list in the cluster
4. Attempt to repeat step number 2 and notice the timeout failure
5. Comment out the first connection and try again, this time it works.

**Expected behavior**

I would expect `uc` to retry by connecting any of the servers listed under `connections` instead of failing at the first one.

**Environment:**

- Uncloud versions:
  - Control (client) node (`uc --version`): `0.12.1` (darwin/arm64)
  - Uncloud daemon (from the server) (`uncloudd --version`): 0.12.1 (linux/amd64)
- OS version (`uname -a`):
  - Client (control node): Darwin Kernel Version 24.6.0
  - Server: Debian 6.12.43-1 (2025-08-27) x86_64 GNU/Linux

---

Thank you
❤️ ❤️ ❤️

psviderski:
Thank you for reporting this with the details! Yeah, we're currently just trying the first connection: https://github.com/psviderski/uncloud/blob/d25864e52f2cb3cb4eb7d29ac4f6f58a2b62a3f8/internal/cli/cli.go#L119-L120 which definitely needs to be addressed.

>I would expect uc to retry by connecting any of the servers listed under connections instead of failing at the first one.

There are multiple options how we can handle retries:
1. Try each connection in order until one succeeds. The downside is that `uc` commands will become slow if the first connection(s) times out.
2. Always try to connect using multiple connections (random? first n?) and use the first successful.

Some optiomisations:
- (1) + tracking the last successful connection and trying it first **<-- sounds like the most straightforward and intuitive approach**
[...]

PR #162, opened 2025-11-03
feat: [draft] Add name property to config connections for use in `--connect {name}` option

jabr:
Adds an alias to the connections config to use in `--connect` option.

It also stores the name with the connection info when adding a machine, updates the name when renaming the machine, and removes the connection entry by name when removing the machine.

Just cli changes which are working for me locally in `--connect` usage, but I have not testing the add/rename/remove bit. (We could remove the auto-update parts for an initial version, too. Even just being able to add manual names/labels is pretty convenient on its own.)

### DIFF
--- a/cmd/uncloud/machine/rm.go
+++ b/cmd/uncloud/machine/rm.go
@@ -146,6 +146,20 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
 	}
 	fmt.Printf("Machine '%s' removed from the cluster.\n", m.Name)
 
+	// Remove the connection to the machine from the uncloud config if it exists.
+	if uncli.Config != nil {
+		context := uncli.Config.Contexts[uncli.Config.CurrentContext]
+		for i, c := range context.Connections {
+			if c.Name == m.Name {
+				context.Connections = slices.Delete(context.Connections, i, i+1)
+				break
+			}
+		}
+		if err := uncli.Config.Save(); err != nil {
+			return fmt.Errorf("save config: %w", err)
+		}
+	}
+
 	if reset && reachable {
 		_, err = client.MachineClient.Reset(mctx, &pb.ResetRequest{})
 		if err != nil {
@@ -154,9 +168,6 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
 			fmt.Println("Machine reset initiated and will complete in the background.")
 		}
 	}
-
-	// TODO: remove the connection to the machine from the uncloud config if it exists. We need a way to associate
-	//  the machine with its connection in the config, e.g. by storing the machine name in the connection metadata.
 
 	// TODO: If Caddy was running on this machine and a cluster domain is reserved,
 	//  let the user know that the DNS records should be updated.
--- a/cmd/uncloud/main.go
+++ b/cmd/uncloud/main.go
@@ -40,6 +40,7 @@ func main() {
 			cli.BindEnvToFlag(cmd, "connect", "UNCLOUD_CONNECT")
 			cli.BindEnvToFlag(cmd, "uncloud-config", "UNCLOUD_CONFIG")
 
+			configPath := fs.ExpandHomeDir(opts.configPath)
 			var conn *config.MachineConnection
 			if opts.connect != "" {
 				if strings.HasPrefix(opts.connect, "tcp://") {
@@ -55,18 +56,34 @@ func main() {
 					conn = &config.MachineConnection{
 						SSHCLI: config.SSHDestination(dest),
 					}
-				} else {
-					dest := opts.connect
-					if strings.HasPrefix(dest, "ssh://") {
-						dest = dest[len("ssh://"):]
-					}
+				} else if strings.HasPrefix(opts.connect, "ssh://") {
+					dest := opts.connect[len("ssh://"):]
 					conn = &config.MachineConnection{
 						SSH: config.SSHDestination(dest),
+					}
+				} else {
+					// If the connect string doesn't contain "@", it could be a machine name.
+					if !strings.Contains(opts.connect, "@") {
+						cfg, err := config.NewFromFile(configPath)
+						if err == nil {
+							for _, c := range cfg.Contexts[cfg.CurrentContext].Connections {
+								if c.Name == opts.connect {
+									conn = &c
+									break
+								}
+							}
+						}
+					}
+
+					// Default to assuming it is an ssh destination.
+					if conn == nil {
+						conn = &config.MachineConnection{
+							SSH: config.SSHDestination(opts.connect),
+						}
 					}
 				}
 			}
 
-			configPath := fs.ExpandHomeDir(opts.configPath)
 			uncli, err := cli.New(configPath, conn)
 			if err != nil {
 				return fmt.Errorf("initialise CLI: %w", err)
--- a/cmd/uncloud/machine/rename.go
+++ b/cmd/uncloud/machine/rename.go
@@ -42,6 +42,20 @@ func rename(ctx context.Context, uncli *cli.CLI, contextName, oldName, newName s
 		return fmt.Errorf("rename machine: %w", err)
 	}
 
+	// Update the machine name in the config file.
+	if uncli.Config != nil {
+		context := uncli.Config.Contexts[uncli.Config.CurrentContext]
+		for i, c := range context.Connections {
+			if c.Name == oldName {
+				context.Connections[i].Name = newName
+				break
+			}
+		}
+		if err := uncli.Config.Save(); err != nil {
+			return fmt.Errorf("save config: %w", err)
+		}
+	}
+
 	fmt.Printf("Machine %q renamed to %q (ID: %s)\n", oldName, machine.Name, machine.Id)
 	return nil
 }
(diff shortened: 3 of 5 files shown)


cmd/uncloud/main.go
```diff
@@ -39,21 +39,37 @@ func main() {
 
 			var conn *config.MachineConnection
 			if opts.connect != "" {
-				if strings.HasPrefix(opts.connect, "tcp://") {
-					addrPort, err := netip.ParseAddrPort(opts.connect[len("tcp://"):])
-					if err != nil {
-						return fmt.Errorf("parse TCP address: %w", err)
+				// If the connect string doesn't contain "@" or "tcp://", it could be a machine name.
+				if !strings.Contains(opts.connect, "@") && !strings.Contains(opts.connect, "tcp://") {
```
jabr: I think this should check for `ssh://` (and now `ssh+cli://`), too. Maybe any "uri protocol prefix"? And if there is some restricted character set for machine names (alphanum, hypen, underscore, dot?), use that instead of just the `@`?

cmd/uncloud/main.go
```diff
@@ -55,18 +56,34 @@ func main() {
...
-					dest := opts.connect
-					if strings.HasPrefix(dest, "ssh://") {
-						dest = dest[len("ssh://"):]
-					}
+				} else if strings.HasPrefix(opts.connect, "ssh://") {
+					dest := opts.connect[len("ssh://"):]
 					conn = &config.MachineConnection{
 						SSH: config.SSHDestination(dest),
 					}
+				} else {
```
psviderski, line 64: I have mixed feeling about this change. Originally, `--connect` was added to connect using an explicitly specified connection string without requiring the config:
```
      --connect string          Connect to a remote cluster machine without using the Uncloud configuration file. [$UNCLOUD_CONNECT]
```

The separation was IMO clear, but now `--connect` may involve the config as well so I'm trying to figure out if we may potentially confuse users with this 🤔 How would you explain `--connect` now?

Most commands also define the `--context` flag which we should have made global long time ago. With the `--context` specified, this branch using the current context the config will be misleading.

I think we shouldn't change the current behaviour of `--connect` to keep things simple.

What scenarios do you have on your mind where you would want to manually specify what connection to use for a specific command? Maybe we should describe some strong use cases first to work out what the best UX is.

From my personal experience so far, I had to manually switch connections only when I try to access my home cluster when I'm not at home (rarely). I described the use case here: https://github.com/psviderski/uncloud/issues/119#issuecomment-3284058940. But in this case, I need to change the default once and then use it for the entire session until I'm back home. For my use case, having an interactive command for changing the default connection or a non-interactive one like `uc ctx connection use <name>/<index>` will be enough.

If for some reason one need to run a one-off command connected to specific machine, I think specifying its SSH target should be fine. More over, @luislavena introduced a new `ssh+cli` way to connect to machines https://github.com/psviderski/uncloud/pull/152 which we plan to make the default. It will support any configuration and shortcuts defined in `.ssh/config` config.

So if one defines something like this in `.ssh/config`:
```
Host myserver
    HostName 192.168.100.200
    User myuser
```
they can ssh to it using just `ssh myserver`. This will automatically become supported by uncloud:
```
uc machine init myserver
uc machine add myserver
uc --connect myserver COMMAND
```

I think this path of supporting `.ssh/config` looks more versatile as it will work with both regular `ssh` and uncloud the same way.

Please let me know what you think @jabr @luislavena

cmd/uncloud/machine/rm.go
```diff
@@ -146,6 +146,20 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
 	}
 	fmt.Printf("Machine '%s' removed from the cluster.\n", m.Name)
 
+	// Remove the connection to the machine from the uncloud config if it exists.
+	if uncli.Config != nil {
+		context := uncli.Config.Contexts[uncli.Config.CurrentContext]
```
psviderski, line 151: As I mentioned in another [comment](https://github.com/psviderski/uncloud/pull/162/files#r2497955547), we should use the manually specified context (`opts.context`) if specified and fall back to the current one.

psviderski (commented):
Overall, I like the idea of being able to name/label the connections 👍 
Please see the comment discussing the change in `--connect` behaviour. Let me know what you think

cmd/uncloud/main.go
jabr: The inspiration to add it came up when working on the wireguard command: for that tool, you really need to be connecting directly to the machine you want to inspect, as proxying the gRPC call is likely not going to work when you actually need to use that command (i.e. the machines aren't talking properly so they can't proxy the call).

So I wanted to use the `--connect` option for that, but I couldn't because my ssh connection strings are complicated:

1. I have a separate keypair for my uncloud machines, so just doing `--connect user@host` wasn't enough -- I would need to specify the key somehow, too. (Maybe the `-i` option works with `--connect`? I didn't think to try that at the time.)

2. The user names for the different machines are a bit odd (cloud host provisioned "system" accounts which are different on some machines and have random numbers for some reason).

3. I'd need to configure `/etc/hosts` aliases for the IPs.

So I was just editing the `config.yaml` to switch the default whenever I _actually_ wanted to use the `--connect` option.

However, the `ssh+cli` option now will make that a lot better (I do have them configured properly in my `~/.ssh/config`) so it'll just be `--connect ssh+cli://{machine-name}` for me.

For what it's worth, I'd tried that `--connect {machine-name}` approach early on when starting with uncloud, expecting it to work like this. I might be the only one, but it was a bit just ux/dx polish, imo -- it felt like that "should work" to me. 🤷

Maybe it should just be logic in the wireguard command, like `uc wg show <machine>`, since that was the most pertinent use case I had? Or another "protocol prefix" like `tcp:`, `ssh+cli:`, etc in the connect string? Another option flag to not overload `connect` and confuse the interaction with `context`?

Anyway, I'm not hard set on anything here. Just going through my thought process on why I ended up with it like this. 😄

jabr: Another thought I had last night:

The one thing in this branch that I think we're likely all agreed is useful is removing the connection entry for a machine when the machine itself is removed from the cluster. (There was a TODO note in the code about doing that.)

To do that, we'll need _something_ to connect the entry to the machine itself, but perhaps that should just be the machine ID. And then this branch (or another branch/PR might be better for a clean start) just does the add/remove bits of this but with the ID not the name (and machine rename no longer being relevant).

I'm not sure the "named connection" is then even necessary with Luis's `ssh+cli` option now, but if we decide to do something like it, we can do that as a separate issue, which should be able to build on the machine ID association we add here.

What do you think?

psviderski: I really appreciate you shared your thought process! This all totally makes a lot of sense

>For what it's worth, I'd tried that --connect {machine-name} approach early on when starting with uncloud, expecting it to work like this. I might be the only one, but it was a bit just ux/dx polish, imo -- it felt like that "should work" to me. 🤷

I'm pretty sure you're not the only one and I'd personally love to be able to make this work like that reliably.
We could easily do that if we had a central authority we can talk to (e.g. a cloud service each machine connects to). But we're in a bit weird position where there is no single place that would maintain an up-to-date information about the machines in the cluster and connections to them.

Slightly tangential to this PR, but I'm planning on creating a managed public service that would optionally allow connecting machines to to simplify access and management. In this case, the connection may look like this (no ssh):
```
contexts:
  managed-cluster-access:
    connections:
      - hub: hub.uncloud.run
        token: <secret>
```
But this will only complement the current ssh options. So the problem with managing ssh connections still exists.

>The one thing in this branch that I think we're likely all agreed is useful is removing the connection entry for a machine when the machine itself is removed from the cluster. (There was a TODO note in the code about doing that.)

Agreed

>To do that, we'll need something to connect the entry to the machine itself, but perhaps that should just be the machine ID. And then this branch (or another branch/PR might be better for a clean start) just does the add/remove bits of this but with the ID not the name (and machine rename no longer being relevant).

Using machine ID sounds like the most "correct" (more stable) approach we considered so far. Although, not the most user-friendly imo. Note also that this won't save us from getting a stale config when there are multiple cluster users and one of them removes a machine. We can do our best effort but this still won't be a complete solution anyway unfortunately. I also don't expect that adding/removing machines is done very often so could not be a big of a problem.

>I'm not sure the "named connection" is then even necessary with Luis's ssh+cli option now, but if we decide to do something like it, we can do that as a separate issue, which should be able to build on the machine ID association we add here.

I would love to make the @luislavena's work on ssh+cli the default first (assume `ssh+cli://` prefix if none is specified) and see how it works. The ssh config provides so many different ways of accessing machine and it would be amazing to piggyback on this. Maybe it will cover all the needs in practice so we won't need to do anything custom on our own.

jabr: > I would love to make the @luislavena's work on ssh+cli the default first (assume ssh+cli:// prefix if none is specified) and see how it works.

Yeah, I like that approach. It would also have "just worked" for me with my initial expectations that way, too, since I have `~/.ssh/config` entries with the same aliases as my machine names.

I'd like to close this PR now, but I'd be happy to make new ones for either of these if you want to pursue them now:
1) making `ssh+cli` the default over `ssh`
2) adding machine id to the connection config entries on `machine add` and removing corresponding ones on `machine rm`

psviderski: >1. making ssh+cli the default over ssh

We can't make it the default until we address the issue described here https://github.com/psviderski/uncloud/pull/173#issuecomment-3510272405. It would also be good to test it for some time for the practical scenarios before switching the default.

>2. adding machine id to the connection config entries on machine add and removing corresponding ones on machine rm

Let's do this one. This approach seems to be the most pragmatic of what we've discussed

jabr: PR for storing Machine ID in corresponding connection (and removing with machine): #182

Closing this one.

Issue #146, opened 2025-10-15
[BUG] uc image push -m all Ignores Platform Compatibility


pilasguru:
### Environment

Software Affected: Uncloud

Version: v0.13.1

Cluster Configuration: 3 nodes (1 arm64 - rita, 2 amd64 - xuxa, amelie).

Source Image: `busybox:latest` (Platform: linux/arm64/v8)

### 👟 Steps to Reproduce

Start with the cluster configuration described above. The image `busybox:latest` is only present on the arm64 node (rita) and is PLATFROM linux/arm64/v8.

Execute the push command:

```
uc image push -m all
```

### 🚫 Actual Behavior

The command uc image push -m all forces the image to be pushed to all three nodes (rita, xuxa, and amelie), ignoring the platform incompatibility of the image (linux/arm64/v8) with the amd64 nodes (xuxa and amelie).

### ✅ Expected Behavior

The `uc image push -m all` command should respect platform compatibility. The push operation should:

Succeed on the compatible arm64 node (rita).

Fail or Skip the incompatible amd64 nodes (xuxa and amelie), or at least issue a clear warning about the platform mismatch.

### ℹ️ Additional Information

The primary concern is that incompatible images are silently pushed to nodes where they cannot run, consuming unnecessary disk space.

psviderski:
Thank you so much for raising this concern! I did think about implementing at least a warning before the release of the push feature but decided to ship it as-is and see how people will use it first. Please see https://github.com/psviderski/uncloud/issues/137

>The primary concern is that incompatible images are silently pushed to nodes where they cannot run, consuming unnecessary disk space.

100% agree, the tool should definitely guide users to do the right things and prevent from shooting in the foot.

From the implementation point of view, I think issuing a warning or skipping incompatible machines cost the same.

As you suggested, I think the the most intuitive approach would be:
- push to all compatible machines
- skip incompatible machines and print a clear message for them describing the platform mismatch

It should be clear for a user why an image hasn't been pushed to particular machines even though they specified `-m all`.

JFYI, I'm also thinking about adding more advanced machine selectors based on their properties. For example, select all machines with certain platform `-m platform=linux/amd64` or all machines that run `myservice` containers `-m service=myservice` (this might be useful for rolling deployments to only push to machines where the containers running the old one are). Not sure about the syntax yet, maybe something similar to [Swarm's constraints](https://docs.docker.com/reference/cli/docker/service/create/#constraint).

PR #164, opened 2025-11-04
[draft] feat: Add labels on machines (with cli for manual add/rm/ls)

jabr:
Initial draft at implementing machine labels (#5).

For context, this is more of a proposed building block for future features based on the labels. I thought a draft/spike might be useful to think through the approach to ensure it will work for everything we want to do with them.

For example, how would we incorporate auto-generated or implicit labels, like `ingress`, platform (`arm64`/`amd64`), etc. Are those actually materialized labels in this set, or inserted dynamically by the API returning the set? And if materialized, due we need some namespacing standard to distinguish internal and user labels? `i:ingress`, `i:arm64`, `u:eu`, `u:worker`, etc?


internal/machine/api/pb/cluster.proto
```diff
@@ -18,6 +18,8 @@ service Cluster {
   rpc GetDomain(google.protobuf.Empty) returns (Domain);
   rpc ReleaseDomain(google.protobuf.Empty) returns (Domain);
   rpc CreateDomainRecords(CreateDomainRecordsRequest) returns (CreateDomainRecordsResponse);
+  rpc AddMachineLabels(MachineLabelsRequest) returns (UpdateMachineResponse);
+  rpc RemoveMachineLabels(MachineLabelsRequest) returns (UpdateMachineResponse);
```
psviderski: What do you think about managing machine labels as part of `rpc UpdateMachine(UpdateMachineRequest)`?
Unless I'm missing something, it seems labels is just a property similar to `name` or `endpoints`.

psviderski (commented):
>how would we incorporate auto-generated or implicit labels, like ingress, platform (arm64/amd64), etc. Are those actually materialized labels in this set, or inserted dynamically by the API returning the set? And if materialized, due we need some namespacing standard to distinguish internal and user labels? i:ingress, i:arm64, u:eu, u:worker, etc?

Tbh, I haven't thought much about the labels design yet. I was delaying it until we have to implement a feature that requires them. This way we can understand the requirements better and likely make better decisions.

I was thinking that we might want labels to be a map (key/value pairs), not a list of strings. This will be more in line with the docker labels design and the [`labels`](https://github.com/compose-spec/compose-spec/blob/main/spec.md#labels) configuration in the Compose spec.

Regarding materialisation, I think that materialising auto-generated labels such as machine arch makes it easier to query them later and integrate in the workflows such as container scheduling. We can have a logic in the `uncloudd` daemon to update its own labels in the cluster store on the daemon start or on schedule.

I don't have any strong opinion on the label format yet. Maybe we shouldn't enforce any at least for now. Like in Docker and k8s there could be a naming convention with some reserved prefix(es) which isn't strictly enforced: https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations
>The com.docker.*, io.docker.*, and org.dockerproject.* namespaces are reserved by Docker for internal use.

internal/machine/api/pb/cluster.proto
jabr: I did it as a field on the `UpdateMachine` request initially, but it felt a bit incongruous that way. The other fields are "set" operations, replacing the corresponding machine value, but the label commands felt like they should be "add"/"remove" individual labels to/from the existing set.

So assuming that intuition is true (the workflow will typically be add/remove, not set), matching the "set" behavior of the rest of `UpdateMachine` meant doing a fetch all, add/remove locally, then push the whole set back, and that felt wrong, too.

Then I considered `labelsToAdd` and `labelsToRemove` in the `UpdateMachineRequest` and that felt like a mismatch with the rest of `internal/machine/cluster/cluster.go#UpdateMachine` as well as making it a pretty long function.

So I ended up with the current form and two new methods, which I felt had clearer semantics that matched the new cli commands and expected workflow better.

Happy to change if you think there's a better approach, but what are your thoughts on the above?

jabr:
> I was delaying it until we have to implement a feature that requires them. This way we can understand the requirements better and likely make better decisions.

The ingress (#8), arch/platform (#146), and general placement constraint (#3) issues related to push/deploy all struck me as something concrete that would benefit from having them now.

It also seems like a feature that will find more uses in other ways once we have it, too, like filtering service/container/machine list based on them. Sort of a "build it and they will come" situation. 😆 

> I was thinking that we might want labels to be a map (key/value pairs), not a list of strings. This will be more in line with the docker labels design and the [labels](https://github.com/compose-spec/compose-spec/blob/main/spec.md#labels) configuration in the Compose spec.

That's a good point. I'm not super familiar with docker labels, and I was thinking of them more like "tags". The map approach solves my namespacing (internal vs user) question, too.

One clarification, though: are docker labels a _map_ or a _multimap_? Can a key, say `role`, have more than one value associated? Like `role=webapp` and `role=worker`?

That was one of the reasons I went with the simple list of "tags" approach, so a user could do something like `:role/webapp` and `:role/worker` on a machine (using whatever structure they wanted in the tags themselves).

And actually, even the docker labels form could fit in the simple "tag" list, too. It looks like the docker compose spec even supports that syntax:
```
labels:
  - "com.example.description=Accounting webapp"
  - "com.example.department=Finance"
  - "com.example.label-with-empty-value"
```

Wondering if there's any real benefit to having a map data structure underneath it... I suppose some lookup value by key use case? Though that's the same as a "namespace" prefix filter...

psviderski:
>One clarification, though: are docker labels a map or a multimap? Can a key, say role, have more than one value associated? Like role=webapp and role=worker?

It's a map, can't have duplicate keys. 

>That was one of the reasons I went with the simple list of "tags" approach, so a user could do something like :role/webapp and :role/worker on a machine (using whatever structure they wanted in the tags themselves).

Docker uses an empty value in this case, see https://docs.docker.com/reference/cli/docker/container/run/#label

>Wondering if there's any real benefit to having a map data structure underneath it... I suppose some lookup value by key use case?

Yeah, I think it's just a bit more convenient to not deal with string processing and having an implicit convention on the separator when doing key lookups. I believe most of the time labels will be key/values any way. For example, used for placement constraints https://docs.docker.com/reference/cli/docker/service/create/#constraint

jabr:
Updated to use key-value map for labels instead of simple list.

miekg:
volumes already support Docker-style labels as do images when build with the `LABEL` command.

For prometheus support (#304) labels would be nice, as I can then autodetect what and where need to be scaped:
`uncloud.prometheus:/metrics` or some such

Issue #56, opened 2025-05-01
New 'uc ps' command to list all service containers in the cluster


psviderski:
`uc ls` lists all services in the cluster. `uc service inspect` can be used to list all containers for a particular service.
Sometimes, it might be helpful to list all containers for all service in the cluster including the machines they run on to troubleshoot service distribution and placement, monitor resource utilisation, or check container health for the entire cluster.

jabr:
I think we can close this with #165 merged.

psviderski:
Thank you! ❤️

PR #165, opened 2025-11-04
feat: Add `uc ps` command to list all containers in the cluster

jabr:
Implementation for #56.

- Sorts by service, machine, or "health"
- Has color highlight on broad categories of status/state (healthy=green/running=normal/unhealthy=red/other=yellow).
- Uses "working" spinner while collecting container info.

### DIFF
--- a/cmd/uncloud/ps.go
+++ b/cmd/uncloud/ps.go
@@ -199,18 +199,29 @@ func collectContainers(ctx context.Context, cli *client.Client) ([]containerInfo
 
 	var containers []containerInfo
 	for _, msc := range machineContainers {
-		if msc.Metadata == nil {
-			continue
-		}
+		machineName := "unknown"
 
-		machineName, ok := machinesNamesByIP[msc.Metadata.Machine]
-		if !ok {
-			machineName = msc.Metadata.Machine
+		// Metadata can be nil if the request was broadcasted to only one machine.
+		if msc.Metadata == nil && len(machineContainers) > 1 {
+			return nil, fmt.Errorf("something went wrong with gRPC proxy: metadata is missing for a machine response")
 		}
-
-		if msc.Metadata.Error != "" {
+		if msc.Metadata != nil && msc.Metadata.Error != "" {
 			client.PrintWarning(fmt.Sprintf("failed to list containers on machine %s: %s", machineName, msc.Metadata.Error))
 			continue
+		}
+
+		if msc.Metadata != nil {
+			var ok bool
+			machineName, ok = machinesNamesByIP[msc.Metadata.Machine]
+			if !ok {
+				// Fallback to machine's IP as name.
+				machineName = msc.Metadata.Machine
+			}
+		} else {
+			// Fallback to the first available machine name.
+			if len(machines) > 0 {
+				machineName = machines[0].Machine.Name
+			}
 		}
 
 		for _, ctr := range msc.Containers {
--- a/cmd/uncloud/ps_test.go
+++ b/cmd/uncloud/ps_test.go
@@ -1,0 +1,424 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"net/netip"
+	"testing"
+
+	"github.com/psviderski/uncloud/internal/machine/api/pb"
+	"github.com/psviderski/uncloud/internal/machine/docker"
+	"github.com/psviderski/uncloud/pkg/client"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+// mockDockerClient implements pb.DockerClient
+type mockDockerClient struct {
+	pb.DockerClient // Embed to avoid implementing all methods
+	listResp        *pb.ListServiceContainersResponse
+	listErr         error
+}
+
+func (m *mockDockerClient) ListServiceContainers(ctx context.Context, in *pb.ListServiceContainersRequest, opts ...grpc.CallOption) (*pb.ListServiceContainersResponse, error) {
+	return m.listResp, m.listErr
+}
+
+// mockClusterClient implements pb.ClusterClient
+type mockClusterClient struct {
+	pb.ClusterClient // Embed to avoid implementing all methods
+	machinesResp     *pb.ListMachinesResponse
+	machinesErr      error
+}
+
+func (m *mockClusterClient) ListMachines(ctx context.Context, in *emptypb.Empty, opts ...grpc.CallOption) (*pb.ListMachinesResponse, error) {
+	return m.machinesResp, m.machinesErr
+}
+
+func TestCollectContainers_NilMetadata(t *testing.T) {
+	// Setup container data
+	containerData := map[string]interface{}{
+		"Id":   "container1",
+		"Name": "test-container",
+		"Config": map[string]interface{}{
+			"Image": "test-image",
+		},
+		"State": map[string]interface{}{
+			"Status":     "running",
+			"StartedAt":  "2023-01-01T12:00:00Z",
+			"FinishedAt": "0001-01-01T00:00:00Z",
+		},
+	}
+	containerJSON, err := json.Marshal(containerData)
+	require.NoError(t, err)
+
+	serviceSpecJSON, err := json.Marshal(map[string]interface{}{})
+	require.NoError(t, err)
+
+	// Setup mocks
+	mockDocker := &mockDockerClient{
+		listResp: &pb.ListServiceContainersResponse{
+			Messages: []*pb.MachineServiceContainers{
+				{
+					Metadata: nil, // Simulating the issue: nil metadata
+					Containers: []*pb.ServiceContainer{
+						{
+							Container:   containerJSON,
+							ServiceSpec: serviceSpecJSON,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	machineIP := "10.0.0.1"
+	mockCluster := &mockClusterClient{
+		machinesResp: &pb.ListMachinesResponse{
+			Machines: []*pb.MachineMember{
+				{
+					Machine: &pb.MachineInfo{
+						Name: "machine-1",
+						Network: &pb.NetworkConfig{
+							ManagementIp: pb.NewIP(netip.MustParseAddr(machineIP)),
+						},
+					},
+					State: pb.MachineMember_UP,
+				},
+			},
+		},
+	}
+
+	// Construct client with mocks
+	cli := &client.Client{
+		Docker: &docker.Client{
+			GRPCClient: mockDocker,
+		},
+		ClusterClient: mockCluster,
+	}
+
+	// Execute
+	containers, err := collectContainers(context.Background(), cli)
+	require.NoError(t, err)
+
+	// Verify
+	assert.Len(t, containers, 1)
+	if len(containers) > 0 {
+		c := containers[0]
+		assert.Equal(t, "container1", c.id)
+		assert.Equal(t, "test-container", c.name)
+		assert.Equal(t, "machine-1", c.machineName, "Should fall back to the single machine name when metadata is nil")
+	}
+}
+
+func TestCollectContainers_NilMetadata_MultipleMachines_Error(t *testing.T) {
+	// If we have multiple machines but receive nil metadata, it should return an error as it is ambiguous
+
+	// Setup container data
+	containerData1 := map[string]interface{}{
+		"Id": "container1",
+	}
+	containerJSON1, _ := json.Marshal(containerData1)
+
+	containerData2 := map[string]interface{}{
+		"Id": "container2",
+		"Config": map[string]interface{}{
+			"Image": "test-image",
+		},
+		"State": map[string]interface{}{
+			"Status":     "running",
+			"StartedAt":  "2023-01-01T12:00:00Z",
+			"FinishedAt": "0001-01-01T00:00:00Z",
+		},
+	}
+	containerJSON2, _ := json.Marshal(containerData2)
+
+	serviceSpecJSON, _ := json.Marshal(map[string]interface{}{})
+
+	// Setup mocks
+	mockDocker := &mockDockerClient{
+		listResp: &pb.ListServiceContainersResponse{
+			Messages: []*pb.MachineServiceContainers{
+				{
+					Metadata: nil, // Nil metadata
+					Containers: []*pb.ServiceContainer{
+						{
+							Container:   containerJSON1,
+							ServiceSpec: serviceSpecJSON,
+						},
+					},
+				},
+				{
+					Metadata: &pb.Metadata{Machine: "10.0.0.2"},
+					Containers: []*pb.ServiceContainer{
+						{
+							Container:   containerJSON2,
+							ServiceSpec: serviceSpecJSON,
+						},
+					},
+				},
+			},
+		},
+	}
+
+	mockCluster := &mockClusterClient{
+		machinesResp: &pb.ListMachinesResponse{
+			Machines: []*pb.MachineMember{
+				{
+					Machine: &pb.MachineInfo{
+						Name: "machine-1",
+						Network: &pb.NetworkConfig{
+							ManagementIp: pb.NewIP(netip.MustParseAddr("10.0.0.1")),
+						},
+					},
+				},
+				{
+					Machine: &pb.MachineInfo{
+						Name: "machine-2",
+						Network: &pb.NetworkConfig{
+							ManagementIp: pb.NewIP(netip.MustParseAddr("10.0.0.2")),
+						},
+					},
+				},
+			},
+		},
+	}
+
+	// Construct client with mocks
+	cli := &client.Client{
+		Docker: &docker.Client{
+			GRPCClient: mockDocker,
+		},
+		ClusterClient: mockCluster,
+	}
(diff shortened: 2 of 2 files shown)


cmd/uncloud/ps.go
```diff
@@ -0,0 +1,262 @@
...
+	groupByService = "service"
+	groupByMachine = "machine"
+)
+
+const (
+	statusHealthy   = "healthy"
+	statusUnhealthy = "unhealthy"
+	statusRunning   = "running"
+	statusOther     = "other"
+)
```
jabr, line 32: I think this should just be an "Go enum"?
```
type StatusClassification int
const (
    statusHealthy StatusClassification = iota
    statusUnhealthy
    statusRunning
    statusOther
)
```

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,262 @@
...
+This command provides a comprehensive overview of all running containers that are part of a service,
+making it easy to see the distribution and status of containers across the cluster.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if opts.groupBy != groupByService && opts.groupBy != groupByMachine {
+				return fmt.Errorf("invalid value for --group-by: %q, must be one of '%s' or '%s'", opts.groupBy, groupByService, groupByMachine)
+			}
+			return runPs(cmd, opts)
+		},
+	}
+	cmd.Flags().StringVarP(&opts.groupBy, "group-by", "g", groupByService, "Group containers by 'service' or 'machine'")
```
jabr: I did this as a `--group-by {property}` rather than a parameter-less `--group-by-machine` thinking there might more properties to group by that we'd later want (e.g. "health", label, etc). Not sure, though.

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,262 @@
...
+			var statusState string
+			healthStatus := ""
+			if ctr.Container.State.Health != nil {
+				healthStatus = ctr.Container.State.Health.Status
+			}
+
+			if healthStatus == container.Healthy {
+				statusState = statusHealthy
+			} else if healthStatus == container.Unhealthy {
+				statusState = statusUnhealthy
```
jabr: We might want to include other states beyond the explicit `Health.Status == "unhealthy"`, such as `OOMKilled` or some other unexpected exit status, restart loop, ...

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+	containers []containerInfo
+	err        error
+}
+
+type containersCollectedMsg struct {
+	containers []containerInfo
+	err        error
+}
+
+func newSpinnerModel(client *client.Client, message string) spinnerModel {
```
psviderski: Do we need a custom spinner model and can just use the spinner with one action? https://github.com/charmbracelet/huh?tab=readme-ov-file#bonus-spinner

The reason we have a custom `connectModel` is because we change the spinner text to the connection we're currently trying when iterating over all of them.

For this command, looks like we're only printing "Collecting container info..." until we get all the containers.

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+	s := spinner.New()
+	s.Spinner = spinner.Jump
+	s.Style = lipgloss.NewStyle().Foreground(lipgloss.Color("205"))
```
psviderski: For the connection spinner, I also tried to use the same spinner type and colour used by the compose progress library (`uc deploy` output) to have a consistent UI language. Would be nice to maintain the consistency here as well

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+
+func (m spinnerModel) collectContainers() tea.Msg {
+	services, err := m.client.ListServices(context.Background())
+	if err != nil {
+		return containersCollectedMsg{err: fmt.Errorf("list services: %w", err)}
+	}
+
+	var containers []containerInfo
+	for _, s := range services {
+		service, err := m.client.InspectService(context.Background(), s.ID)
```
psviderski: The client calls that are broadcasted to all machines are quite expensive, that's why it takes a while to list all containers at the moment.

You can try using only one call `client.Docker.ListServiceContainers`
https://github.com/psviderski/uncloud/blob/a30ed600b6fe7c74368f390e6f1d2402fb1df4be/internal/machine/docker/client.go#L479-L481

See the `ListServices` implementation for details: https://github.com/psviderski/uncloud/blob/458d282357da4e462c567beb82897f7168a7c037/pkg/client/service.go#L277

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		machines, err := m.client.ListMachines(context.Background(), nil)
+		if err != nil {
+			return containersCollectedMsg{err: fmt.Errorf("list machines: %w", err)}
+		}
+		machinesNamesByID := make(map[string]string)
+		for _, m := range machines {
+			machinesNamesByID[m.Machine.Id] = m.Machine.Name
+		}
```
psviderski, line 207: `machines` are always the same, they could be fetched and `machinesNamesByID` computed only once, not for every service to optimise the response time

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+		header = "SERVICE\tMACHINE\tCONTAINER ID\tNAME\tIMAGE\tSTATUS"
+	}
+	fmt.Fprintln(w, header)
+
+	for _, ctr := range containers {
+		id := ctr.id
+		if len(id) > 12 {
+			id = id[:12]
+		}
+		name := strings.TrimPrefix(ctr.name, "/")
```
psviderski: I don't think this is needed, we trim it when deserialising in our `api.Container`.

Not a big deal but also `stringid.TruncateID(c.ID)` from the docker package can be used to not hardcode the logic for shortening the container IDs

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,275 @@
...
+func printContainers(out io.Writer, containers []containerInfo, groupBy string) error {
+	w := tabwriter.NewWriter(out, 0, 0, 3, ' ', 0)
+	defer w.Flush()
+
+	var header string
+	if groupBy == groupByMachine {
+		header = "MACHINE\tSERVICE\tCONTAINER ID\tNAME\tIMAGE\tSTATUS"
+	} else {
+		header = "SERVICE\tMACHINE\tCONTAINER ID\tNAME\tIMAGE\tSTATUS"
+	}
```
psviderski: I quite like the layout of pod listing in k9s (similar to kubectl):
Image

It prints the node/machine column last. We do something similar for `uc inspect SERVICE`:

Image

I think we need to decide on the layout and stick to it for all relevant commands to decrease the cognitive load on users. I might be wrong but I also feel that changing the order when grouping by another column might not be desirable as it again requires additional effort to read headers and understand where the data is relocated now.

k9s, for example, support sorting by any column and doesn't change the layout. It's like tables on website. You can click on the header and change the sorting order, but don't swap columns. So I believe the `--group-by` option isn't much different to this. Potentially the flag could even be called `--sort`. Not sure which one is more intuitive though.

To me, it seems natural to present data ordered from the largest dimension to smallest.

service > container > container properties

If we have a namespace/stack concept, then

namespace > service > container > container properties

The machine a container runs on is somewhat an odd one. It's not one to many relationship like service-container, it's many to many. So it's hard to find a good place for it. Putting it as the first column somewhat breaks the logical hierarchy. Putting it as the last one still allows to find it quickly but it remains less intrusive. That's my current reasoning

psviderski (commented):
Overall looks good! Please see a few comments on the use of the custom bubble tea model, optimising the queries, and table layout.

cmd/uncloud/ps.go
jabr: Updated.

cmd/uncloud/ps.go
jabr: Fixed.

cmd/uncloud/ps.go
jabr: Done.

cmd/uncloud/ps.go
jabr: Done.

cmd/uncloud/ps.go
jabr: Fixed.

cmd/uncloud/ps.go
jabr: I've moved MACHINE to the last column.

One odd thing I noticed, it seems the lipgloss color styling of some cells in the second to last column and cause tabwriter to layout things incorrectly.

Have you encountered that issue?

psviderski: Ah yeah I think I did and that was the reason I used the lipgloss table instead of tabwriter for formatting the `uc image ls` output: https://github.com/psviderski/uncloud/blob/939645e1441a2e348445bb0e347b5b1ba377ebc1/cmd/uncloud/image/ls.go#L261-L304.

I think we need to use lipgloss table or something similar from the bubble tea world to correctly print styled data. Perhaps create our own convenient table abstraction wrapping one of those and use it for all CLI commands to make the formatting consistent.

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,241 @@
...
+	machinesNamesByIP := make(map[string]string)
+	md := metadata.New(nil)
+	for _, m := range machines {
+		if addr, err := m.Machine.Network.ManagementIp.ToAddr(); err == nil {
+			machineIP := addr.String()
+			machinesNamesByIP[machineIP] = m.Machine.Name
+			md.Append("machines", machineIP)
+		}
+	}
+	listCtx := metadata.NewOutgoingContext(ctx, md)
```
psviderski: JFYI you can also use this helper to create a context for broadcasting: https://github.com/psviderski/uncloud/blob/abea3e229e0989b33d3adac6e54b0ae37cddb4d8/pkg/api/client.go#L62-L98

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil, fmt.Errorf("list service containers: %w", err)
+	}
+
+	var containers []containerInfo
+	for _, msc := range machineContainers {
+		if msc.Metadata == nil {
+			continue
+		}
+		if msc.Metadata.Error != "" {
+			return nil, fmt.Errorf("list containers on machine %s: %s", msc.Metadata.Machine, msc.Metadata.Error)
```
psviderski: I think we shouldn't fail here as this will make `uc ps` command useless if there is a machine that is down or temporarily unavailable. What we can do instead is print a warning(s) about unavailable machine(s) but still list the containers on the available machines.

cmd/uncloud/ps.go
jabr: Thanks for pointing that out! I've updated this to use `ProxyMachinesContext`.

cmd/uncloud/ps.go
jabr: I've changed to this a `PrintWarning` (like the volume command uses) and continue.

cmd/uncloud/ps.go
jabr: > I think we need to use lipgloss table or something similar from the bubble tea world to correctly print styled data.

Yeah, that fixed it. I just re-used the setup from `image ls` for this PR, but...

> Perhaps create our own convenient table abstraction wrapping one of those and use it for all CLI commands to make the formatting consistent.

I'll do this as a follow-up PR with `lipgloss.table`. I was also thinking the abstraction might give us a good opportunity to add JSON output support, too, since that should be a simple second "formatter" for the table data these all generate.

jabr: btw, a draft of the "table abstraction" idea (though it ended up more of a JSON output option with some table helpers): https://github.com/psviderski/uncloud/pull/191

cmd/uncloud/ps.go
```diff
@@ -0,0 +1,255 @@
...
+		listCtx, "", container.ListOptions{All: true},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list service containers: %w", err)
+	}
+
+	var containers []containerInfo
+	for _, msc := range machineContainers {
+		if msc.Metadata == nil {
+			continue
```
psviderski: I agree this handling of broadcasted requests is very confusing and I have plans to rework it to simplify significantly. But for now, we need to still handle `Metadata` that is `nil` which could be a valid case when the cluster consists of only one machine. See https://github.com/psviderski/uncloud/blob/458d282357da4e462c567beb82897f7168a7c037/pkg/client/service.go#L123-L140

Please test that it works with one-machine cluster

jabr: I see. I've added similar logic from `service.go` to the usage in `ps.go` now.

I tested it against a single machine ucind cluster locally.

I also added some unit tests on the `ps.go#collectContainers` for these various edge cases. It relies heavily on mocking the data returned to it, though, and not sure how you feel about that testing approach. Let me know.

psviderski: Perfect, thank you so much for updating the Metadata handling! 🙏 

Regarding these mocked tests, I'm not a big fan of them as it's hard to replicate all the intricacies of the dependent components and keep them up to date with the changes in those components. For example, I'm going to change how the grpc router works which will slightly change the Metadata values. Then we will need to update all the mocks to reflect the change. Hopefully LLMs will help to do it correctly.

That's why I've been implementing mainly e2e tests so far to test the real workflows. Although, these are the slowest tests and they started to become a bit flaky recently. We will likely need to do something about it soon.

psviderski: And it's fine to keep the tests you added, appreciate your effort!

jabr: > Regarding these mocked tests, I'm not a big fan of them as it's hard to replicate all the intricacies of the dependent components and keep them up to date with the changes in those components.

Absolutely agree.

> That's why I've been implementing mainly e2e tests so far to test the real workflows. Although, these are the slowest tests and they started to become a bit flaky recently. We will likely need to do something about it soon.

Yeah, these really should be e2e tests, but I recalled you mentioning not really wanting to add more cluster setups in the e2e tests, too.

I looked at adding it to an existing e2e test, but they're oriented more around the functionality being tested rather than the cluster configuration needed, and it felt wrong to add these to e.g. the exec e2e tests...

Maybe we just need to change how the e2e tests are structured/grouped. Start with with "context" then the "functionality"? For example:
```
- in a one machine cluster
  - `exec`
    - does foo
  - `ps`
    - does bar
- in a two machine cluster
  - `dns`
    -  ... 
  - with a suspect node
    - `cmd`
      - ...
```

Something like that to let us reuse the heavy, time-consuming cluster setup across as many different tests as possible. Once they are up, the individual test cases can be very quick.

psviderski: >Yeah, these really should be e2e tests, but I recalled you mentioning not really wanting to add more cluster setups in the e2e tests, too.

That's a tradeoff between performance and good structure. The fewer clusters we create when running the entire test suit, the faster it will run. And we want it to be faster because this will let us run it more often when developing locally and catching bugs earlier. Also, I'm not too sure how close we are to reaching the memory limit on the free GitHub CI agents we're running it now as every cluster consumes a few hundreds MB of memory.
So I was postponing any substantial work on e2e tests until they break, and they still work, not perfectly but work 😄 

>Maybe we just need to change how the e2e tests are structured/grouped. Start with with "context" then the "functionality"?

Yeah, that's a good option! I don't remember if we have any test suits that use different clusters or they always 1-to-1. It would be great if we don't need to split tests for one logical feature into multiple different files/suits.

I remember that we have a few cases that can't be run in parallel (e.g. depending on setting env vars) so this is something to keep in mind when restructuring test suits.

I've just asked Claude if it's possible to have something like pytest fixtures (Python) in Go to inject a cluster as a dependency. It suggested an interesting approach with a lazy cluster pool. It will allow to structure tests based on the feature/concept they test but also share clusters of the required configuration between them. We can even split the current long `TestXXX` suits into more scoped ones for easier management.

## Plan: Shared Cluster Fixtures for E2E Tests

### Goal

 Reduce ucind cluster creation from 10 clusters to 2 shared clusters (one 3-machine, one 1-machine) while maintaining the ability to run individual tests in isolation when needed.

 Recommended Approach: TestMain with Lazy Cluster Pools

 Use Go's TestMain to manage shared cluster fixtures at the package level, with lazy initialization and fallback support for isolated test runs.

### Architecture
```
 test/e2e/
 ├── main_test.go      # NEW: TestMain + cluster pool management
 ├── cluster_test.go   # Modified: use shared cluster
 ├── service_test.go   # Modified: use shared cluster
 ├── ...
```
### Key Design Decisions

 1. Two shared cluster pools:
   - shared3Machine - for tests requiring 3 machines (8 tests currently)
   - shared1Machine - for tests requiring 1 machine (2 tests currently)
 2. Lazy initialization with sync.Once: Clusters are created on first request, not upfront. This means running a single test still works.
 3. Cleanup at package exit: TestMain handles cluster teardown after m.Run() completes.
 4. Fallback support: If TEST_CLUSTER_NAME env var is set, use that cluster (existing behavior preserved).

### Trade-offs

#### Pros:
 - Significant reduction in test setup time and resources
 - Tests still runnable in isolation (lazy init creates cluster on demand)
 - Backwards compatible with TEST_CLUSTER_NAME env var

#### Cons:
 - Tests must be careful about cleanup (already mostly true)
 - Test failures could potentially affect other tests sharing the cluster
 - Slightly more complex test infrastructure

Issue #163, opened 2025-11-04
GitHub API rate limit in install script on GitHub Actions


psviderski:
[Jiralite on Discord](https://discord.com/channels/1371726032104587335/1371726032838594565/1435051204127232223) got a 403 installing the CLI: https://github.com/thatskyapplication/thatskyapplication/actions/runs/19052532546/job/54415805274#step:5:1

403 was returned when trying to fetch the latest release version from github (https://api.github.com/repos/${GITHUB_REPO}/releases/latest): https://github.com/psviderski/uncloud/blob/5b66cec627c8fc939704efadeb63694c128ba0af/scripts/install_cli.sh#L23-L26

It likely hit the github api rate limit (60 unauthenticated reqs/hour per IP). I think we need to pass the auth header if `$GITHUB_TOKEN` is defined (which is the default on GitHub Actions, no configuration is required). 

And maybe add a few retries with sleep, just in case.

psviderski:
This didn't automatically fix the issue on github actions: https://github.com/thatskyapplication/thatskyapplication/actions/runs/19424931935/job/55570100301

as it looks like it has to be explicitly passed to the step or entire job with:
```
    env:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

I think the best solution would be to rewrite this logic that fetches the latest release version using api.github.com. Instead we can just extract the version from the https://github.com/psviderski/uncloud/releases/latest page.

PR #169, opened 2025-11-05
bug: Use GITHUB_TOKEN if present in scripts/install_cli.sh

jabr:
Fix for https://github.com/psviderski/uncloud/issues/163

Issue reproduced locally (403 error after 60 runs) and then working again when GITHUB_TOKEN was present in the environment.


psviderski:
Thank you! ❤️

Issue #125, opened 2025-09-15
Interactive uc command to select preferred cluster connection


psviderski:
https://github.com/psviderski/uncloud/issues/119 implements support for multiple `connections` in Uncloud config. `uc` tries each connection in the specified order until one succeeds.

```
current_context: default
contexts:
  default:
    connections:
      - ssh: ubuntu@1.2.3.4
      - ssh: root@5.6.7.8
      - ssh: user@9.10.11.12
```

To prioritise a particular connection, a user can edit the config and put the connection on the first place.

An interactive `uc` command for choosing the first/default connection would be handy. For example, we can extend the `uc ctx` command. Use up/down arrows to choose the context. Then, before pressing enter, you can press 'c', and a nested selector appears that allows you to select the default connection for that context (which is optional).

psviderski:
A new `uc ctx connection` command has been implemented in https://github.com/psviderski/uncloud/pull/170. 

<img width="587" height="165" alt="Image" src="https://github.com/user-attachments/assets/43fb98e7-aa66-4ee2-beef-dd004db90ebf" />

We can revisit it later if we need to extend the the `uc ctx` command with an additional action/view to select a connection.

PR #170, opened 2025-11-05
feat: Interactive uc command to select preferred cluster connection

jabr:
Implements https://github.com/psviderski/uncloud/issues/125 as `uc ctx connection`

### DIFF
--- a/cmd/uncloud/context/connection.go
+++ b/cmd/uncloud/context/connection.go
@@ -52,17 +52,8 @@ func selectConnection(uncli *cli.CLI) error {
 		return fmt.Errorf("select connection: %w", err)
 	}
 
-	// Reorder the connections with the selected one at the top.
-	selected := currentCtx.Connections[selectedConnection]
-	newConnections := make([]config.MachineConnection, 0, len(currentCtx.Connections))
-	newConnections = append(newConnections, selected)
-	for i, conn := range currentCtx.Connections {
-		if i == selectedConnection {
-			continue
-		}
-		newConnections = append(newConnections, conn)
-	}
-	currentCtx.Connections = newConnections
+	currentCtx.SetDefaultConnection(selectedConnection)
+	selected := currentCtx.Connections[0]
 
 	if err := uncli.Config.Save(); err != nil {
 		return fmt.Errorf("save config: %w", err)
--- a/internal/cli/config/context.go
+++ b/internal/cli/config/context.go
@@ -4,3 +4,21 @@ type Context struct {
 	Name        string              `yaml:"-"`
 	Connections []MachineConnection `yaml:"connections"`
 }
+
+func (c *Context) SetDefaultConnection(index int) {
+	// Do nothing if the index is out of range.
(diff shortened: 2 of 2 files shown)


cmd/uncloud/context/connection.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if len(currentCtx.Connections) == 0 {
+		return fmt.Errorf("no connections found in context '%s'", currentCtxName)
+	}
+
+	var selectedConnection *config.MachineConnection
+	form := huh.NewForm(
+		huh.NewGroup(
+			huh.NewSelect[*config.MachineConnection]().
```
psviderski: I think this could be significantly simplified by using an `int` type (index) instead of `MachineConnection` for the selected item. In this case, we don't need to compare connections and maintain the `Equal` method.

You may also want to move the reordering logic to a method of the `Context` struct, something like `currentCtx.SetDefaultConnection(index)`. It's up to you.

jabr: Ah, yeah, good idea! The non-comparable `MachineConnection` due to the key's byte slice sent both Gemini and I down a rabbit hole on that one... 😭

jabr:
> BTW, did you consider implementing this as part of uc ctx as described in https://github.com/psviderski/uncloud/issues/125?
Any reason you decided to make it a separate command?

I think I just misunderstood what you meant by "extend the uc ctx command" then. I figured you meant something like what I did here `uc ctx connection`.

So are you thinking `uc ctx` would have ui for selecting both the context and then the connection? Or a flag/option to select default connection rather than default context? Or do you mean extending/reusing the ui code from `ctx` in a separate command?

cmd/uncloud/context/connection.go
jabr: Updated.

jabr:
I somehow missed the last bit from your comment #125 about extending `uc ctx` originally:

> Then, before pressing enter, you can press 'c', and a nested selector appears that allows you to select the default connection for that context 

I just looked into doing that, and while I see we could have a second group/page with the connections selection after context, I don't see anything in `huh` to add a "custom command" like the "c" option to trigger going into that "nested" selector.

Did you have any idea in mind on how you would implement that UI?

Another option would be to make `uc ctx` _always_ a two step progress (first context, then connection). With the current default already selected, I don't think the extra enter would be a terrible ux for the default case.

psviderski:
>Did you have any idea in mind on how you would implement that UI?

Not really, I just fantasised the UX I'd likely want to have but unsure what's the best way to implement it. Looks like it will require wrapping the form in a Bubble Tea program and calling its View/Update methods instead of Run.

>Another option would be to make uc ctx always a two step progress (first context, then connection). 

It's not a terrible UX but I switch clusters often and in my 99.9% use cases I don't care about the connection so that will annoy me personally.

I like having a separate command for selecting only the connection that this PR introduces, so let's have it regardless of trying to make the `uc ctx` even fancier.

PR #175, opened 2025-11-09
feat: Allow container matching by ID prefix

tonyo:
Extends `Client.InspectContainer` to support prefix ID matching.

Should be at least applicable for:
* `uc exec --container XYZ ...`
* (future) `uc logs`


psviderski (approved):
👍

PR #180, opened 2025-11-17
fix: Extract version from latest redirect in install_cli.sh

### LINKED ISSUE #163: GitHub API rate limit in install script on GitHub Actions

Jiralite:
Should resolve #163 by using the latest redirect which does not require authentication.

### DIFF
--- a/scripts/install_cli.sh
+++ b/scripts/install_cli.sh
@@ -20,12 +20,8 @@ print_manual_install() {
 }
 
 fetch_latest_version() {
-    api_url="https://api.github.com/repos/${GITHUB_REPO}/releases/latest"
-    auth_header=""
-    if [ -n "$GITHUB_TOKEN" ]; then
-        auth_header="Authorization: Bearer $GITHUB_TOKEN"
-    fi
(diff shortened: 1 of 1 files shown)


psviderski (approved):
Looks great, thank you! ❤️

PR #182, opened 2025-11-19
fix: Store machine id on connection entries so it can be removed with machine

jabr:
Spun out of #162 to associate machine id with corresponding connection entry.

### DIFF
--- a/cmd/uncloud/machine/rm.go
+++ b/cmd/uncloud/machine/rm.go
@@ -154,11 +154,7 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
 
 	// Remove the connection to the machine from the uncloud config if it exists.
 	if uncli.Config != nil {
-		contextName := opts.context
-		if contextName == "" {
-			contextName = uncli.Config.CurrentContext
-		}
-
+		contextName := uncli.GetContextOverrideOrCurrent()
 		if context, ok := uncli.Config.Contexts[contextName]; ok {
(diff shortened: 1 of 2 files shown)


cmd/uncloud/machine/rm.go
```diff
@@ -155,8 +155,25 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
 		}
 	}
 
-	// TODO: remove the connection to the machine from the uncloud config if it exists. We need a way to associate
-	//  the machine with its connection in the config, e.g. by storing the machine name in the connection metadata.
+	// Remove the connection to the machine from the uncloud config if it exists.
+	if uncli.Config != nil {
+		contextName := opts.context
```
psviderski: I think this should be updated after https://github.com/psviderski/uncloud/pull/174 has been merged as the `--context` is now a global flag/option.

jabr: Updated. I added a helper function to the cli with that bit of "override or current" logic since it was used in two places now:
```
func (cli *CLI) GetContextOverrideOrCurrent() string {
  contextName := cli.contextOverride
  if contextName == "" {
    contextName = cli.Config.CurrentContext
  }
  return contextName
}
```

The cli's `ConnectClusterWithOptions` also has something similar, but it's mixed in with validation of the name and error logging. It might make sense to refactor that more later, but that would be more involved and seemed better to leave for a separate PR. These other uses have (simpler) error handling of their own already, too.

Issue #148, opened 2025-10-16
Install `uc` CLI on each cluster machine and allow running uc command agains local cluster


psviderski:
At the moment, you can run `uc` CLI only locally or on a control machine to control the remote cluster.

We can also install `uc` on every cluster machine and allow it to control the local cluster so a user can SSH into any machine and run `uc` commands there if they want to.

Technically, we can achieve this by automatically detecting if `/run/uncloud/uncloud.sock` is present on the machine when running `uc`. If the control socket is present and the uncloud config is not preset, try to connect to the cluster via the local unix socket. Essentially running `uc` as:
```
uc --connect unix:///run/uncloud/uncloud.sock ...
```
We also need to implement a unix socket [cluster connector](https://github.com/psviderski/uncloud/tree/2969b40ad42d0b2c6f58e269114853605fff123e/pkg/client/connector) for this to work. Should look similar to the [TCP one](https://github.com/psviderski/uncloud/blob/2969b40ad42d0b2c6f58e269114853605fff123e/pkg/client/connector/tcp.go).

psviderski:
The CLI cluster connector using a unix socket has been implemented in https://github.com/psviderski/uncloud/pull/186

miekg:
think the missing bit is just this? (no pr pushed - as I haven't tested it)
``` patch
diff --git a/cmd/uncloud/main.go b/cmd/uncloud/main.go
index 777adc3..95bc286 100644
--- a/cmd/uncloud/main.go
+++ b/cmd/uncloud/main.go
@@ -11,7 +11,7 @@ import (
 	cmdcontext "github.com/psviderski/uncloud/cmd/uncloud/context"
 	"github.com/psviderski/uncloud/cmd/uncloud/dns"
 	"github.com/psviderski/uncloud/cmd/uncloud/image"
-	"github.com/psviderski/uncloud/cmd/uncloud/machine"
+	cmdmachine "github.com/psviderski/uncloud/cmd/uncloud/machine"
 	"github.com/psviderski/uncloud/cmd/uncloud/service"
 	"github.com/psviderski/uncloud/cmd/uncloud/volume"
 	"github.com/psviderski/uncloud/cmd/uncloud/wg"
@@ -19,6 +19,7 @@ import (
 	"github.com/psviderski/uncloud/internal/cli/config"
 	"github.com/psviderski/uncloud/internal/fs"
 	"github.com/psviderski/uncloud/internal/log"
[...]

PR #186, opened 2025-11-21
feat: Add cli connector for unix domain socket

jabr:
Implements cli connection via local unix socket as described in #148:
> We also need to implement a unix socket cluster connector for this to work. Should look similar to the TCP one.

Supports connection configs:
```
current_context: local
contexts:
  local:
    connections:
      - unix: /run/uncloud/uncloud.sock
```

And running cli with connect option:
```
uc --connect unix:///run/uncloud/uncloud.sock ...
```

Tested on my cluster.

### DIFF
--- a/cmd/uncloud/main.go
+++ b/cmd/uncloud/main.go
@@ -81,7 +81,7 @@ func main() {
 
 	cmd.PersistentFlags().StringVar(&opts.connect, "connect", "",
 		"Connect to a remote cluster machine without using the Uncloud configuration file. [$UNCLOUD_CONNECT]\n"+
-			"Format: [ssh://]user@host[:port], ssh+cli://user@host[:port], tcp://host:port, or unix:///path/to/socket")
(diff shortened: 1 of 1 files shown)


cmd/uncloud/main.go
```diff
@@ -80,7 +81,7 @@ func main() {
 
 	cmd.PersistentFlags().StringVar(&opts.connect, "connect", "",
 		"Connect to a remote cluster machine without using the Uncloud configuration file. [$UNCLOUD_CONNECT]\n"+
-			"Format: [ssh://]user@host[:port], ssh+cli://user@host[:port], or tcp://host:port")
+			"Format: [ssh://]user@host[:port], ssh+cli://user@host[:port], tcp://host:port, or unix:///path/to/socket")
```
psviderski:
```suggestion
			"Format: [ssh://]user@host[:port], ssh+cli://user@host[:port], tcp://host:port, or unix:///path/to/uncloud.sock")
```

psviderski (approved):
Looks great! Thank you for this feature ❤️

PR #191, opened 2025-11-23
[spike] refactor/feat: Abstract cli table rendering and add JSON output format

jabr:
An AI generated draft of the table refactor with JSON output format idea. https://github.com/psviderski/uncloud/pull/165#discussion_r2553567779

Not sure how I feel about the overall implementation it took here, but it seemed like a useful draft to start a conversation on how we want to do it.

In brief, this approach is basically "JSON-first" now, generating structs that are more or less directly output for the JSON "format" and then the standard TUI table format is just defining column->field map with optional transforms, like "-" for nil, humanized times, etc. 

It also uses lipgloss table for everything now, rather than tabwriter. Not sure if there's any reason to not... do we want to keep tabwriter for some commands? Or as another output option (a simpler, non-styled version)?

### DIFF
--- a/cmd/uncloud/context/ls.go
+++ b/cmd/uncloud/context/ls.go
@@ -3,34 +3,48 @@ package context
 import (
 	"fmt"
 	"maps"
-	"os"
 	"slices"
-	"text/tabwriter"
 
 	"github.com/psviderski/uncloud/internal/cli"
+	"github.com/psviderski/uncloud/internal/cli/output"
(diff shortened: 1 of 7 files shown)


jabr:
I'm going to close this PR, as I'm not currently planning to pursue it further. It's a fairly involved changed with a lot of remaining work and not really a pressing issue for me.

I'll leave my branch, though, in case it's a useful reference for picking this back up again later.

PR #193, opened 2025-11-24
[spike] feat: Rough draft of a uncloud webui using Datastar

jabr:
Just a quick example Datastar-backed webui component as a potential starting point.

Screenshot 2025-11-23 at 10 19 47 PM


jabr:
I'm closing this branch as it was more just an experiment in how we might use Datastar, and I'm not currently planning to pursue it further. I'll leave the branch in case it is useful for reference later.

Issue #183, opened 2025-11-20
CLI commands to stop/start a service


psviderski:
As reported in https://github.com/psviderski/uncloud/issues/177, it might be useful to temporarily stop or restart a service without removing it.

For example, `uc service stop/start` similar to `docker stop/start`.

Regarding having `uc service restart` command for restarting a service, consider if we need to provide a way to do rolling restart.

jabr:
Can be closed with #195 merged.

psviderski:
Thanks!

PR #195, opened 2025-12-02
feat: Add `uc service start/stop` commands

jabr:
Basic implementation of service start/stop commands from #183.

~~The linter isn't happy with the duplication, but I'm not sure the best way to deal with that while also keeping with project conventions...~~ edit: ignoring linter complaints about duplication for these commands.

### DIFF
--- a/cmd/uncloud/service/stop.go
+++ b/cmd/uncloud/service/stop.go
@@ -12,9 +12,10 @@ import (
 )
 
 type stopOptions struct {
-	services []string
-	signal   string
-	timeout  int
+	services       []string
+	signal         string
+	timeoutChanged bool
+	timeout        int
 }
 
 func NewStopCommand() *cobra.Command {
@@ -27,6 +28,7 @@ func NewStopCommand() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			uncli := cmd.Context().Value("cli").(*cli.CLI)
 			opts.services = args
+			opts.timeoutChanged = cmd.Flags().Changed("timeout")
 			return stop(cmd.Context(), uncli, opts)
 		},
 	}
@@ -45,7 +47,7 @@ func stop(ctx context.Context, uncli *cli.CLI, opts stopOptions) error {
 	stopOpts := container.StopOptions{
 		Signal: opts.signal,
 	}
-	if opts.timeout != 0 {
+	if opts.timeoutChanged {
 		stopOpts.Timeout = &opts.timeout
 	}
 


cmd/uncloud/service/start.go
```diff
@@ -0,0 +1,49 @@
+package service
```
psviderski, line 2: Regarding the linter, let's just ignore it for these commands. I would rather have this very straightforward duplicate code than try to make the 10 common lines reusable for no real benefits. Just add `//nolint:dupl` at the beginning of the file.

```suggestion
//nolint:dupl
package service
```

cmd/uncloud/service/stop.go
```diff
@@ -0,0 +1,49 @@
+package service
```
psviderski, line 2:
```suggestion
//nolint:dupl
package service
```

jabr:
Good call on the stop signal/timeout options. Added.

cmd/uncloud/service/stop.go
```diff
@@ -0,0 +1,62 @@
...
+		return fmt.Errorf("connect to cluster: %w", err)
+	}
+	defer client.Close()
+
+	stopOpts := container.StopOptions{
+		Signal: opts.signal,
+	}
+	if opts.timeout != 0 {
+		stopOpts.Timeout = &opts.timeout
+	}
```
psviderski, line 52: It seems this behaviour is different to the default docker's behaviour of `--timeout`. Please see https://github.com/docker/cli/blob/master/cli/command/container/stop.go#L60-L64 how it handles when `--timeout` is not set or not (`.Changed("timeout")`).

Note that `0` is valid value that is different to `nil`: https://github.com/moby/moby/blob/8068dfb686a09fa66a577a4b94646d15349802dc/client/container_stop.go#L16-L24

jabr: Updated to use the same flag `Changed` check to set the timeout option properly.

cmd/uncloud/service/stop.go
```diff
@@ -0,0 +1,64 @@
...
+		Args:  cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			uncli := cmd.Context().Value("cli").(*cli.CLI)
+			opts.services = args
+			opts.timeoutChanged = cmd.Flags().Changed("timeout")
+			return stop(cmd.Context(), uncli, opts)
+		},
+	}
+	cmd.Flags().StringVarP(&opts.signal, "signal", "s", "", "Signal to send to the container")
+	cmd.Flags().IntVarP(&opts.timeout, "timeout", "t", 0, "Seconds to wait before killing the container")
```
jabr, line 36: Note: I kept the "default" `0` here as that's what the docker code does: https://github.com/docker/cli/blob/master/cli/command/container/stop.go#L48

cmd/uncloud/service/stop.go
psviderski: Thank you!

PR #204, opened 2025-12-05
feat(uc/caddy): expose 443/udp port for HTTP3

f7zorn:
Next to TCP 443 also expose UDP 443 port so that caddy can accept HTTP3 connections. HTTP3 is enabled by default in Caddy.


z0rrn:
I do not know if this is the correct approach here, I just what I found. If this is completely wrong I can also just write an issue.

f7zorn:
I now tested this and this change works perfectly and allows HTTP3 requests.

psviderski:
Yeah, this sounds good to me 👍

PR #208, opened 2025-12-07
namespaces

nick-potts:
# Namespace isolation and DNS scoping

Introduces namespaces as a first-class networking boundary:
- Services carry a validated namespace (default `default`).
- `uncloud-system` is reserved for infra and exempt from isolation to keep ingress working across namespaces.
- Internal DNS responses are scoped to the caller’s namespace; host nslookups default to `default`.
- Containers are labeled with their namespace; ipsets/iptables enforce drop rules across namespaces on the Uncloud bridge.
- Published ports/HTTPS ingress remain reachable across namespaces.

CLI/Compose: `x-namespace` on services plus deploy-time override:

```yaml
services:
  api:
    image: ghcr.io/acme/api:latest
    x-namespace: prod
```

```bash
uc deploy --namespace staging   # overrides all services
uc run --namespace prod ghcr.io/acme/api:latest
uc service ls --namespace prod  # filter + show namespace column
```

Namespace validation follows DNS labels (lowercase alnum + hyphens, 1–63 chars) and fails fast in CLI/API. Service subcommands propagate namespace context for exec/inspect/logs/rm/run/scale/list.

Networking/resolver:
- Resolver derives namespace from the source container IP and only returns records from that namespace (service name, service ID, machine-scoped).
- Host-level queries use `default`.
- Firewall/ipset builds per-namespace sets and drops lateral cross-namespace bridge traffic; ingress/published ports are unaffected.

Migration:
- One-time migration redeploys Caddy into `uncloud-system` on startup to keep ingress reachable after isolation is enabled.
- Cluster controller waits for initial sync, triggers migration, and logs outcome without crashing on failure.

Docs/tests:
- Namespace concept and compose reference pages; CLI docs note namespace flags; internal DNS doc mentions namespace scoping/host default.
- Tests cover namespace validation, compose overrides, resolver scoping, deploy plumbing, service listing, connector tweaks, and deploy container handling.

Operational notes:
- Workloads without an explicit namespace stay in `default`.
- `uncloud-system` is reserved and bypasses isolation.
- Isolation applies to internal bridge traffic, not published ports/ingress.
- Host nslookup uses `default`, so resolve other namespaces from a container in that namespace.


miekg:
would love to see this, haven't seen much docs on multi tentant mode - which would be really nice, is there an open issue/discussion for this feature?

psviderski:
This is an old discussion on the namespace/env/stack/environment concept: https://github.com/psviderski/uncloud/discussions/94

PR #212, opened 2025-12-08
resource reservations

nick-potts:
Implemented adaptive scheduling and surfaced reservations in the CLI:
  - Added a no-reservation ranker to keep round-robin spreading when CPU/memory reservations are zero, while retaining the existing spread ranker (existing+scheduled) when reservations are set. Tie-breaking stays deterministic.
  - Updated scheduler/strategy tests to cover adaptive ranking, reservation rollback, port conflicts, force-recreate, volume constraints, and resource accounting.
  - Added scheduling docs for clarity, documenting eligibility, adaptive ranking, resource accounting, modes, conflicts, and troubleshooting; updated the global vs replicated table to match current behavior.
  - Exposed --reserve-cpu and --reserve-memory on uc service run to set CPUReservation/MemoryReservation, aligning ad-hoc runs with deploy/Compose scheduling.

Design choices:
  - Opted for an implicit adaptive ranker instead of a new x-spread knob; balanced behavior remains for reserved workloads, round-robin for best-effort workloads.
  - CLI change is additive and backward compatible; limits remain as before, reservations now available for scheduling-aware runs.


nick-potts:
On reflection, I think perhaps this would would be better off waiting for other scheduler improvements first so I'm going to mark it as a draft.

I think if/when `depends_on`, `start_stop` and perhaps `rollbacks` are in place it'd make more sense.

PR #214, opened 2025-12-09
Simplify Makefile with configurable server list

nick-potts:
Replace hardcoded server IPs with a SERVERS variable that can be overridden via .env file or command line. Switch from scp to rsync for faster transfers. All dev targets now loop over the same server list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,13 @@ demo-reset:
 		ssh $$server "AUTO_CONFIRM=true sudo -E uncloud-uninstall"; \
 	done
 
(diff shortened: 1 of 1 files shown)


psviderski:
We deprecated the `Makefile` and started migrating targets to [`mise tasks`](https://mise.jdx.dev/tasks/).
Now we have `dev:install` and `dev:reset` tasks that implement the proposed changes. See docs: https://github.com/psviderski/uncloud/blob/main/HACKING.md#manual-testing-on-dev-machines

PR #216, opened 2025-12-09
Add StopGracePeriod

nick-potts:
Wire up `stop_grace_period` from Docker Compose to rolling deployments.

### DIFF
--- a/website/docs/8-compose-file-reference/1-support-matrix.md
+++ b/website/docs/8-compose-file-reference/1-support-matrix.md
@@ -32,6 +32,7 @@ The following table shows the support status for main Compose features:
 | `pull_policy`      | ✅ Supported        | `always`, `missing`, `never`                                                          |
 | `secrets`          | ❌ Not supported    | Use configs or environment variables                                                  |
 | `security_opt`     | ❌ Not supported    |                                                                                       |
+| `stop_grace_period`| ✅ Supported        | Time to wait after SIGTERM before SIGKILL                                             |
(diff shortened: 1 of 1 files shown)


nick-potts:
Current behavior:
  1. RemoveContainerOperation calls StopContainer() → Docker sends SIGTERM
  2. During `stop_grace_period`, container state is still `Running=true`
  3. Healthy() returns true → Caddy keeps routing traffic
  4. Container may be refusing connections or returning errors
  5. Only after container exits does Caddy config get regenerated

Perhaps something like this would be a good natural fit:

```golang
  const (
      SyncStatusSynced   = "synced"
      SyncStatusOutdated = "outdated"
      SyncStatusDraining = "draining"  // NEW
  )
```

The drain flow becomes:
  1. Set `sync_status = 'draining'` → container excluded from Caddy
  2. Wait for drain period (allow in-flight requests to complete)
  3. Send SIGTERM, respect stop_grace_period
  4. Delete from store

  Proposed x-drain extension:
```yaml
services:
  web:
    image: nginx
    stop_grace_period: 30s
    x-drain-period: 30s # time to wait after removing from load balancer before SIGTERM
```

This would give users control over the connection drain window separately from the process shutdown window.

psviderski:
Resolved conflicts and changed the `StopGracePeriod` from `*int` to `*time.Duration` in https://github.com/psviderski/uncloud/commit/d9934b9dbadc7a6ff795473f86e840fef6975a45
Thanks!

PR #219, opened 2025-12-14
fix: Set volume driver name to "local" if not specified

jabr:
`MatchesDockerVolume` returns a error when volume driver name is blank versus "local".

Not certain this is the right fix, but I'm updating the volumes `SetDefaults` to set the driver name to "local" if not set.

volume-mismatch-error

https://discord.com/channels/1371726032104587335/1371726032838594565/1449615928185720895


psviderski (approved):
Good catch, thank you! ❤️

PR #224, opened 2025-12-19
Add Nix flakes to have reproducible build and shell

divaltor:


psviderski:
I appreciate your time for putting this nix shell together.
We're already using [mise](https://mise.jdx.dev/) ([`.mise.toml`](https://github.com/psviderski/uncloud/blob/main/.mise.toml)) for managing local dev and build environments. It works pretty well and is super easy to use.

We haven't used its [tasks](https://mise.jdx.dev/tasks/) feature yet but I'd like to migrate the current Makefile to mise tasks.

I also had pretty good experience with a nix devshell but only when I had a team member who was looking after it. Idk, I don't feel like I want to deal with it again and learn what all this black magic means. Maybe I'm just too lazy and stupid for nix but it's over my head for now. Perhaps Claude can help with changes but I generally prefer to understand how the things I use work. However, I feel more comfortable continuing with mise unless nix solves problems that mise can't and that we actually have and need to solve them.

PR #226, opened 2025-12-20
feat: Machine "ping" (RTT from corrosion) times

jabr:
Just experimenting with an idea, but I was thinking the round-trip time (RTT) data from corrosion might be useful for some future Uncloud functionality (e.g. an improved "nearest" local DNS that can meaningfully order past machine-local, possibly some integration with Caddy upstream routing, replica placement preference ["spread out", "keep close"], etc).

I also added a `uc machine rtt` command to show the ~avg~median and stddev of RTT times for each machine/peer pair, which might be a useful diagnostic tool all on its own. Output is something like this:
```
MACHINE    PEER       MEDIAN   STDDEV
amalthea   himalia    140ms    ±19.4ms
amalthea   thebe      39ms     ±1.1ms
himalia    amalthea   168ms    ±18.5ms
himalia    thebe      203ms    ±42.3ms
thebe      amalthea   40ms     ±2.0ms
thebe      himalia    158ms    ±15.2ms```

Also added `RTT` column to `wg show` output.

### CODE REVIEW

**Comment by jabr** [comment]
Pulling these comments from discord for discussion here:

> On the implementation side, FYI I've just introduced a new InspectMachine gRPC method that in addition to the MachineInfo returns the Corrosion db version (lamport logical time). I think we can extend this API with all sorts of useful information about the machine if it's something more or less always available. Corrosion is a critical component so this method already depends on it.

The timing is always available from corrosion, but it's another admin API call similar to that for the membership states (but with `{"Cluster":"Members"}`). I don't think it would add much overhead, but the reason I did not add it to `InspectMachine` was because that function is heavily used other commands and would incur some cost across commands that don't need it.

> Although, I'm not sure if RTTs should be part of this API. At least while still experimenting, it's probably easier to keep separate. (just thinking outloud) 

Yeah, that's pretty much where I landed with this and added a new gRPC api for the `uc machine rtt` command. Once we have a better idea of how we might want to use these numbers in the CLI, it make might sense to revisit then. For example, another option might be an "extended info" flag to `InspectMachine` that includes additional data like these RTT statistics?

**Comment by jabr** [comment]
> did not add it to InspectMachine was because that function is heavily used other commands

Looking at the MR again, I realized I had mixed up `ListMachines` and `InspectMachine` and thought you were suggesting adding it to `ListMachines`...

Looking closer at `InspectMachine`, that makes a lot more sense. I'll add it to that.

**Comment by jabr** [comment]
For some reason, github is not letting me edit the title of the PR, but I think it is ready for review now, not a "draft".

edit: seems to be a bug in the github web ui. the mobile app and gh cli still let you update titles. 🤷

File: internal/corrosion/admin.go [hunk 1]
```diff
@@ -256,3 +257,97 @@ func (c *AdminClient) ClusterMembershipStates(latest bool) ([]ClusterMembershipS
 	}
 	return states, parseErr
 }
+
+type MemberRTTStats struct {
+	Addr    netip.AddrPort
+	Average float64
```
psviderski: What are the units for rtts? Are they milliseconds?

cmd/uncloud/machine/rtt.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		return rows[i].machine < rows[j].machine
+	})
+
+	// Print table
+	tw := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+	fmt.Fprintln(tw, "MACHINE\tPEER\tRTT")
+
+	for _, r := range rows {
+		fmt.Fprintf(tw, "%s\t%s\t%.1f ±%.1f\n", r.machine, r.peer, r.avg, r.stdDev)
```
psviderski: What do you think about adding units to the numbers so it's clear what they mean?

internal/corrosion/admin.go
```diff
@@ -256,3 +257,97 @@ func (c *AdminClient) ClusterMembershipStates(latest bool) ([]ClusterMembershipS
...
+
+		if len(rtts) == 0 {
+			continue
+		}
+
+		var sum float64
+		for _, rtt := range rtts {
+			sum += rtt
+		}
+		avg := sum / float64(len(rtts))
```
psviderski: What are your considerations regarding using avg vs median?
I guess there could be a network hiccup and one data point could be significantly larger than others which may skew the average. Is this what we typically want or the median would better suit us e.g. for sorting upstreams based on the latency (rtt)?

cmd/uncloud/machine/rtt.go
```diff
@@ -0,0 +1,104 @@
...
+	// Sort by machine name then peer name
+	sort.Slice(rows, func(i, j int) bool {
+		if rows[i].machine == rows[j].machine {
+			return rows[i].peer < rows[j].peer
+		}
+		return rows[i].machine < rows[j].machine
+	})
+
+	// Print table
+	tw := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
```
psviderski: Please use https://github.com/psviderski/uncloud/blob/main/internal/cli/tui/table.go#L9-L24. See other commands have been consistently migrated to it.

psviderski (approved):
This looks really good! Sorry it took that long for me to take a look at this PR.
I actually need this for my multi-region setup to sort upstreams in Caddy and route traffic to the local region first on each Caddy instance.

Please see a few questions and minor comments below.

internal/corrosion/admin.go
jabr: Yeah, milliseconds. https://github.com/superfly/corrosion/blob/1b0fcf1b12a81c5394dd4b9e5a7815b0ab95d575/crates/corro-types/src/members.rs#L176

internal/corrosion/admin.go
jabr: Makes sense. That’s basically the same reason I included the standard deviation, but median would be a better single number metric to have for most applications using this (eg routing, “nearest” dns, etc).

Should I keep mean/stddev, too? Or just replace both of those with median?

cmd/uncloud/machine/rtt.go
jabr: Yeah, I saw those updates last night when rebasing the proxy branch. I should have time tonight to update this and make the millisecond label and median vs avg/stddev changes.

internal/corrosion/admin.go
psviderski: Up to you. I think the standard deviation is still useful for detecting unstable links (high jitter). Not sure if we end up using it for nearest detection but might be useful for network troubleshooting. But IMO the mean itself doesn't add value once you have the median.

psviderski:
Just another thought after sleeping on it. Did you consider displaying RTTs in the output of any existing commands?
I'm thinking that even though we take this data from the Corrosion cluster members (machines), it's about the network layer. And the network layer in uncloud is WireGuard.

We have the `uc wg show` command that lists all peers of the machine (currently connected to or `--machine`). It feels very natural to add the RTT column along HANDSHAKE, RECEIVED, SENT there to see all the relevant data in one place.

The only difference with the `machine rtt` command is that it prints all the combinations of machine <-> peer.

I'm still happy to merge this one as is, just food for thought to potentially improve the UX.

internal/corrosion/admin.go
jabr: I went with median and stddev. New output table looks like:
```
MACHINE    PEER       MEDIAN   STDDEV
amalthea   himalia    140ms    ±19.4ms
amalthea   thebe      39ms     ±1.1ms
himalia    amalthea   168ms    ±18.5ms
himalia    thebe      203ms    ±42.3ms
thebe      amalthea   40ms     ±2.0ms
thebe      himalia    158ms    ±15.2ms
```

jabr:
> We have the uc wg show command ... It feels very natural to add the RTT column ...

Good idea! I added a RTT column to the `wg show` output, too. Very natural fit for that command.

```
PEER       PUBLIC KEY   ENDPOINT          HANDSHAKE   RTT     RECEIVED   SENT      ALLOWED IPS
amalthea   9dK...Oio=   65.x.x.x:51820    22s ago     121ms   735.3MB    764.7MB   fdcc:xxx/128, 10.210.1.0/24
thebe      1Wg...yFI=   93.x.x.x:51820   1m14s ago   138ms   839.1MB    774.4MB   fdcc:xxxx/128, 10.210.0.0/24
```

I also thought about other commands that might make sense, but none felt like a great fit, imo. `machine ls` is the closest, but it doesn't have the "from a particular machine" perspective needed for the directional RTT to make sense.

`wg show` has the "from connection machine" perspective, and `machine rtt` shows all pair-wise combos of machines, so it's an "omni" perspective. `machine ls` is technically a view from the connection machine, as well, but that's more of an implementation detail than something obvious/emphasized to the user like `wg show` is.

jabr:
Everything from the review should be updated now and ready for another look. Added `ms` unit to output (and improved table column labels), using median instead of average, moved to `tui.Table`, and added RTT to `wg show`.

Let me know if you see anything else.

PR #233, opened 2025-12-28
fix(configs): Create non-existent parent directories automatically

### LINKED ISSUE #232: Can't mount Compose config at a path that doesn't exist in container
The deployment of the following compose file:
```yaml
services:
  grafana:
    image: grafana/grafana:12.2.0
    configs:
      - source: grafana-dashboard-victoriametrics
        target: /var/lib/grafana/dashboards/victoriametrics.json
    volumes:
      - grafanadata:/var/lib/grafana

configs:
  grafana-dashboard-victoriametrics:
    file: ./grafana/dashboards/victoriametrics.json

volumes:
  grafanadata:
```
fails with the following error:
```
Error: deploy services: create container: rpc error: code = Internal desc = inject configs: copy config file 'grafana-dashboard-victoriametrics' to container: copy to container: Error response from daemon: Could not find the file /var/lib/grafana/dashboards in container a9a6b7eeda4dbd8a6ea1f43bb66b2cba085d7c0fa2c9a71fb21f4d86a3a917b7
```

`/var/lib/grafana/dashboards/` directory doesn't exist in the container when we mount a new empty volume at `/var/lib/grafana`. However, `docker compose up` successfully runs the following compose file and mounts the config as expected.

We need to compare the implementation details of copying configs to created containers in uncloud and docker compose and make it work in uncloud in a similar way.
I don't know the details off the top of my head, but I suspect that we might want the `target` for the docker `CopyToContainer` call to always be the root `"/"`. And then create the full file paths that start from the container root in the tar archive. This way unarchiving such a tar will result in creating all intermediate directories that weren't originally present.

cc @tonyo

tonyo:
Fixes #232 and another issue that didn't allow to create two config mounts with the same source config.

It should more closely mirror now the Compose behavior which also uses the full target path in the tar header and extracts it to the root:

* https://github.com/docker/compose/blob/59f04b85af552dab7561255dcc08e9328c5bfe67/pkg/compose/secrets.go#L131
* https://github.com/docker/compose/blob/59f04b85af552dab7561255dcc08e9328c5bfe67/pkg/compose/secrets.go#L162

### DIFF
--- a/pkg/client/compose/config.go
+++ b/pkg/client/compose/config.go
@@ -13,18 +13,27 @@ import (
 func configSpecsFromCompose(
 	configs types.Configs, serviceConfigs []types.ServiceConfigObjConfig, workingDir string,
 ) ([]api.ConfigSpec, []api.ConfigMount, error) {
-	var configSpecs []api.ConfigSpec
 	var configMounts []api.ConfigMount
 
+	// Temporary map to hold config specs
+	configSpecsMap := make(map[string]api.ConfigSpec)
+
+	// We iterate over all service config objects (config mounts)
 	for _, serviceConfig := range serviceConfigs {
 		var spec api.ConfigSpec
 
-		if projectConfig, exists := configs[serviceConfig.Source]; exists {
-			if projectConfig.External {
-				return nil, nil, fmt.Errorf("external configs are not supported: %s",
-					serviceConfig.Source)
-			}
+		projectConfig, exists := configs[serviceConfig.Source]
+		if !exists {
+			return nil, nil, fmt.Errorf("config '%s' not found in project configs", serviceConfig.Source)
+		}
 
+		if projectConfig.External {
+			return nil, nil, fmt.Errorf("external configs are not supported: %s",
+				serviceConfig.Source)
+		}
+
+		spec, exists = configSpecsMap[serviceConfig.Source]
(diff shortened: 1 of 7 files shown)


pkg/client/compose/config.go
```diff
@@ -71,5 +78,11 @@ func configSpecsFromCompose(
 		configMounts = append(configMounts, mount)
```
psviderski, line 78: Just found out that `ContainerSpec.Equal` depends on the order of config mounts. This is not a big deal as usually people update and deploy the same Compose file so the order of `service.configs` normally stays the same hence the spec comparison works as expected.

However, while we're on this, it would be good to make the comparison order agnostic, like with volume mounts to not trigger a spec change/redeployment if only the order is different: https://github.com/psviderski/uncloud/blob/79dc05cb665d6541ea2ff1ca16af4119ae6b47d4/pkg/api/service.go#L289-L300

psviderski (approved):
Excellent! Just one note about the `Equal` implementation for the config mounts. We can do it in a separate PR. It's up to you to decide, just raising to not forget.

pkg/client/compose/config.go
tonyo: Makes sense, but will handle it in another PR

PR #235, opened 2025-12-29
fix: Equals and Clone for configs

tonyo:
Follow-up to #233.
Also fixes deep copy for `ContainerSpec.Env`.


psviderski (approved):
LGTM! ❤️

PR #236, opened 2025-12-30
build: Generate CLI docs in GHA

tonyo:
CI now should remind us if the CLI docs are out of date.

### DIFF
--- a/cmd/uncloud/service/exec.go
+++ b/cmd/uncloud/service/exec.go
@@ -59,7 +59,7 @@ If the service has multiple replicas and no container ID is specified, the comma
 
 	execCmd.Flags().BoolVarP(&opts.detach, "detach", "d", false, "Detached mode: run command in the background")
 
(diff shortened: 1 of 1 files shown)


cmd/uncloud/service/exec.go
```diff
@@ -59,7 +59,7 @@ If the service has multiple replicas and no container ID is specified, the comma
 
 	execCmd.Flags().BoolVarP(&opts.detach, "detach", "d", false, "Detached mode: run command in the background")
 
-	execCmd.Flags().BoolVarP(&opts.noTty, "no-tty", "T", !cli.IsStdoutTerminal(),
+	execCmd.Flags().BoolVarP(&opts.noTty, "no-tty", "T", false,
```
tonyo, line 62: Moved this logic below; otherwise the docs would look differently when generated with and without the attached stdout.

PR #239, opened 2026-01-06
Add support for `sysctls` compose key

4ndv:
Adds support for `sysctls` key, which is useful for VPN and other network-related containers (I have a separate VPS for VPN which I want to include in my uncloud cluster)

Tested same as #238: deploy an app without sysctls, redeploy the same app with sysctls, watch containers to be recreated with all the keys included:

```
lynx@uncloud:~$ sudo docker inspect dde4712ea39c | grep Sysctls -A 3
            "Sysctls": {
                "net.ipv4.conf.all.src_valid_mark": "1",
                "net.ipv4.ip_forward": "1"
            },
```

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -362,7 +362,7 @@ func (s *ContainerSpec) Clone() ContainerSpec {
 		for k, v := range s.Sysctls {
 			spec.Sysctls[k] = v
 		}
-  }
+	}
 	return spec
 }
 


pkg/api/service.go
```diff
@@ -347,15 +357,12 @@ func (s *ContainerSpec) Clone() ContainerSpec {
...
-		spec.CapDrop = make([]string, len(s.CapDrop))
-		copy(spec.CapDrop, s.CapDrop)
-	}
-
+	if s.Sysctls != nil {
+		spec.Sysctls = make(map[string]string, len(s.Sysctls))
+		for k, v := range s.Sysctls {
+			spec.Sysctls[k] = v
+		}
+  }
```
psviderski:
```suggestion
	}
```

4ndv:
Thank you!

PR #243, opened 2026-01-13
feat: auto-provision volumes on all machines for global services

### LINKED ISSUE #234: Auto-provision volumes on all machines for global services
When deploying a `mode: global` service that uses local volumes, uncloud currently creates the volume on only one machine and deploys only one container there. This replicates the same behaviour as for the default `replicated` mode to not give an impression that a local Docker volume could magically be shared by multiple replicas running on different machines.

However, this also contradicts the expected semantics of global services which should run on every machine.

The current workaround is to manually create a volume with the same name on each machine before deploying:
```
uc volume create myvolume -m machine1
uc volume create myvolume -m machine2
uc volume create myvolume -m machine3
uc deploy
```

I think we should change the default behaviour for deploying `global` services using volumes and automatically provision the declared volumes on all targeted machines. Each machine gets its own independent volume with the same name.

This change will only affects global services. Replicated services retain current behaviour where volumes are created on one machine and replicas are scheduled there (users can still manually create volumes on additional machines to spread replicas). We should properly document this difference.

zasdaym:
Create volumes on all eligible machines for global service. Fixes #234

### DIFF
--- a/pkg/client/deploy/scheduler/volume.go
+++ b/pkg/client/deploy/scheduler/volume.go
@@ -167,9 +167,15 @@ func (s *VolumeScheduler) Schedule() (map[string][]api.VolumeSpec, error) {
 			placedVolumes[volumeName] = struct{}{}
 		}
 	}
-	// Skip constraint propagation for volumes only used by global services.
+	// Skip constraint propagation for volumes used by global services.
+	// Also check for invalid configuration: volume shared between global and replicated services.
 	for volumeName := range s.volumeSpecs {
-		if s.isVolumeOnlyForGlobalServices(volumeName) {
+		if s.isVolumeSharedBetweenGlobalAndReplicated(volumeName) {
+			return nil, fmt.Errorf("volume '%s' cannot be shared between global and replicated services: "+
+				"global services require the volume on all machines while replicated services require "+
+				"co-location with the volume", volumeName)
+		}
+		if s.isVolumeForGlobalService(volumeName) {
 			placedVolumes[volumeName] = struct{}{}
 		}
 	}
@@ -184,24 +190,16 @@ func (s *VolumeScheduler) Schedule() (map[string][]api.VolumeSpec, error) {
 	scheduledVolumes := make(map[string][]api.VolumeSpec)
 	for volumeName, volumeSpec := range s.volumeSpecs {
 		existingMachines := s.existingVolumeMachines[volumeName]
-
-		// Check for invalid configuration: volume shared between global and replicated services.
-		if s.isVolumeSharedBetweenGlobalAndReplicated(volumeName) {
-			return nil, fmt.Errorf("volume '%s' cannot be shared between global and replicated services: "+
-				"global services require the volume on all machines while replicated services require "+
-				"co-location with the volume", volumeName)
-		}
-
 		serviceNames := s.volumeServices[volumeName]
 		if len(serviceNames) == 0 {
 			return nil, fmt.Errorf("bug detected: no services using volume '%s'", volumeName)
 		}
 
 		// Get the eligible machines for this volume.
-		// For volumes only used by global services: compute UNION of all services' eligible machines.
+		// For volumes used by global services: compute UNION of all services' eligible machines.
 		// For other volumes: any service will have the same set after constraint convergence.
 		var eligibleMachines mapset.Set[string]
-		if s.isVolumeOnlyForGlobalServices(volumeName) {
+		if s.isVolumeForGlobalService(volumeName) {
 			// Compute union of eligible machines for all global services using this volume.
 			eligibleMachines = mapset.NewSet[string]()
 			for _, serviceName := range serviceNames {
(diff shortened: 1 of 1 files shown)


pkg/client/deploy/scheduler/volume.go
```diff
@@ -177,22 +186,32 @@ func (s *VolumeScheduler) Schedule() (map[string][]api.VolumeSpec, error) {
...
-			serviceEligibleMachines[serviceName] = mapset.NewSet(machineID)
-		}
-		placedVolumes[missingVolumeName] = struct{}{}
-		scheduledVolumes[machineID] = append(scheduledVolumes[machineID], missingVolumeSpec)
 
-		// Propagate the updated constraints.
-		if err := s.propagateConstraintsUntilConvergence(serviceEligibleMachines, placedVolumes); err != nil {
-			return nil, fmt.Errorf("unexpected error while propagating constraints after "+
-				"scheduling volume '%s' on machine '%s': %w", missingVolumeName, machineID, err)
+		if s.isVolumeForGlobalService(missingVolumeName) {
```
psviderski: Am I right that if multiple global services use the same volume then it will be created on the intersection of their eligible machines? This is the set of machines we get after the first `propagateConstraintsUntilConvergence` before we start scheduling volumes.

It's been a while since I wrote this non-trivial logic 😄

test/e2e/service_test.go
```diff
@@ -1052,37 +1052,6 @@ myapp.example.com {
...
-				},
-			},
-		}
-
-		d := deploy.NewDeployment(cli, spec, nil)
-		_, err = d.Run(ctx)
-		require.Error(t, err, "Global deployment should fail when required volume doesn't exist")
-		require.Contains(t, err.Error(), "no machines available")
-	})
-
```
psviderski, line 1085: Does this test pass now?

pkg/client/deploy/scheduler/volume_test.go
```diff
@@ -802,6 +802,297 @@ func TestVolumeScheduler_Schedule(t *testing.T) {
...
+				"machine3": {
+					{
+						Name: "vol1",
+						Type: api.VolumeTypeVolume,
+					},
+				},
+			},
+		},
+		{
+			name: "global service with volume on some machines schedules remaining",
```
psviderski, line 867: That's actually an interesting one.

I'm thinking of a case when you deployed a global service with a volume on all machines. Then you added one more machine to the cluster. Then according to the new semantics, it feels like the next deployment of this global service should create a volume on the new machine and deploy a container there, not just skip the new machine. Otherwise, the behaviour feels inconsistent and again requires to manually create a volume as it was before this PR.

What's your opinion on this?

pkg/client/deploy/scheduler/volume.go
psviderski, line 12: Please update this description according to the changes

pkg/client/deploy/scheduler/volume_test.go
zasdaym: You're right. I'll fix.

pkg/client/deploy/scheduler/volume.go
zasdaym: Yes. I decided to create the missing volumes on the union of eligible machines. WDYT?

test/e2e/service_test.go
zasdaym: I don't know, I'm struggling to get the E2E tests working on my machine.

zasdaym:
A diagram or table for this global/replicated/constraint/volume relationship in docs would be great.

pkg/client/deploy/scheduler/volume.go
```diff
@@ -147,52 +156,99 @@ func (s *VolumeScheduler) Schedule() (map[string][]api.VolumeSpec, error) {
...
+	//    to be scheduled on machines that don't have them.
+	// 2. Volumes only used by global services - these need UNION of eligible machines, not intersection.
 	placedVolumes := make(map[string]struct{})
 	for volumeName := range s.existingVolumeMachines {
-		placedVolumes[volumeName] = struct{}{}
+		if !s.isVolumeForGlobalService(volumeName) {
+			placedVolumes[volumeName] = struct{}{}
+		}
+	}
+	// Skip constraint propagation for volumes only used by global services.
```
psviderski: Is this block with setting `placedVolumes[volumeName] = struct{}{}` needed for global only volumes? We exit with an error if a global and a replicated service share the same volume below so it seems we anyway won't run propagation for global-only volumes.
Maybe I'm missing something but the overall logic becomes harder to reason about.

psviderski: Ah, this is needed to not consider global volumes in the first `propagateConstraintsUntilConvergence` run. Ok makes sense. It's probably fine for this PR but I think if we were to add something else to this volume scheduling logic in the future, it makes sense to separate volume placement for global and replicated services as they became quite different now.

pkg/client/deploy/scheduler/volume.go
```diff
@@ -147,52 +156,99 @@ func (s *VolumeScheduler) Schedule() (map[string][]api.VolumeSpec, error) {
...
+	for volumeName := range s.volumeSpecs {
+		if s.isVolumeOnlyForGlobalServices(volumeName) {
```
psviderski: Please correct me if I'm wrong. Isn't `isVolumeOnlyForGlobalServices` redundant and equivalent to `!isVolumeSharedBetweenGlobalAndReplicated && isVolumeForGlobalService`?

We can also check volumes for `isVolumeSharedBetweenGlobalAndReplicated` in this loop and fail fast. Then check for `isVolumeForGlobalService` and mark as placed.

Or are you also addressing the potential future case when another mode could be added later?

test/e2e/service_test.go
psviderski: I've rebuilt the `ghcr.io/psviderski/ucind` image (used for running uncloud in docker in e2e tests) so now e2e tests should work locally. You can also build it yourself with `make ucind-image`. It's required if there were changes in the daemon code. Hope this helps

I believe we should keep this test as the `Deployment` doesn't do the volume scheduling so the deployment should fail.

pkg/client/deploy/scheduler/volume.go
psviderski: Yeah, the union sounds reasonable, thanks! 👍

psviderski:
The change looks great now, thank you for the updates! Please reply to/address the last minor comments and we can merge this.

pkg/client/deploy/scheduler/volume.go
zasdaym: Yes, that was my original intention. But, let's do it as simple as possible for now.

test/e2e/service_test.go
zasdaym: Test passed.

Issue #221, opened 2025-12-16
Support ulimits for service containers ('uc run' flag and Compose file)


psviderski:
Relevant discussion: https://github.com/psviderski/uncloud/discussions/220

Add support for specifying `ulimits` for service containers:
- `uc run` flag `--ulimits`. Use the same format as in `docker run` but provide the format description and examples in the flag description
- `ulimits` in a Compose file

The flag and the Compose field could be mapped to the [`ContainerResources`](https://github.com/psviderski/uncloud/blob/3b4bcb7932f65bb651160e76fa065a95666c0337/pkg/api/resources.go#L12-L22) spec in uncloud's `ContainerSpec`. Use the docker's `Ulimit` Go type for convenience.

psviderski:
- `--ulimits` flag for `uc run` added in https://github.com/psviderski/uncloud/commit/7f92a5db27d5d10f1bd7a04e4679af5ca393f08e
- `ulimits` in Compose implemented in https://github.com/psviderski/uncloud/pull/244 and released in https://github.com/psviderski/uncloud/releases/tag/v0.17.0

PR #244, opened 2026-01-13
feat: add support for ulimits in compose and container creation (#221)

ipaddicting:
Changes included:

- added support for `ulimits`
- added `GEMINI.md`
- gitignored added `.devcontainer`


pkg/api/service.go
psviderski, line 249: Given that we have the nested `Resources` struct with system resources for a container, I think it makes sense for the `Ulimits` to put there as well as ulimits are also resource constraints (file descriptors, processes, memory locking).

What do you also think of using a map in a similar way Compose uses (maybe without the `Single` field, just soft and hard) instead of a slice? This way we don't need to check if there are any items with the same name. I'm not sure how Docker handles that. This will also be more aligned with the `Sysctls` type and simplify spec comparison.

GEMINI.md
```diff
@@ -0,0 +1,258 @@
+# AI.md - Uncloud Project Guide
```
psviderski: Doesn't gemeni need the agent file to be called `GEMENI.md` and it doesn't automatically detect the current `AI.md`? Maybe we need to call it `AGENT.md` for it to pick it up? I honestly didn't invest much time in the agent setup yet so not sure what the latest best practices are.

@jabr what's your setup with gemeni?

In the case we do need to have a file called `GEMENI.md` then maybe we can create a symlink like we did for `CLAUDE.md` instead of duplicating the file?

ipaddicting: By default, the gemini-cli will use `GEMINI.md`, and the name of the context file can be changed: [Customize the context file name](https://geminicli.com/docs/cli/gemini-md/).

To avoid potential conflicts, I duplicated the `CLAUDE.md`... But I guess both CLAUDE & GEMINI should be smart enough to share the same context.

pkg/api/service.go
ipaddicting: Agreed, will do that change.

ipaddicting:
I've updated the PR with the requested changes:
- Moved `Ulimits` to `ContainerResources` and changed it to `map[string]Ulimit`.
- Updated `Ulimit` struct to remove the `Name` field.
- Updated Docker Compose integration and Docker execution logic accordingly.
- Cleaned up sorting logic where it's no longer needed.
- Verified with tests.

GEMINI.md
psviderski: Yeah, let's remove this file in this PR and I will renamve `AI.md` to `AGENTS.md` according to https://agents.md/ which apparently becomes a standard.
The gemini CLI can be easily configured to automatically use `AGENTS.md`: https://agents.md/#faq

ipaddicting:
> Excellent! ❤️ Please remove the duplicate `GEMINI.md` as I will replace it with `AGENTS.md` according to https://agents.md/ which apparently becomes a standard. And we're good to merge this PR.

`GEMINI.md` removed. 😀

PR #245, opened 2026-01-20
add support for service.labels and deploy.labels

nick-potts:
Two label types with different update semantics:                                                                                                                                                                                                           
                                                                                                                                                                                                                                                             
  - Labels (from service.labels) — applied to Docker containers. Changes require container recreation.                                                                                                                                                       
  - DeployLabels (from deploy.labels) — stored in service spec only. Can be updated without touching containers.                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
  User labels cannot override uncloud.* prefixed labels.                                                                                                                                                                                                     
                                                                                                                                                                                                                                                             
  CLI                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                             
  - uc run --label key=value — set container labels                                                                                                                                                                                                          
  - uc service inspect — displays both label types


nick-potts:
I *believe* deploy labels like this is a good choice and somewhat mirrors swarm. You can then use something like deployment_id under a deploy label, and it won't require re-creating a stateful resource like postgres.

Annoyingly in Swarm, deploy.labels are "service labels" (metadata on the service itself), while service.labels are actually "container labels" (applied to containers) and is quite confusing. Maybe its worth considering whether the ServiceSpec field names could be more explicit - e.g., ContainerLabels vs ServiceLabels to avoid the same confusion.

Issue #120, opened 2025-09-10
[BUG] misleading volume warning with `deploy` and missing node


luislavena:
**Describe the bug**

Attempt to deploy a compose project with a node down shows a misleading warning about unable to retrieve list of volumes of the powered off node:

```
$ uc --uncloud-config ./blatta.yaml deploy -f jokes.yaml
WARNING: failed to list volumes on machine 'fdcc:3d5:76c5:7014:6f65:ab45:6115:6966': rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [fdcc:3d5:76c5:7014:6f65:ab45:6115:6966]:51000: i/o timeout"
Deployment plan:
- Deploy service [name=jokes]
  - blatta1: Run container [image=myjokes:latest]

Do you want to continue?

Choose [y/N]:
```

There are no volumes defined in the nodes or part of the compose file (see below the specific file)

**How to reproduce**

```yaml
# jokes.yaml
---
services:
  jokes:
    image: myjokes:latest
    user: 5001:5001
    environment:
      - DB_PATH=/tmp
    x-machines:
      - blatta1
    x-ports:
      - jokes.example.com:3000/http
```

```yaml
current_context: default
contexts:
  default:
    connections:
      # blatta1
      - ssh: provision@192.168.1.203
      # blatta2
      - ssh: provision@192.168.1.194
      # blatta3
      - ssh: provision@192.168.1.246
```

1. Have a working uncloud cluster and validate configuration
2. Run `uc deploy -f jokes.yaml` once to validate it runs without warning
3. Poweroff/shutdown second or third instance of the cluster
4. Run `uc deploy` again
5. Notice the warning with the i/o timeout about the second instance

Also note that not trying to deploy to the instance that is down, but instead `blatta1` which is operational.

**Expected behavior**

**Environment:**

- Uncloud versions:
  - Control (client) node (`uc --version`): 0.12.1 (darwin/arm64)
  - Uncloud daemon (from the server) (`uncloudd --version`): 0.12.1 (linux/amd64)
- OS version (`uname -a`):
  - Client (control node): Darwin Kernel Version 24.6.0
  - Server: Debian 6.12.43-1 (2025-08-27) x86_64 GNU/Linux

---

Thank you
❤️ ❤️ ❤️

psviderski:
It's indeed misleading. To be honest, I haven't really tested the different failure modes yet. So it's very good that you're testing this now, thank you!

[...]

PR #247, opened 2026-01-26
refactor: Improvements to machine proxy calls and response metadata

jabr:
_Update: This is now complete, I believe, and ready for review. I have my cluster running on it now with no problems so far._

Moves the machine id/name resolution to the server and returns metadata with machine addr, id, and name for responses (where appropriate). This streamlines machine name mapping and error and edge case handling of `ProxyMachinesContext` to look like this:

```
mctx := cli.ProxyMachinesContext(ctx, nil)
machineContainers, err := cli.Docker.ListServiceContainers(mctx, ...)
for _, msc := range machineContainers {
  machineName := msc.Metadata.MachineName
  ...
}
```


psviderski:
Thank you for taking a look at this which I agree needs more love and clean up. The direction looks good but I think we can do even better by hiding all the details in the grpc-proxy layer: https://github.com/psviderski/uncloud/tree/79dc05cb665d6541ea2ff1ca16af4119ae6b47d4/internal/machine/api/proxy

I left this TODO to explore that path: https://github.com/psviderski/uncloud/blob/79dc05cb665d6541ea2ff1ca16af4119ae6b47d4/pkg/client/client.go#L86
The idea is to move the machine resolution logic from the client to the server which should simplify the API and reduce the number of requests.

At the moment `ProxyMachinesContext` sets the `machines` metadata field on the request to a list of internal IPv6 machine addresses. This was done that way to make it easy to implement the proxy layer which was new to me so I didn't want to overcomplicate it from the beginning. It simply extracts the addresses from the metadata and routes requests there: https://github.com/psviderski/uncloud/blob/da3634b690eb4917bae03d4cc0b763152fed729d/internal/machine/api/proxy/director.go#L41-L85

It's simple and agnostic from knowing anything about machine IDs/names and how to map them to network addresses. The obvious downside is that now every client needs to handle this. Another downside is when logging errors we again need to map network addresses to meaningful machine names.

Now, when the proxy layer works pretty well, we can iterate on it and make it smarter. How about we add a mapper to the `Director` directly. We can still abstract it with a mapper interface so that the proxy won't know about the local corrosion db and the subscription logic to watch for machine changes.

The client will be able to set `machines` to a list of machine IDs or names (we can discuss if supporting both needed/desired later). Then the director will use the mapper to translate these names to addresses it needs to create remote backends. The mapper will watch corrosion to keep its mapping up to date.

Something similar needs to be done for responses as well. We can map ipv6 addresses back to machine IDs/names and set them in the response `Metadata`: https://github.com/psviderski/uncloud/blob/b23699470b22429c29a079ad56c9e82018561642/internal/machine/api/pb/common.proto#L10-L19
This way the client will only work with and see machine IDs/names everywhere.

Moreover, we can even implement a shortcut for broadcasting to all machines, e.g. set something like `machines: *` in the request metadata. Then the client won't even need to `ListMachines` to request all machines (the mapper knows the full list). But this is more like nice to have so could be done as a follow-up if we really want to.

I also remember we have non-trivial error handling logic when broadcasting requests to multiple machines. I hope it can also be simplified by unifying the proxy layer.

What do you think about this approach?

jabr:
Yeah, that makes sense. It was a little unwieldy to wrap the current implementation, and that probably should have been a sign that I'd missed a better refactor. 😆 

I'll take a look at moving it into the `Director` layer next. I should have some more time this week.

psviderski:
Awesome! Let me know if you need any guidance or want to discuss something

jabr:
Curious about this bit in `Director` where there is `"machines"` metadata:
```
if len(backends) == 1 {
	return proxy.One2One, backends, nil
}
return proxy.One2Many, backends, nil
```

I'm guessing that is where the special case for "one machine" `ProxyMachinesContext` usage comes from (and the need for the `if r.Metadata == nil` check in callers). 

Just wondering if that optimization is worth keeping... how much overhead does the `One2Many` proxy add there? A quick scan of [the proxy code](https://github.com/siderolabs/grpc-proxy/tree/main/proxy) looks like it's pretty similar, so I would be surprised if it's not negligible compared to the network overhead.

psviderski:
I can't confidently say off the top of my head. Maybe you can't simply use `One2Many` for rpc calls that are not designed to support broadcast. To support broadcasting requests, the response should be designed upfront, e.g. https://github.com/psviderski/uncloud/blob/79dc05cb665d6541ea2ff1ca16af4119ae6b47d4/internal/machine/api/pb/docker.proto#L82-L85

I was mainly replicating the usage in Talos Linux: https://github.com/siderolabs/talos/blob/e48c6d7ab9c8a2e28ebe2115ac09f1557bbcca33/internal/app/apid/pkg/director/director.go#L50-L94 and didn't try to optimise it since then.

jabr:
Okay, I had the AI take a pass at the server-side machine id/name resolution. It's got some rough spots still, but the basic idea is there now.

edit: I think there's a simpler solution to the comments below with a fairly simple refactor to the "machine directory" and how the proxy director/backends use it. See https://github.com/psviderski/uncloud/pull/247#discussion_r2730399972

---
One deeper issue I noticed: the directory maintains a cache of machine backends, which now have the name (along with ID as well and the management IP addr they originally had), but that isn't updated if a machine is renamed. (Can the IP or ID also change?)

Thinking the proxy backends might need to interact with the new `MachineDirectory` to get the current name for the response rather than storing it itself. But that also means we might want to cache the machine list in the directory, since it would need the names both at the start to determine which backends to call and at the end to attach metadata.

Perhaps it should work like the Cluster code does with a machine list and then subscribe for updates to maintain its peer list. And if so, maybe those two pieces should be sharing a common, dynamic corrosion machine list...

cmd/uncloud/machine/rm.go
```diff
@@ -55,14 +55,21 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
...
+	if len(allMachines) == 0 {
 		return fmt.Errorf("machine '%s' not found in the cluster", nameOrID)
 	}
-	m := machines[0].Machine
+	m := allMachines[0].Machine
+
+	mctx, err := client.ProxyMachinesContext(ctx, []string{m.Id})
+	if err != nil {
+		return err
+	}
```
jabr: This looks wrong at first glance. I think the AI got confused along the way in these refactors.

internal/machine/api/proxy/directory.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (d *CorrosionDirectory) ListMachines(ctx context.Context) ([]*pb.MachineInfo, error) {
+	return d.store.ListMachines(ctx)
+}
+
+func (d *CorrosionDirectory) ResolveMachine(ctx context.Context, nameOrID string) (string, string, netip.Addr, error) {
+	machines, err := d.store.ListMachines(ctx)
+	if err != nil {
+		return "", "", netip.Addr{}, fmt.Errorf("list machines: %w", err)
+	}
```
jabr: Haha. ✨ 🤣 Well, this needs to be cached through at least the loop of calls to this...

internal/machine/api/proxy/directory.go
```diff
@@ -0,0 +1,104 @@
...
+	return nil, nil
+}
+
+func (m *MockDirectory) ResolveMachine(ctx context.Context, nameOrID string) (string, string, netip.Addr, error) {
+	if m.ResolveFunc != nil {
+		return m.ResolveFunc(ctx, nameOrID)
+	}
+	return "", "", netip.Addr{}, fmt.Errorf("machine not found: %s", nameOrID)
+}
```
jabr: Ugh. Not sure what it was thinking with this... maybe using it for unit tests?

internal/machine/api/proxy/directory.go
```diff
@@ -0,0 +1,104 @@
...
+
+// IsAllMachines returns true if the machine name/ID indicates all machines in the cluster.
+func IsAllMachines(nameOrID string) bool {
+	return nameOrID == "*" || strings.ToLower(nameOrID) == "all"
```
jabr: Should only match `"*"` I think. `"all"` _could_ be a machine name.

internal/machine/api/proxy/director.go
```diff
@@ -64,34 +71,68 @@ func (d *Director) Director(ctx context.Context, fullMethodName string) (proxy.M
...
+		}
+	} else {
+		for _, name := range machineNames {
+			id, mName, ip, err := d.directory.ResolveMachine(ctx, name)
+			if err != nil {
+				return proxy.One2One, nil, status.Error(codes.InvalidArgument, fmt.Sprintf("failed to resolve machine %s: %v", name, err))
+			}
+			targets = append(targets, target{id: id, name: mName, addr: ip.String()})
+		}
+	}
```
jabr: This logic should probably move into the `Directory` -- a `ResolveMachines` that just returns a list of addresses.

It could then do a single `store.ListMachines` at the start and hold on to the most recent list, and then the backends could call another method on it to get the name/ID for its addr (using that stored last list).

That would solve the machine name change issue, too.

pkg/client/volume.go
```diff
@@ -55,41 +56,53 @@ func (cli *Client) ListVolumes(ctx context.Context, filter *api.VolumeFilter) ([
...
 		return nil, fmt.Errorf("create request context to broadcast to all machines: %w", err)
 	}
 
-	machineVolumes, err := cli.Docker.ListVolumes(listCtx, volume.ListOptions{})
+	optsBytes, err := json.Marshal(volume.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("marshal options: %w", err)
+	}
+
+	machineVolumes, err := cli.Docker.GRPCClient.ListVolumes(mctx, &pb.ListVolumesRequest{Options: optsBytes})
```
jabr: Not sure what it was doing here.

pkg/client/volume.go
```diff
@@ -55,41 +56,53 @@ func (cli *Client) ListVolumes(ctx context.Context, filter *api.VolumeFilter) ([
...
+
+		machineID := mv.Metadata.MachineId
+		machineName := mv.Metadata.MachineName
+
+		if machineID == "" {
+			// We need the machine ID to construct the MachineVolume response.
+			// If we couldn't resolve the machine from the metadata (IP), we skip it.
+			PrintWarning(fmt.Sprintf("machine ID missing in response from: %s", mv.Metadata.Machine))
+			continue
+		}
```
jabr: I don't think this case could happen now. Machine ID should always be set.

psviderski:
>Can the IP or ID also change?

No, the machine ID is generated and stored once when the machine is initialised as well as the IP (which is derived from the first 14 bytes of the machine public key): https://github.com/psviderski/uncloud/blob/35d0a9012549d65b298385777b151c7ca1f3a6ad/internal/machine/network/ip.go#L16-L22
The public key cannot be changed as well.

>Perhaps it should work like the Cluster code does with a machine list and then subscribe for updates to maintain its peer list. And if so, maybe those two pieces should be sharing a common, dynamic corrosion machine list...

Yeah, I think abstracting these `remoteBackends` into a something I called a mapper earlier is how we can consolidate the current caching and Corrosion watching logic in one place: https://github.com/psviderski/uncloud/blob/da3634b690eb4917bae03d4cc0b763152fed729d/internal/machine/api/proxy/director.go#L17

It can expose just one method: "get a backend by machine ID/name". And it can cache them internally as `remoteBackends` map does now as well as invalidate the cache and keep the mapping in sync with machines in Corrosion. The directory itself won't care about all this logic, it will continue calling `.remoteBackend(machineNameOrID)` (give me the backend for this machine).

We can start passing the machine ID/name to the remote backend so that it includes it in the metadata as is: https://github.com/psviderski/uncloud/blob/f86e63c851b4f05e2199f5a9c07661acd0abf654/internal/machine/api/proxy/remote.go#L41. So if a user passed an ID in the requests, the same ID will come in the response. Same for the name if we want to support both.

Note one more things is that the grpc-proxy establishes permanent connections to backends and doesn't close them (well, maybe there is a long deadline but I don't know the details). If a client communicated with a machine and it gets renamed later, the proxy will still hold a connection to it. I don't know how it identifies if it needs to establish a new grpc connection or not, maybe just based on the `target` address.
I expect it should correctly handle the case when the director starts getting a backend with the same `target` address that contains a changed `machine` (new name) but this should be properly tested.

internal/machine/api/proxy/director.go
psviderski: See another message https://github.com/psviderski/uncloud/pull/247#issuecomment-3803351909

I think we should try to encapsulate all the logic related to machine resolution in one component + `RemoteBackend` itself. And almost don't change anything in the director to keep it simple and manageable.

psviderski: Maybe I'm missing something but this is what I'd try to do

jabr: Yeah, I think director-mapper-backend interaction just needs some refactoring to clean this up (and handle the machine rename case).

This last commit was some AI slop to look at the approach, and it probably wasn't worth the tokens. 😆

psviderski: I didn't look at the code tbh, only replying to your messages haha 😄

jabr:
> Maybe you can't simply use One2Many for rpc calls that are not designed to support broadcast. To support broadcasting requests, the response should be designed upfront

Ah, I see. Some of these ProxyMachines uses are always just calling with one machine and their proto doesn't have the metadata field... hmm.

Three options coming to mind:
1. Keep the current behavior with metadata-less `one2one` invocation for single machine, but that means we need to handle the nil metadata case again in the client commands.
2. Update those other protos to add a metadata field.
3. Specify the target machine with a different call metadata property (or add a new flag option) to get the current metadata-less response behavior just when we want it (and not in the unintended single machine but want metadata cases)

Personally, I like option 3 (feels like it'll be the smaller change between 2 and 3), and I'd really like to get rid of the current option 1, nil check cases.

psviderski:
100% agree that metadata nil check is annoying. I'm not sure I understand what you meant in option 3. Can you please elaborate?

So if the metadata is missing only for a one2one request, can we explicitly set it for such a request to behave the same as one2many? Not sure what is actually responsible for that, `LocalBackend`?

This is what I guess I was alluding to here: https://github.com/psviderski/uncloud/blob/1e6aaf451e431d3bc43f8d05f3d36d74f27ddccb/pkg/client/image.go#L76-L81

Then one2one won't be an edge case and could be handled in the same way as one2many in the client (always expecting the `Metadata` to be set). Each machine knows what ID or name it is, so we should be able to inject this information in a response at the proxy layer.

psviderski:
Ah, we probably can't do this for a generic case because the proxy layer doesn't know the type of a response whether it includes a `Metadata` field or not. Need to think more about it

jabr:
> Ah, we probably can't do this for a generic case because the proxy layer doesn't know the type of a response whether it includes a Metadata field or not.

Yeah, my option three was to basically have the cli caller tell the server if it wants metadata or not. 

And since the “no metadata” cases seem to always be “proxy to a single machine”, I was thinking it might just be a different key used in the call:
- “machine**s**” is 1 or more id/names and should return metadata
- “machine” is just 1 id/name and returns no metadata

psviderski:
Right, thanks for clarifying. Hm, that sounds clever. Ah, isn't this the reason why Talos also supports both `node` and `nodes`? https://github.com/siderolabs/talos/blob/e48c6d7ab9c8a2e28ebe2115ac09f1557bbcca33/internal/app/apid/pkg/director/director.go#L60-L61. I didn't get it when I tried to understand the logic.
I like your approach to instruct on the client how the server should handle the requests 👍

cmd/uncloud/machine/rm.go
```diff
@@ -55,22 +55,30 @@ func remove(ctx context.Context, uncli *cli.CLI, nameOrID string, opts removeOpt
...
 	if err != nil {
 		return fmt.Errorf("inspect proxy machine: %w", err)
 	}
 	if proxyMachine.Id == m.Id {
-		allMachines, err := client.ListMachines(ctx, nil)
+		// Use already fetched machines or list them if needed?
+		// We can just list machines to see if there are others.
+		var allMachines api.MachineMembersList
+		allMachines, err = client.ListMachines(ctx, nil)
 		if err != nil {
```
jabr, line 84: Need to look at this closer again. Thinking it might be better to just do a ListMachines at the start to combine the existence check and presence of more than one machine when trying to remove the connection machine.

internal/machine/api/proxy/backend.go
```diff
@@ -4,65 +4,27 @@ import (
...
-//	}
-//
-//	message SomeReply {
-//	  common.Metadata metadata = 1;
-//	  <other fields go here ...>
-//	}
-//
-// As 'SomeReply' is repeated in 'SomeResponse', if we concatenate protobuf representation
-// of several 'SomeResponse' messages, we still get valid 'SomeResponse' representation but with more
-// entries (feature of protobuf binary representation).
```
jabr, line 36: Undo comment deletions in this file

internal/machine/api/proxy/director.go
```diff
@@ -18,24 +19,29 @@ type Director struct {
...
+func (d *Director) UpdateLocalMachine(id, name, addr string) {
 	d.mu.Lock()
 	defer d.mu.Unlock()
 
 	d.localAddress = addr
-	// Replace the local backend with the one that has local address set.
-	d.localBackend = NewLocalBackend(d.localBackend.sockPath, addr)
+	d.localID = id
+	d.localName = name
+	// Local backend doesn't need re-creation as it doesn't store metadata anymore.
```
jabr: Revert this.

internal/machine/api/proxy/director.go
```diff
@@ -18,24 +19,29 @@ type Director struct {
 	// mu synchronizes access to localAddress.
 	mu           sync.RWMutex
 	localAddress string
+	localID      string
+	localName    string
```
jabr: Remove. No longer used.

internal/machine/api/proxy/director.go
jabr: Oh. Actually might not need to recreate the backend since it no longer keeps track of the addr. `d.localAddress` takes care of the mapping to it, and the actual backend implementation connection (via unix socket) doesn't change when this address does.

jabr: I think we can get rid of the r/w locks around the local address swap now, too, since they no longer need to be changed together atomically.

internal/machine/machine.go
```diff
@@ -418,7 +419,7 @@ func (m *Machine) Run(ctx context.Context) error {
 			slog.Info("Starting cluster controller.")
 			// Update the proxy director's local address to the machine's management IP address, allowing
 			// the proxy to identify which requests should be proxied to the local machine API server.
-			m.proxyDirector.UpdateLocalAddress(m.state.Network.ManagementIP.String())
+			m.proxyDirector.UpdateLocalMachine(m.state.ID, m.state.Name, m.state.Network.ManagementIP.String())
```
jabr: revert

pkg/client/volume.go
```diff
@@ -55,41 +54,46 @@ func (cli *Client) ListVolumes(ctx context.Context, filter *api.VolumeFilter) ([
...
+	machineVolumes, err := cli.Docker.ListVolumes(mctx, volume.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
 	var volumes []api.MachineVolume
 	// Process responses from all machines.
 	for _, mv := range machineVolumes {
-		if mv.Metadata != nil && mv.Metadata.Error != "" {
-			// TODO: return failed machines in the response.
```
jabr, line 74: Move this todo comment down to the actual `Metadata.Error` check below. Not sure I know what needs to be done to complete this todo right now, though.

pkg/client/volume.go
jabr: Still think this is not needed. AI keeps ignoring this bit.

jabr:
Still needs a little more cleanup, but I added the machine/machines grpc call modes (for the no metadata cases) and refactored the Director/Mapper machine lookup (and how backends work with that and returning current machine name).

The overall approach is looking much better now, I think.


> (re: machine/machines separation) isn't this the reason why Talos also supports both node and nodes

Yeah, I think so. That Talos code looks vaguely similar to the new Director implementation, and it definitely seems to be dealing with the same sorts of issues (one2one vs one2many proxying for different cases).

jabr:
I was thinking about this MR, and it might need some thought about backwards compatibility (for at least a release or so). I think the current state would require client and daemon to both be upgraded to work. e.g. upgrading the cli would probably be broken against machines running a previous version with its expectations on metadata that would be returned.

psviderski:
I really appreciate your concerns! My thinking is that we're still at the stage where we can introduce breaking changes in a release (we almost haven't so far though), as long as we clearly communicate it. To do backward compatibility right is a lot of extra work and if we can skip it, that would be great.

This refactoring is only for interactive `uc` commands and it doesn't affect the runtime, right? Can you confirm that if you upgrade only the client or only daemons on machines all services will be working fine? It's only `uc` commands that will be broken.

If so, then we can just include a warning in the release notes that there're breaking changes and users have to upgrade both machines and the client to the latest release. Like this note on https://github.com/psviderski/uncloud/releases/tag/v0.16.0

Screenshot 2026-03-02 at 5 25 31 pm

Let's not overcomplicate it if you're ok with this. Maybe what we can do is for the new release CLI commands check the versions of the server and if its older than the CLI's version, fail with a clear message that servers need to be upgraded. But we don't have the required infra yet. Something like this https://github.com/psviderski/uncloud/pull/246 would be needed first. But then introducing an extra logic to fetch versions on each command run is also meh.

Ah, one more thought. Maybe we should test what errors the proxy/api layer will return when the new client communicates with the old server. Then somewhere in the common place in the client that covers all commands check for this error and provide some more information about it. Why this likely happens and how to resolve it -- upgrade to the latest version following these instructions.

This will at least cover the case when the client is upgraded automatically without the user knowing about it (e.g. with `brew` upgrade). It's less likely there will be an unattended upgrade for the uncloud daemon on machines.

jabr:
> This refactoring is only for interactive `uc` commands and it doesn't affect the runtime, right? Can you confirm that if you upgrade only the client or only daemons on machines all services will be working fine? It's only `uc` commands that will be broken.

Yes, it's just the cli communication with the daemons, and specifically just those commands using the internal grpc proxy. But I will be testing it further still on my cluster, as well.

> Let's not overcomplicate it if you're ok with this.

Agreed. A backward compatible approach to this change would get pretty hairy. We'd probably need to do it incrementally over a few releases if we did want to do that.

> Maybe what we can do is for the new release CLI commands check the versions of the server ... But then introducing an extra logic to fetch versions on each command run is also meh.

We'll probably need something like this eventually. Might make sense to do it first, if we can come up with a good approach.

One idea to avoid the extra "what version are you exchange": if the version is always in the "header" of grpc requests/responses, we could use that as an abort trigger on a minimum version supported of the other side. So for this MR, we'd bump the min version to its release. And then a newer client would immediately reject if the server's response was old, and a newer server would return an error immediately to the cli if it was old. Then we could just bump that min version when needed on subsequent releases (and it wouldn't necessarily have to be symmetrical -- cli and daemon could have different supported min versions of each other).

Might make sense to break out this part of the conversation into a separate issue.

> Maybe we should test what errors the proxy/api layer will return when the new client communicates with the old server.

This was my thinking if we don't do the min-version idea above first. Basically we just add some extra checks around areas that will break down with mismatched versions on this proxy change, and add "upgrade" warnings/errors.

Some of the edge cases are a bit tricky to detect, though, so I'm not sure if it's good enough. I had an AI do a quick summary of edge cases, which looks mostly correct (there are a couple details to confirm):

```
Version Mismatch Analysis

Key Changes in This Branch

1. ProxyMachinesContext Signature & Behavior Change
- OLD: Client resolves names/IDs → IPs locally, sends IPs in "machines" header, returns (ctx, MachineMembersList, error)
- NEW: Client sends names/IDs directly in "machines" header, server resolves via MachineMapper, returns context.Context

2. New "machine" Header (Singular)
- NEW: ProxyMachineContext() sets "machine" header for One2One routing without metadata injection
- OLD: Did not exist - all proxying used "machines" header

3. Metadata Message Changes
- OLD: Only machine field (which became machine_addr)
- NEW: Added machine_id, machine_name fields; machine renamed to machine_addr

4. Server-Side Machine Resolution
- NEW: MachineMapper interface resolves names/IDs/IPs to machine targets
- OLD: Server expected raw IP addresses in "machines" header

---

Failure Scenarios

Scenario 1: New CLI + Old Server (HIGH RISK)
What happens:
1. CLI sends "machines" header with machine names/IDs (e.g., "machine-1", "abc-123")
2. Old server receives request but doesn't have MachineMapper
3. Old server tries to use the name/ID as a raw IP address
4. FAILURE: Connection fails or routes to wrong destination
Error manifestation:
- If name isn't valid IP: Connection timeout/refused
- If name happens to be valid IP format but wrong machine: Silent routing to wrong machine
- If the server happens to be the target: Works accidentally
Additional issue:
- CLI expects Metadata with machine_id and machine_name fields
- Old server only returns machine (now machine_addr)
- Code checks for nil metadata and prints warning, but continues with empty machine info
- Result: Client code accessing mc.Metadata.MachineId gets empty string
Affected operations:
- uc ps (lists containers across machines) - containers appear without machine IDs
- uc volume ls - volumes appear without machine info
- uc image ls - images appear without machine info
- Service inspection - may fail to correlate containers with machines

Scenario 2: Old CLI + New Server (LOW RISK)
What happens:
1. Old CLI resolves names/IDs → IPs locally
2. Sends IPs in "machines" header
3. New server has MachineMapper that can resolve IPs back to machines
4. Result: Works correctly
Minor issues:
- Old CLI returns MachineMembersList which it resolved locally (redundant but harmless)
- Old CLI doesn't use new machine_id/machine_name fields, so it just ignores them

Scenario 3: New CLI + Old Server Using "machine" Header (CRITICAL)
What happens:
1. New CLI uses ProxyMachineContext() (singular) for One2One operations
2. Sets "machine" header with machine name/ID
3. Old server doesn't recognize "machine" header
4. Falls back to local-only backend
5. FAILURE: Operation runs on wrong machine (the server receiving the request, not the target)
Operations affected:
- CreateContainer - Container created on wrong machine
- CreateVolume - Volume created on wrong machine
- StartContainer, StopContainer, RemoveContainer - Operations fail (container not found on local machine)
- ExecContainer - Exec fails or runs on wrong container
- ContainerLogs - Wrong container logs returned
Error manifestation:
- container not found errors
- Operations succeed but on wrong machine
- Data inconsistency (volumes/containers on wrong machines)

Scenario 4: Mixed Machine Versions (Cluster with Old & New Daemons)
New CLI + Mixed Servers:
- Requests to new servers: Work correctly (can resolve names/IDs)
- Requests to old servers: Fail to resolve names/IDs
- Result: Partial failures, inconsistent cluster state
Old CLI + Mixed Servers:
- All requests send IPs
- New servers resolve IPs correctly
- Old servers use IPs directly
- Result: Works correctly (backwards compatible)

Scenario 5: Metadata Field Mismatch
New Client + Old Server:
- Server returns Metadata{machine: "10.0.0.1"} (old format)
- Client code expects Metadata{machine_id: "abc", machine_name: "machine-1", machine_addr: "10.0.0.1"}
- Fields machine_id and machine_name are empty
- Client code has fallback: if mc.Metadata == nil { PrintWarning("metadata is missing..."); continue }
- BUT: Metadata is NOT nil, just missing new fields
- Result: Operations continue with empty machine IDs/names, causing data corruption or lost tracking
Example in pkg/client/service.go:132:
machineID := mc.Metadata.MachineId  // Empty string from old server!
Impact:
- Container operations succeed but machine association is lost
- Subsequent operations targeting specific machines fail
- Cluster state becomes inconsistent

---

Error Reporting Gaps

Current Issues:
1. Silent failures: When old server returns metadata without new fields, code continues with empty values instead of failing
2. Ambiguous errors: "machine not found" could mean:
   - Machine doesn't exist in cluster
   - Server couldn't resolve the identifier
   - Network routing issue
3. No version detection: No way to detect server version mismatch early
4. Warning fatigue: Legacy fallback prints warnings that may be ignored:
      if mc.Metadata == nil {
       PrintWarning("metadata is missing in response from unknown server")
       continue
   }
5. Missing validation: No check that required metadata fields are populated
```

The new client sending a "machine" (not "machines") route to an old server is a pernicious one. It would appear to work perfectly fine, except that it ran it on the wrong machine. 😬

psviderski:
Love the idea of including the CLI/server version in the message headers. Looks like we're aligned. Don't stress too much about identifying all the edge cases. We'll quickly fix if we break something.
Let me know when you want me to take a closer look at the code.

jabr:
I've implemented a first draft of the version check idea in #260.

I thought it'd be better to do it as a separate PR to keep it easier to read. And it also gives us the option of releasing that before this PR to prepare users for a breaking change, min version bump release.

internal/grpcversion/interceptor.go
```diff
@@ -31,8 +31,8 @@ const (
 	//
 	// The two minimums are independent: a client might require a newer daemon for new
 	// features, while that same daemon could still handle requests from older clients.
-	MinClientVersion = "0.0.0"
-	MinServerVersion = "0.0.0"
+	MinClientVersion = "0.20.0"
+	MinServerVersion = "0.20.0"
```
jabr, line 35: Note: we'll want to bump these versions to whatever release this ends up going out in.

cmd/uncloud/machine/rtt.go
```diff
@@ -51,6 +48,15 @@ func rtt(ctx context.Context, uncli *cli.CLI) error {
 	// Map machine IDs to names for display from the response.
 	machineNames := make(map[string]string)
 	for _, m := range resp.Machines {
+		// NOTE: Metadata should never be nil in practice. This is legacy fallback that will be removed.
```
psviderski, line 51: Is it safe to remove this now if we bump the min client/server version in the next release?

pkg/client/image.go
```diff
@@ -80,32 +78,35 @@ func (cli *Client) ListImages(ctx context.Context, filter api.ImageFilter) ([]ap
...
-			}
-			if msg.Metadata.Error != "" {
-				continue
-			}
+			tui.PrintWarning("metadata is missing in response from unknown server")
+			continue
+		}
+
+		if msg.Metadata.Error != "" {
+			// TODO: any reason to not return these and let the caller decide what to do?
```
psviderski: I guess this is to allow you troubleshoot the cluster when some machines went down. Otherwise, if one machine is down or not responding, you will start getting an error for the CLI command that this machine is down and will not able to see what is going on on other machines.

Or do you mean to return the error as part of the `Metadata` in `MachineImages` slice?

psviderski (approved):
I really enjoyed reading this change. Nice work!

Love how you simplified backends with a nice `MetadataBackend` wrapper. one2one and one2many are very clean separate scenarios now.

Calling `ListMachines` from store on every request resolution should always provide the most up-to-date data. I just left a TODO that we might want to implement some cleanup logic for outdated remote backends: https://github.com/psviderski/uncloud/pull/247/changes/af6a3d3fe359b3ae1e6c374146f6309fbbfef29b#diff-1e0051ecd0d31e89e53e49e3cd411058e613b83861753e2dd360d3125636631fR108-R112

I wish we implemented it this way from the very beginning so we didn't have to migrate. But you never know and it's better to be prepared for the migrations to iterate and improve rather than to try to avoid them.

cmd/uncloud/machine/rtt.go
jabr: Yeah, I think so. 

It would only happen if someone was using a new client with a 0.17 or older (pre version check) daemon. The metadata would be nil in the single server "one2many" case and throw an exception.

But if the daemon is on 0.18 or 0.19, it would just reject immediately at the start of the request with the 0.20 client.

If we think enough time has passed for that, it's probably fine to remove now. Or we can do it in the 0.21 release, and remove all the `TODO: remove later` bits from the version check code, as well as this.

pkg/client/image.go
jabr: Ah, yeah, I think I was wondering why not just return that entry with its error along with the other ones that did work (not return an error for the whole function, just not drop those), and then the code calling this could decide what to do with it.

Not critical by any means. More of an idle thought there (and some other places if I recall) where I was thinking we might be able to do something nicer in the cli output rather than print those warnings inline during processing like that... something to think about for a followup maybe.

jabr:
`ProxySingleMachineContext`/`ProxyMachinesContext` is a much better naming than my `ProxyMachineContext`/`ProxyMachinesContext`!

The single letter distinction always felt wrong, but the clearer name for the singular case never occurred to me for some reason. 😂

cmd/uncloud/machine/rtt.go
psviderski: It's not a big deal, let's do the cleanup in 0.21 then. The interceptors could be cleaned up there as well.

pkg/client/image.go
psviderski: Right, we should probably just return that entry with its error along with the other ones in this particular case. Regarding the warnings in other places, I keep kicking this can down the road as I don't know what a better approach is yet. I shared some thoughts earlier here https://github.com/psviderski/uncloud/issues/120#issuecomment-3277054429.

Working out a more or less generic and safe approach for handling unavailable machines would be nice as this can significantly affect the result of `uc` operations. Feel free to share your opinions on the topic or just load it in the background hoping to come up with something later.

Issue #151, opened 2025-10-17
Support configuring 'stop-first' order for rolling deployments


psviderski:
A persistent deployment such as database require its containers to be stopped first before starting new ones to not corrupt its data.

We should support `order: stop-first` in `deploy.update_config` in Compose: https://github.com/compose-spec/compose-spec/blob/main/deploy.md#update_config. The current behaviour is always `start-first` and uncloud just ignores the `update_config`.

We should also consider defaulting to `stop-first` for all deployments using a data volume.
Suggested by `nulo.lol` on discord:
>i would assume that any service with volumes should default to stop-first, but i might be wrong. (it could make sense to have start-first for something like a WALed SQLite database, but i think it shouldn't be the default)

psviderski:
Related issue https://github.com/psviderski/uncloud/issues/24 may need to be implemented first to have a robust primitive for implementing start/stop ordering.

psviderski:
The update order using a new `ReplaceContainerOperation` has been implemented in https://github.com/psviderski/uncloud/pull/248. 
The missed health check monitoring for a new container and rollback if it fails to start or become healthy in `ReplaceContainerOperation` implemented in https://github.com/psviderski/uncloud/commit/7a9eac593f9be0f02b42b1bef5b82f311e6af243

psviderski:
Update docs: [Rolling deployments](https://uncloud.run/docs/guides/deployments/rolling-deployments/)

PR #248, opened 2026-01-26
`update_config` order support - start_stop and stop_start

### LINKED ISSUE #24: Introduce ReplaceContainerOperation to roll out a container and rollback on failure
It will help encapsulate the logic for replacing a service container with a new one (start-first or stop-first) and restart the old container in case a new one fails to start.

nick-potts:
An initial first pass of this. Once depends_on is supported it'd be great to also do parallelism. 

resolves #24 and #151

### DIFF
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -19,6 +19,13 @@ const (
 	ServiceModeReplicated = "replicated"
 	ServiceModeGlobal     = "global"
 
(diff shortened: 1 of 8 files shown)


nick-potts:
Also perhaps the volume stuff is trying to be too clever and it should just be easier defaults to comprehend.

PR #251, opened 2026-02-13
Add support for Rancher Desktop

mitio:
Rancher Desktop is a free alternative to Docker Desktop (which prohibits commercial use) and has gained significant popularity. When running Rancher Desktop on non-native Docker platforms such as macOS and Windows, the secondary tunnel from the host OS to the Docker's Linux machine is necessary. At the moment, the check for this seems to be based on the vhost name of the docker engine's VM.

For Rancher Desktop it seemms to be `lima-rancher-desktop`:

```
$ docker info --format 'json' | jq .Name
"lima-rancher-desktop"
```

This might not be the best approach but is simple and should save a lot of headaches for a lot of people until a potentially better solution is devised.

See also: https://github.com/psviderski/unregistry/issues/62

### DIFF
--- a/pkg/client/image.go
+++ b/pkg/client/image.go
@@ -370,7 +370,7 @@ func isDockerVirtualised(ctx context.Context, dockerCli *docker.Client) (bool, e
 		return false, fmt.Errorf("get Docker info: %w", err)
 	}
 
-	virtualisedHostnames := []string{"docker-desktop", "colima"}
+	virtualisedHostnames := []string{"docker-desktop", "colima", "lima-rancher-desktop"}
 	for _, name := range virtualisedHostnames {
 		if strings.Contains(strings.ToLower(info.Name), name) {
 			return true, nil


jabr:
I'm not entirely clear on when this comes up (when using your laptop/desktop as a machine?), but do we want to also include `orbstack` if we're adding to this list? It's another alternative to Docker Desktop like Rancher. Any others to include?

```
$ docker info --format json | jq .Name
"orbstack"
```

mitio:
@jabr apologies for not providing enough context.

This happens when I try e.g. `uc deploy` from my dev machine (macOS) that runs Rancher Desktop, to an uncloud cluster/machine that might be running Linux. What matters is the source, i.e. my dev machine, as this is where images are built and deployed from.

[This thead and specifically this comment](https://github.com/psviderski/unregistry/issues/13#:~:text=A%20possible%20simplification) have a bit more context on potential alternative solutions.

PR #260, opened 2026-03-04
feat: Add version check mechanism to gRPC calls

jabr:
This adds a minimum version mechanism to gRPC communication to ensure CLI and daemon versions are compatible. It enables future breaking changes by allowing minimum version requirements to be enforced.

See [#247](https://github.com/psviderski/uncloud/pull/247#issuecomment-3989558618) for example use-case.

Changes:
- New `pkg/versioncheck` package with gRPC interceptors
- Client interceptors inject version metadata (`x-uncloud-cli-version`, `x-uncloud-min-daemon-version`) into requests
- Server interceptors validate client version headers and respond with `x-uncloud-daemon-version`
- Initial minimum versions set to "0.0.0" for backward compatibility with pre-version-check clients/daemons
- After first real min-version bump occurs, client warns if daemon response is missing version metadata (useful during initial version-check release transition; once user has upgraded to version checking daemons it will preemptively reject the client request)
  * In the current version, at min 0.0.0, it doesn't warn in order to not be unnecessarily annoying since they are presumably compatible versions. The hope is that people would have upgraded both client and daemon sometime before the first real min-version bump is necessary. But we could change this to start warning right away when a new client is used with a old server in order to prompt people to upgrade...

Behavior:
- Server rejects requests if CLI version is below minimum or daemon version is below CLI's required minimum
- Missing version headers are treated as "0.0.0", which passes initial checks but will fail once minimums are bumped in future releases

Files changed:
- `pkg/versioncheck/interceptor.go` - new package with interceptors
- `pkg/client/connector/*.go` - added client interceptors
- `internal/machine/machine.go` - added server interceptors to proxy servers


jabr:
I've tested this on my cluster against a variety of combinations: old/current (no version check logic) client and server with new (with version check logic) client and server, as well as different future cases with different version dependency and current version combinations in client and server.

Note that this initial release has min client/daemon version requirements of `0.0.0` and it treats old clients/daemons with no version specified as implicitly `0.0.0` to make it backwards compatible for the initial release of this.

## old client <-> old server [ok, obviously]
```
$ uc ls
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

## min-versions 0.0.0 [all ok for backwards compatibility on initial release]

### new client <-> old server [ok]
```
$ ./dist/uncloud ls
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

### old client <-> new server [ok]
```
$ uc ls
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

### new client <-> new server [ok]
```
$ ./dist/uncloud ls
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

## min-versions 0.0.1 (or whatever the first real compatibility version break will be)

### new client <-> old server [warning because version was unchecked, but no way to prevent what happened]
```
$ ./dist/uncloud ls
WARNING: daemon version is below minimum required version 0.0.1. The daemon did not verify this CLI's minimum version requirement, so the operation may not have behaved as intended. Please upgrade the daemon: https://github.com/psviderski/uncloud/releases/latest
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

### old client <-> new server [incompatible]
```
$ uc ls
Error: list services: list machines: rpc error: code = FailedPrecondition desc = version check failed: client version is below minimum 0.0.1. Please upgrade: https://github.com/psviderski/uncloud/releases/latest
```

### new client <-> new server [ok]
```
$ ./dist/uncloud ls
NAME          MODE         REPLICAS   IMAGE              ENDPOINTS
caddy         global       1          caddy:latest       (custom Caddy config)
```

## min-client-version > client version [incompatible according to daemon]
```
$ ./dist/uncloud ls
Error: list services: list machines: rpc error: code = FailedPrecondition desc = version check failed: client version is below minimum 999.0.0. Please upgrade: https://github.com/psviderski/uncloud/releases/latest
```

## min-daemon-version > daemon version [incompatible according to client]
```
$ ./dist/uncloud ls
Error: list services: list machines: rpc error: code = FailedPrecondition desc = version check failed: daemon version 999.0.0-dev is below client's minimum required version 999.0.0. Please upgrade the daemon: https://github.com/psviderski/uncloud/releases/latest
```

jabr:
Also, tested the streaming case:

## new client, min-daemon=0.0.0 <-> old server
```
$ ./dist/uncloud logs wernicke -f
Mar  5 21:09:49.836 himalia wernicke[2e370] INFO:wernicke:→ GET /health | 200 (15)
Mar  5 21:10:19.905 himalia wernicke[2e370] INFO:wernicke:→ GET /health | 200 (15)
(15)
```
## new client, min-daemon=0.0.1 <-> old server
```
$ ./dist/uncloud logs wernicke -f
WARNING: daemon version is below minimum required version 0.0.1. The daemon did not verify this CLI's minimum version requirement, so the operation may not have behaved as intended. Please upgrade the daemon: https://github.com/psviderski/uncloud/releases/latest
Mar  5 21:07:49.572 himalia wernicke[2e370] INFO:wernicke:→ GET /health | 200 (15)
Mar  5 21:08:19.635 himalia wernicke[2e370] INFO:wernicke:→ GET /health | 200 (15)
```

## old client <-> new server, min-client=0.0.1
```
$ uc logs wernicke -f
Error: stream logs for service 'wernicke': inspect service: list machines: rpc error: code = FailedPrecondition desc = version check failed: client version is below minimum 0.0.1. Please upgrade: https://github.com/psviderski/uncloud/releases/latest
```

## new client <-> new server, min-client=999.0.0
```
$ ./dist/uncloud logs wernicke -f
Error: stream logs for service 'wernicke': inspect service: list machines: rpc error: code = FailedPrecondition desc = version check failed: client version is below minimum 999.0.0. Please upgrade: https://github.com/psviderski/uncloud/releases/latest
```

## new client, min-daemon=999.0.0 <-> new server
```
$ ./dist/uncloud logs wernicke -f
Error: stream logs for service 'wernicke': inspect service: list machines: rpc error: code = FailedPrecondition desc = version check failed: daemon version 999.0.0-dev is below client's minimum required version 999.0.0. Please upgrade the daemon: https://github.com/psviderski/uncloud/releases/latest
```

internal/machine/machine.go
```diff
@@ -421,6 +424,8 @@ func (m *Machine) Run(ctx context.Context) error {
 			m.proxyDirector.UpdateLocalAddress(m.state.Network.ManagementIP.String())
 			proxyServer := grpc.NewServer(
 				grpc.ForceServerCodecV2(proxy.Codec()),
+				grpc.UnaryInterceptor(versionpkg.ServerUnaryInterceptor),
+				grpc.StreamInterceptor(versionpkg.ServerStreamInterceptor),
```
psviderski, line 432: Not sure I understand why we put server interceptors on the proxies. Especially this non-local one that can proxy to different gRPC servers which could have different versions.

I'd put them on the local machine server: https://github.com/psviderski/uncloud/blob/51c7b5ba8f6d3511531fe2c8e6f8d9a496c68f66/internal/machine/machine.go#L303

The actual server that responds, not the proxies that just relay messages so that no matter where the request comes from, it goes through the interceptors on each machine. Am I missing something?

pkg/versioncheck/interceptor.go
```diff
@@ -0,0 +1,155 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	MetadataKeyCLIVersion       = "x-uncloud-cli-version"
+	MetadataKeyMinDaemonVersion = "x-uncloud-min-daemon-version"
+	MetadataKeyDaemonVersion    = "x-uncloud-daemon-version"
```
psviderski: How about using gRPC terminology for this grpc metadata? I.e.:
- `uncloud-client-version`
- `uncloud-min-server-version`
- `uncloud-server-version`

I guess `x-` can also be dropped if we're prefixing with `uncloud-`. Open telemetry uses even unprefixed headers `traceparent` and `tracestate`.

pkg/versioncheck/interceptor.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, err
+	}
+	return handler(ctx, req)
+}
+
+func ServerStreamInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	if err := checkClientVersionHeaders(ss.Context()); err != nil {
+		return err
+	}
+	if err := grpc.SetHeader(ss.Context(), metadata.Pairs(MetadataKeyDaemonVersion, currentVersion.String())); err != nil {
```
psviderski: `SetHeader` docs say
>Streaming RPCs should prefer the SetHeader method of the ServerStream.

 https://github.com/grpc/grpc-go/blob/master/server.go#L2083

Does this do the same as using `ss.Context()`?

pkg/versioncheck/interceptor.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func ClientUnaryInterceptor(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	ctx = metadata.AppendToOutgoingContext(ctx,
+		MetadataKeyCLIVersion, currentVersion.String(),
+		MetadataKeyMinDaemonVersion, MinDaemonVersion,
+	)
+
+	var respMD metadata.MD
+	opts = append(opts, grpc.Header(&respMD))
```
psviderski, line 116: Is this options really needed? I thought that just updating the context is enough: https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-metadata

pkg/versioncheck/interceptor.go
```diff
@@ -0,0 +1,155 @@
...
+
+	err := invoker(ctx, method, req, reply, cc, opts...)
+	if err != nil {
+		return err
+	}
+
+	checkDaemonVersionInResponse(respMD)
```
psviderski, line 124: Hm, I'm a bit confused who is responsible for checking what. We have the server interceptor that knows its own version and the min one required by the client received via the header. It may return `FailedPrecondition` if they mismatch.

Why do we need to check this on the client again?

pkg/versioncheck/interceptor_test.go
```diff
@@ -0,0 +1,235 @@
...
+				}
+				if st.Code() != tt.errCode {
+					t.Errorf("error code = %v, want %v", st.Code(), tt.errCode)
+				}
+				if !strings.Contains(st.Message(), tt.errContain) {
+					t.Errorf("error message = %q, want to contain %q", st.Message(), tt.errContain)
+				}
+			} else {
+				if err != nil {
+					t.Errorf("checkClientVersionHeaders() unexpected error: %v", err)
```
psviderski: Please tell your agent to use `testify` library in tests. Feel free to add this to `AGENTS.md` if it doesn't automatically pick up the style from the other code.

jabr:
> Ah, I just recalled the big change to the proxy layer to use a new format for machines header. I guess this is the case when you'd consider setting global min version, right?

Yes, this was intended for complex, backwards incompatible changes, possibly affecting many commands. A change like the proxy refactor doesn't just throw errors when things are mismatched, it can silently _do the wrong thing_ in some cases. For example, it could run a command on a different server than the user intended and return that result with no obvious errors or exceptions.

It was also meant to be a general solution for changes like it that may be required in the future. It's certainly overkill for something as simple as a new gRPC method missing, but it would enable forcing an upgrade if we ever need to change the "semantics" of an existing call. e.g. something with the deploy process that would be similarly painful to support multiple versions of for the sake of backwards compatibility.

Also, this implementation has some bits that are dealing with the initial transition from _no version info_ available, which is why things like the client check on the server response is needed to at least warn the user after the fact.

> Then make some direct grpc call from ProxyMachinesContext to extract the version of the server from the header in the response. 

That's the pre-flighting approach, which I was trying to avoid needing with this implementation. For the proxy refactor, that means a preflight check on a significant fraction of all calls the the cli makes.

jabr:
Also, for clarity, the reason for the two different "min-versions":

1. client sends `min-daemon-version`

This is so we don't need to preflight requests: the newer client code _knows_ it needs a daemon of at least version x.y.z to handle its request properly. And _any_ version-checking daemon is capable of recognizing it is below the threshold and can then abort with an immediate error.

2. server checks `min-client-version`

This is so the server can just tell the client it is too old and no longer supported. We could still support a transition phase of backwards compatibility with the server having conditional responses based on the client version (e.g. return v0.1 payload, v0.2 payload, etc).

But when we finally drop support for an old client (whether that is immediately or eventual), the min-client-version gives us a cut-off where we can safely know that old clients will just stop working until upgraded.

pkg/versioncheck/interceptor.go
jabr: This check is something that could probably go away eventually, but it is used for now for the case of a new client talking to a daemon that pre-dates this version checking system. That means the daemon did not enforce the min-daemon version requirement on the cli call, so here we simply warn the user in that check, because the daemon might have just done something they did not intend.

But once everyone has upgrade daemons to a version with this version-checking, then this check would never fail -- the daemon would always be rejecting immediately at the point.

pkg/versioncheck/interceptor.go
jabr: There might be a better way to write it, but this is what I had to do to get the headers out of the response. Not set them.

internal/machine/machine.go
jabr: This isn't necessary for the proxy refactor change, but it could be relevant for other kinds of backwards-compatibility breaking changes.

For example, if the semantics of a gRPC function itself changed, then it's the target daemon we are proxying to that needs to be checking versions as well.

The client has a min-daemon of "1.2" and connects to a machine running daemon "1.2" (so that check passes) but it wants the command proxy'd to another machine. And that machine is running daemon "1.1" still, so _it_ needs to check the version, too.

Otherwise the new "foobar" implementation in 1.2 that the client is expecting is going to actually get the old 1.1 response. 😄

pkg/versioncheck/interceptor.go
psviderski: Ah right, this is to extract headers from the response 👍

pkg/versioncheck/interceptor.go
```diff
@@ -0,0 +1,155 @@
...
+	"google.golang.org/grpc/status"
+)
+
+const (
+	MetadataKeyCLIVersion       = "x-uncloud-cli-version"
+	MetadataKeyMinDaemonVersion = "x-uncloud-min-daemon-version"
+	MetadataKeyDaemonVersion    = "x-uncloud-daemon-version"
+
+	MinCLIVersion    = "0.0.0"
+	MinDaemonVersion = "0.0.0"
```
psviderski, line 33: A comment for these would be helpful as it's not clear how they relate to each other and what it means when they have different versions.

pkg/versioncheck/interceptor.go
psviderski: It would be helpful to leave TODOs to clean up stuff that is only needed for the transition period after a couple more releases to keep this simple

internal/machine/machine.go
psviderski: >And that machine is running daemon "1.1" still, so it needs to check the version, too.

Still not following why this should be checked at the proxy, not at the server on the machine with 1.1 version.

>For example, if the semantics of a gRPC function itself changed, then it's the target daemon we are proxying to that needs to be checking versions as well.

I think I know what you're talking about but that should be a super rare case. Changing the semantics of an API call is shooting yourself in the foot and asking for trouble. A changed semantics should usually be implemented as a new method to not deal with any versioning, only the minimal version the method was added in.

psviderski:
I'm happy to merge this if we can remove the checks that are necessary only for the transition period after a few more releases. Adding TODOs for this and transferring explanations from here to the code comments will also be helpful, especially the version constants in `interceptor.go` and what each interceptor checks.

internal/machine/machine.go
jabr: > Still not following why this should be checked at the proxy, not at the server on the machine with 1.1 version.

Ah, I see. It's my understanding that the client connection (or remote server proxying to it) comes in at this proxy server (and then routes it to its local machine server and/or remote servers' proxy server entry point).

So we need to do the version check there -- at the proxy entry point and before we get to the local machine server -- in order to make use of it for changes to the proxy routing logic itself (like we need for the proxy refactor branch).

It's entirely possible I missed something here, though, and have the connection flow wrong.

Also, we might need to add it to the local machine server, as well, if it is possible something connects directly to it and not via its proxy server.

pkg/versioncheck/interceptor.go
jabr: Added some notes/TODOs on that method and code supporting it that can all be removed eventually.

pkg/versioncheck/interceptor.go
jabr: Nice catch. Updated.

pkg/versioncheck/interceptor.go
jabr: Updated header names.

jabr:
> I'm happy to merge this if we can remove the checks that are necessary only for the transition period after a few more releases. Adding TODOs for this and transferring explanations from here to the code comments will also be helpful, especially the version constants in interceptor.go and what each interceptor checks.

Sounds good. I still have a few more follow-ups to make and then re-run my manual tests (hopefully in the next day or two).

jabr:
Okay, I believe have made all of the updates we discussed. Let me know if you have any other questions, particularly about the server interceptors on the proxy endpoints.

PR #277, opened 2026-03-18
feat: error on unsupported compose features

### LINKED ISSUE #237: Warn on unsupported compose keys
I'm trying to deploy an app with requires `sysctls` and `cap_add` compose keys, but seems like uncloud right now does not support any of these

It would be nice to warn about such things on app deployment

miekg:
This implements a check for the unimplemented features of the uncloud support matrix and points to the matrix in the error.

Fixes: #237

### DIFF
--- a/pkg/client/compose/project.go
+++ b/pkg/client/compose/project.go
@@ -13,6 +13,7 @@ import (
 	"github.com/compose-spec/compose-go/v2/tree"
 	"github.com/compose-spec/compose-go/v2/types"
 	"github.com/psviderski/uncloud/pkg/api"
+	"github.com/psviderski/uncloud/pkg/client"
 )
 
 var registerComposeOverrides sync.Once
@@ -67,7 +68,7 @@ func LoadProject(ctx context.Context, paths []string, opts ...composecli.Project
 	}
 
 	if err = validateServicesFeatures(project); err != nil {
-		return nil, err
+		client.PrintWarning(err.Error())
 	}
 
 	// Process image templates in services to expand Go template expressions using git repo state.
--- a/pkg/client/compose/project.go
+++ b/pkg/client/compose/project.go
@@ -66,6 +66,10 @@ func LoadProject(ctx context.Context, paths []string, opts ...composecli.Project
 		return nil, err
 	}
 
+	if err = validateServicesFeatures(project); err != nil {
+		return nil, err
+	}
+
 	// Process image templates in services to expand Go template expressions using git repo state.
 	if project, err = ProcessImageTemplates(project); err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/client/compose/service.go
```diff
@@ -423,3 +423,52 @@ func validateServicesExtensions(project *types.Project) error {
 
 	return nil
 }
+
+// validateServicesFeatures checks the service for unsupported features and returns an error for this first one found.
+func validateServicesFeatures(project *types.Project) error {
+	err := func(service, feature string) error {
+		return fmt.Errorf("service: '%s': unsupported feature: '%s', see %s", service, feature, "https://uncloud.run/docs/compose-file-reference/support-matrix")
```
psviderski, line 430: What do you think about only printing a warning for the detected unsupported attributes (we have `tui.PrintWarning` helper)?

A service may still work correctly even some of them are ignored. But failing with an error will require the user to remove unsupported attributes from the compose file. This might not be desired if the same file is used (e.g. as a base) for both `docker compose` to run locally and uncloud to deploy remotely.

pkg/client/compose/service.go
```diff
@@ -423,3 +423,52 @@ func validateServicesExtensions(project *types.Project) error {
...
+		// we only allow the 'default' network, nothing else.
+		if x := service.Networks; x != nil {
+			if len(x) != 1 {
+				return err(service.Name, "networks")
+			}
+			config, ok := x["default"]
+			if !ok || config != nil {
+				return err(service.Name, "networks")
+			}
+		}
```
psviderski, line 470: We don't support a hundred more attributes from the [spec](https://github.com/compose-spec/compose-spec/blob/main/spec.md). It's good to start with at least some but just curious why you chose only these ones?

If you use any AI assistants, an exhaustive list could be easily built with their help.

miekg: this list comes from the matrix of support features

pkg/client/compose/service.go
miekg: happy to warn, i'll change it

pkg/client/compose/project.go
```diff
@@ -67,7 +68,7 @@ func LoadProject(ctx context.Context, paths []string, opts ...composecli.Project
 	}
 
 	if err = validateServicesFeatures(project); err != nil {
-		return nil, err
+		client.PrintWarning(err.Error())
```
psviderski, line 71: The branch is likely using the old base commit as this is now in the `internal/cli/tui` package. So the linter and tests fail

miekg: fucked up my brranch and didn't dare force push, see #288

PR #278, opened 2026-03-18
chore: modernize

miekg:
This runs:

```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix ./...
```

over the codebase, as this is using go 1.26, it can also use the new new() functionallity so AsPtr and boolPtr is are needed anymore.


psviderski (approved):
Thanks! 👍

PR #281, opened 2026-03-21
feat: uc ls show status

### LINKED ISSUE #13: Show service status when listing services
Add information about running containers or service health.

At the moment `uc ls`:
```
SERVICE ID                         NAME         MODE         REPLICAS
c392fcf3eddda03e238835efa1ff24b3   nginx-agpy   replicated   1
2bb115b387da8f7b879900813cd281c0   nginx-dxyy   replicated   1
2c0367f9a1a9650bb64753d168f31c88   nginx-hh8w   replicated   1
7c4734ecd6a1ff9b41d3df33c562a32c   nginx-f4bz   replicated   1
760dbd214535dc4fdb6b2616fa0a00e8   caddy        global       2
```

miekg:
Plumb through the status of the containers.

```
NAME         MODE         REPLICAS   IMAGE                   ENDPOINTS   STATUS
excalidraw   replicated   1          excalidraw/excalidraw               running
```

Fixes: #13

### DIFF
--- a/cmd/uncloud/service/ls.go
+++ b/cmd/uncloud/service/ls.go
@@ -58,10 +58,11 @@ func list(ctx context.Context, uncli *cli.CLI) error {
 			return fmt.Errorf("write header: %w", err)
 		}
 	}
-	if _, err = fmt.Fprintln(tw, "NAME\tMODE\tREPLICAS\tIMAGE\tENDPOINTS"); err != nil {
+	if _, err = fmt.Fprintln(tw, "NAME\tMODE\tREPLICAS\tIMAGE\tENDPOINTS\tSTATUS"); err != nil {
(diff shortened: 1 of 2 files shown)


psviderski (commented):
So the service status is a concatenation of all replica statuses. I think it won't work well if we have 10 or even more containers/replicas for a service.

What we likely want is some sort of a summary or computed status based on individual replica statuses. For example, `X/Y running` (X out of Y replicas are running) or `X running, Z restarting`.

Maybe also add red/yellow/green dot to highlight which services need attention.

miekg:
ah, I thought it might be too easy :) Feel free to close ,'cause I don't have time to also work on this.
(wait, I'll close)

Issue #158, opened 2025-10-27
CLI command to stream uncloud daemon logs from machines


psviderski:
As suggested in https://github.com/psviderski/uncloud/discussions/157, a CLI command for streaming the orchestrator logs (`uncloudd` daemon on machine(s)) will be helpful for troubleshooting.

For example, `uc machine logs` or `uc machine logs -m specific-machine`
Consider adding options for also streaming `uncloud-corrosion` or `docker` logs, e.g.
* `uc machine logs --corrosion`
* `uc machine logs --docker`
* `uc machine logs --uncloudd --corrosion`

Relates to https://github.com/psviderski/uncloud/issues/12.

miekg:
I've implemented something like that in https://github.com/miekg/pgo/blob/main/compose/compose.go#L44

think some part might be lifted from there, I'll take a look

psviderski:
Awesome! Please see the primitives we implemented for chronologically merging multiple container streams into a service logs stream (`uc logs`) in https://github.com/psviderski/uncloud/pull/196: https://github.com/psviderski/uncloud/blob/832c00d0712e6d4caa67306312a48f278ada8c9d/pkg/client/logmerger.go
Might be useful for this as well.

I briefly described the approach here: https://x.com/psviderski/status/1992952074930446745

miekg:
posting x here (as I want to avoid that site as much as possible)
~~~
 Today I’m solving a classic distributed systems problem: how to reliably stream logs from multiple replicas in chronological order?

Surprisingly, none of the mainstream container tools do this.

docker service logs, kubectl logs, stern, kubetail, k9s - they all just interleave streams as they arrive. The standard workaround is to pipe through | sort 🙃

To do it properly, you can't immediately print a log entry because a slower stream might have an earlier timestamp that hasn't arrived yet. You need to know when it's "safe" to print.

Here's how I'm solving this for the new 'uc logs' command. You need two things:

1. Low watermark

Track the minimum timestamp across all log streams and safely print buffered logs with timestamps below this watermark.

> watermark = min(latest_timestamp for each stream)
> sort and print all buffered logs where timestamp < watermark

2. Heartbeat timestamps

What if a container becomes quiet? The entire stream gets stuck because the watermark wouldn’t advance.

The solution is to periodically send empty heartbeat entries with the current timestamp from the servers. This is a promise: "I have no logs older than this"

This lets the client advance the watermark immediately instead of blocking the output. 500ms–1s interval works well in practice.

[...]

PR #282, opened 2026-03-22
Add server side of journal logs

miekg:
This add the server side and grpc methods to get a journal logs from a machine. It repurposes ServiceLogEntry for these logs to keep the changes somewhat to a minimum. And it lets us re-use the merging of the various logs. In the protobufs ContainerLog has been renamed to just Log and LogEntry, as these are now also used for journal logs.

It does api.LogOptions in more places to reduce the various logOpts that were used.

It does not yet plumb it through to the uc client, that needs a follow up pr. Following logs is also not yet implemented.

See #273 for linked discussions

### DIFF
--- a/internal/journal/journal.go
+++ b/internal/journal/journal.go
@@ -51,25 +51,19 @@ func logs(ctx context.Context, unit string, opts api.ServiceLogsOptions) (io.Rea
 	return p, nil
 }
 
-// follow synchronously follows the io.Reader, writing each new journal entry to writer. The
-// follow will continue until a single time.Time is received on the until channel (or it's closed).
+// follow synchronously follows the io.Reader, writing each new journal entry to channel.
 func follow(ctx context.Context, reader io.Reader, outCh chan api.LogEntry) error {
 	scanner := bufio.NewScanner(reader)
 
-	go func() {
-		for scanner.Scan() {
-			outCh <- entry(scanner.Bytes())
-		}
-		if err := scanner.Err(); err != nil {
-			outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
-		}
-		return
-	}()
-
-	for {
+	for scanner.Scan() {
 		select {
+		case outCh <- entry(scanner.Bytes()):
 		case <-ctx.Done():
 			return nil
 		}
 	}
+	if err := scanner.Err(); err != nil {
+		outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
+	}
+	return nil
 }
--- a/internal/journal/logs.go
+++ b/internal/journal/logs.go
@@ -13,6 +13,15 @@ import (
 
 // Logs streams logs from a service and returns entries via a channel.
 func Logs(ctx context.Context, unit string, opts api.ServiceLogsOptions) (<-chan api.LogEntry, error) {
+	// Hard code unit check for now
+	switch unit {
+	case "uncloud":
+	case "corrosion":
+	case "docker":
+	default:
+		return nil, fmt.Errorf("journal logs: invalid unit: %s", unit)
+	}
+
 	reader, err := logs(ctx, unit, opts)
 	if err != nil {
 		return nil, err
@@ -46,10 +55,6 @@ func Logs(ctx context.Context, unit string, opts api.ServiceLogsOptions) (<-chan
 			}
 		}()
 	}
-
-	go func() {
-		<-ctx.Done()
-	}()
 
 	return outCh, nil
 }
--- a/internal/journal/logs_test.go
+++ b/internal/journal/logs_test.go
@@ -22,7 +22,7 @@ func TestLogs(t *testing.T) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	ch, err := Logs(ctx, "notused", api.ServiceLogsOptions{})
+	ch, err := Logs(ctx, "uncloud", api.ServiceLogsOptions{})
 	require.NoError(t, err)
 
 	i := 0
@@ -30,23 +30,20 @@ func TestLogs(t *testing.T) {
 		i++
 	}
 	assert.Equal(t, i, 6)
-}
-
-func TestLogs_Tail(t *testing.T) {
-	t.Parallel()
+	cancel()
 
 	commandContext = func(ctx context.Context, _ string, _ ...string) *exec.Cmd {
 		return exec.CommandContext(ctx, "/usr/bin/tail", "-f", "testdata/logs")
 	}
 
-	ctx := context.Background()
-	ctx, cancel := context.WithCancel(ctx)
+	ctx = context.Background()
+	ctx, cancel = context.WithCancel(ctx)
 	go func() { time.Sleep(1 * time.Second); cancel() }()
 
-	ch, err := Logs(ctx, "notused", api.ServiceLogsOptions{})
+	ch, err = Logs(ctx, "uncloud", api.ServiceLogsOptions{Tail: 3})
 	require.NoError(t, err)
 
-	i := 0
+	i = 0
 	for range ch {
 		i++
 	}


miekg:
see #158 for that actual discussion.

internal/machine/api/pb/docker.proto
```diff
@@ -17,7 +16,7 @@ service Docker {
   rpc RemoveContainer(RemoveContainerRequest) returns (google.protobuf.Empty);
 
   rpc ExecContainer(stream ExecContainerRequest) returns (stream ExecContainerResponse);
-  rpc ContainerLogs(ContainerLogsRequest) returns (stream ContainerLogEntry);
+  rpc ContainerLogs(LogsRequest) returns (stream LogEntry);
```
psviderski, line 19: Am I right that this is a backward compatible change at the binary wire format level? As long as the field numbers stay the same, the message name doesn't matter.

internal/machine/machine.go
```diff
@@ -1073,3 +1077,93 @@ func (m *Machine) InspectService(
 	}
 	return &pb.InspectServiceResponse{Service: svc}, nil
 }
+
+// logsHeartbeatInterval is the interval at which heartbeat entries are sent when there are no logs to stream.
+const logsHeartbeatInterval = 200 * time.Millisecond
+
+// MachineLogs streams logs from a container.
```
psviderski:
```suggestion
// MachineLogs streams logs from a systemd service.
```

internal/machine/machine.go
```diff
@@ -1073,3 +1077,93 @@ func (m *Machine) InspectService(
...
+	ctx := stream.Context()
+
+	opts := api.ServiceLogsOptions{
+		Follow: req.Follow,
+		Tail:   int(req.Tail),
+		Since:  req.Since,
+		Until:  req.Until,
+	}
+
+	logsCh, err := journal.Logs(ctx, req.Id, opts)
```
psviderski, line 1098: Do we want to restrict the API to only a set of allowed unit names to not unnecessarily expose access to what is not needed for uncloud operation and troubleshooting?
Or is the idea to have a unified API that would allow to troubleshoot both containerised and any system services on the machine?

internal/machine/machine.go
```diff
@@ -1073,3 +1077,93 @@ func (m *Machine) InspectService(
...
+	if err != nil {
+		if errdefs.IsNotFound(err) {
+			return status.Error(codes.NotFound, err.Error())
+		}
+		return status.Errorf(codes.Internal, "get journal logs: %v", err)
+	}
+
+	log := slog.With("unit", req.Id, "stream_id", fmt.Sprintf("%p", stream)[2:])
+	log.Debug("Starting container logs streaming.",
```
psviderski:
```suggestion
	log.Debug("Starting systemd service logs streaming.",
```

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+	if opts.Tail > 0 {
+		args = append(args, "-n")
+		args = append(args, fmt.Sprintf("%d", opts.Tail))
+	}
+	if opts.Follow {
+		args = append(args, "-f")
+	}
+
+	args = append(args, "-o")
+	args = append(args, "short-iso")
```
psviderski: How about using `short-iso-precise` to get the milliseconds and make the format consistent with `uc logs`?

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if opts.Since != "" {
+		args = append(args, "-S")
+		args = append(args, opts.Since)
+	}
```
psviderski, line 35: We also have `opts.Until` that needs to be handled here

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+
+	cmd := exec.Command(journalctl, args...)
```
psviderski: Did you consider using `exec.CommandContext` and just passing the scoped context from `Logs`? The process killing will be automatically handled on the context cancellation.
Just curious why you chose handling cancellation manually.

internal/machine/machine.go
```diff
@@ -1073,3 +1077,93 @@ func (m *Machine) InspectService(
 	}
 	return &pb.InspectServiceResponse{Service: svc}, nil
 }
+
+// logsHeartbeatInterval is the interval at which heartbeat entries are sent when there are no logs to stream.
+const logsHeartbeatInterval = 200 * time.Millisecond
+
+// MachineLogs streams logs from a container.
+func (s *Machine) MachineLogs(
```
miekg, line 1085: this duplicates `ContainerLogs`, wasn't sure to clean that up now or later, hence just copied it

internal/journal/logs.go
```diff
@@ -0,0 +1,90 @@
...
+			if err := scanner.Err(); err != nil {
+				outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
+			}
+		}()
+
+	case true:
+		fw := &FollowWriter{outCh}
+
+		go func() {
+			defer close(outCh)
```
psviderski, line 50: Looks like we leak the `journalctl` process for the follow case. I think doing cleanup on the context cancelation will make the logic a bit simpler. Note also you can assume that the `stream.Context()` is automatically cancelled when the stream's RPC method returns.

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+// follow synchronously follows the io.Reader, writing each new journal entry to writer. The
+// follow will continue until a single time.Time is received on the until channel (or it's closed).
+func follow(ctx context.Context, reader io.Reader, writer io.Writer) error {
+	scanner := bufio.NewScanner(reader)
+	bufch := make(chan []byte)
+	errch := make(chan error)
+
+	go func() {
+		for scanner.Scan() {
+			if err := scanner.Err(); err != nil {
```
psviderski: I think `scanner.Err()` is checked at wrong point as it will always return `nil` if `Scan` returned `true`.

internal/journal/logs.go
```diff
@@ -0,0 +1,90 @@
...
+			err := follow(ctx, reader, fw)
+			if err != nil {
+				outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
+			}
+		}()
+	}
+
+	go func() {
+		<-ctx.Done()
+		reader.Close()
```
psviderski: Not sure if calling `reader.Close()` concurrently with the scanner reading from it is a good idea. Maybe all the cleanup should be made in one place.
Perhaps we don't need to close it manually as `Wait()` closes it according to the docs.

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/psviderski/uncloud/pkg/api"
+)
+
+const journalctl = "journalctl"
+
+func logs(unit string, opts api.ServiceLogsOptions) (io.ReadCloser, func() error, error) {
+	cancel := func() error { return nil }
+
+	args := []string{"-u", unit, "--no-hostname"}
+	if opts.Tail > 0 {
```
psviderski: Do we want to support `all` logs (tail is `-1`)? Also, why not support `0`?

internal/journal/journal.go
```diff
@@ -0,0 +1,91 @@
...
+		}()
+		return cmd.Process.Kill()
+	}
+
+	return p, cancel, nil
+}
+
+// follow synchronously follows the io.Reader, writing each new journal entry to writer. The
+// follow will continue until a single time.Time is received on the until channel (or it's closed).
+func follow(ctx context.Context, reader io.Reader, writer io.Writer) error {
```
psviderski: `FollowWriter` feels like an unnecessary indirection here. `follow` already splits lines with a scanner then writes them to an `io.Writer` only for `FollowWriter.Write()` to parse them back into `LogEntry` values and send them on the channel. Could `follow` just send entries directly on the channel instead? That would remove the `io.Writer` detour and simplify the flow.

Also `follow` calls `io.WriteString(writer, "\n")` separately. Since each `Write` call triggers parsing the data into a `LogEntry`, I expect `\n` gets parsed as its own entry that results in a log entry with a zero timestamp and empty message.

miekg:
> Overall looks good but the `journalctl` subprocess handling needs a bit more work. Please see the comments.

Yeah, thanks for those - once I'll write the tests this will probably show 'not working' and require some fixes.

internal/machine/api/pb/docker.proto
miekg: ooohh... I don't know actually...
(should be testable, at least in a one-off)

psviderski:
JFYI: I updated the doc with a couple new commands I use for manual testing of daemon changes on my dev cluster: https://github.com/psviderski/uncloud/blob/main/HACKING.md#manual-testing-on-dev-machines

Sometimes it's too cumbersome to properly test certain functionality using unit or e2e tests. Especially for new experimental features when it's unclear whether they will stick. In those case I make sure I at least thoroughly test them manually on my dev machines.

internal/journal/journal.go
miekg: done

internal/journal/journal.go
miekg: done

internal/journal/journal.go
miekg: done

internal/journal/journal.go
miekg: Much better idea. Why this is done this way: when that original code was written CommandContext didn't exist yet.

I'll take a closer look at this whole cancelation and starting of the process

internal/journal/journal.go
miekg: good points. Done

internal/journal/journal.go
miekg: moved

internal/journal/logs.go
miekg: using CommandContext. This simplifies it more.  Think btw that `follow` is not needed anymore as this is now almost identical the the non-follow code.

internal/journal/logs.go
miekg: removed for now

internal/machine/api/pb/docker.proto
miekg: it should be backwards compat, the (type) names are not used on the wire.

internal/machine/machine.go
miekg: applied manually

internal/machine/machine.go
miekg: the first. This PR already clashes it bit with multitenant (you need cluster admin right or some such). But exposing more logging is a risk.

I vagely remember docker compose being not good at hiding things which could lead to a compromise; hence I would restrict it to 3 units: incloud(d), docker and corrosion

internal/machine/machine.go
miekg: done manually

miekg:
thanks for your feedback; if mostly just applied it - journal/journal.go and logs.go looks way more sane now; i suspect there is still work there, process cancelation is somewhat tricky; but I need to actually run it to see how it fares :)
(I wanted to do tests in the 3rd PR, but prolly can't hurt to add some now - esp unit test wtr to cancelation, but these are somewhat shit, as a normal use can't see the logs and thus run journalct)

And add unit test in journal/

miekg:
> JFYI: I updated the doc with a couple new commands I use for manual testing of daemon changes on my dev cluster: https://github.com/psviderski/uncloud/blob/main/HACKING.md#manual-testing-on-dev-machines
> 
> Sometimes it's too cumbersome to properly test certain functionality using unit or e2e tests. Especially for new experimental features when it's unclear whether they will stick. In those case I make sure I at least thoroughly test them manually on my dev machines.

Ack, I'll try, but my dev machines are mostly arm64, and e.g.  ucind also didn't work (on amd64 it did btw)

miekg:
i dont understand the test failure, but this is ready for (another) review

psviderski:
Please see the quick Claude's review below which seems to have picked up something useful. Ignore what doesn't make sense.

## Addressed comments

1. **`short-iso-precise` format** — Done. `journal.go:30-31` now uses `-o short-iso-precise`.
2. **`opts.Until` handling** — Done. `journal.go:38-41` now passes `-U` when `opts.Until` is set.
3. **`exec.CommandContext`** — Done. `journal.go:47` uses `commandContext(ctx, ...)` which defaults to `exec.CommandContext`. Process cleanup on context cancellation is handled automatically.
4. **`FollowWriter` indirection removed** — Done. `follow()` now sends entries directly on the channel via `outCh <- entry(scanner.Bytes())`.
5. **`scanner.Err()` placement** — Partially addressed, see new issues below.
6. **Comment suggestion `// MachineLogs streams logs from a systemd service.`** — Done. `machine.go:1089`.
7. **Log message suggestion** — Done. `machine.go` uses `"Starting systemd service logs streaming."`.
8. **Tail `-1` / `0` support** — Done. `journal.go:26` checks `opts.Tail > -1` and passes `"all"` otherwise.
9. **Process leak / `reader.Close()` concurrency** — Using `CommandContext` now handles process cleanup. Manual close removed.

## New issues

### Bug: Duplicate `scanner.Err()` check in `journal.go:69-73`

```go
if err := scanner.Err(); err != nil {
    if err := scanner.Err(); err != nil {  // duplicate!
        outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
    }
    return
}
```

The `scanner.Err()` check is nested inside itself. The inner check is redundant. Should be:

```go
if err := scanner.Err(); err != nil {
    outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
}
```

### Bug: `follow()` blocks forever if scanner goroutine finishes

In `follow()`, the scanner goroutine closes when scanning ends, but the main goroutine only exits on `ctx.Done()`. If `journalctl` terminates (e.g. error or `--until` reached while following), the goroutine finishes and closes nothing, but `follow()` blocks forever waiting on `ctx.Done()`. The channel `outCh` is never closed by `follow()` itself, so the caller in `Logs()` that reads from `outCh` will also hang.

The fix would be to signal from the goroutine back to the main flow, or restructure so the goroutine itself is the whole function:

```go
func follow(ctx context.Context, reader io.Reader, outCh chan api.LogEntry) error {
    scanner := bufio.NewScanner(reader)
    for scanner.Scan() {
        select {
        case outCh <- entry(scanner.Bytes()):
        case <-ctx.Done():
            return nil
        }
    }
    if err := scanner.Err(); err != nil {
        outCh <- api.LogEntry{Err: fmt.Errorf("journal logs: %w", err)}
    }
    return nil
}
```

This eliminates the need for the separate goroutine entirely.

### Nit: Stale comment in `follow()`

The comment says "writing each new journal entry to writer" but there is no writer anymore (it was removed per previous feedback). Should say "to the outCh channel" or similar.

### Nit: Dead goroutine in `Logs()` at line 139-141

```go
go func() {
    <-ctx.Done()
}()
```

This goroutine does nothing useful. It looks like a leftover from the previous cleanup approach. Should be removed.

### Nit: `TestLogs_Tail` does not actually test the follow path

The `commandContext` mock uses `tail -f` to simulate following, but `Logs()` is called with `api.ServiceLogsOptions{}` (where `Follow` is `false`). So `follow()` is never actually called by this test. If the intent is to test the follow path, `opts.Follow` should be `true`.

Also the comment says "still six is hardbeats" (typo: "heartbeats").

### Nit: `commandContext` package-level var is not test-safe

Both `TestLogs` and `TestLogs_Tail` set the package-level `commandContext` var and run with `t.Parallel()`. This works by coincidence today but is fragile. Consider using a struct with the dependency injected, or at minimum note the limitation.

### Not yet addressed: Restricting allowed unit names

The previous review discussed restricting the API to only allowed unit names (uncloud, docker, corrosion). miekg agreed to restrict it, but there is no validation in `MachineLogs()`. It passes `req.Id` directly to `journal.Logs()`. This should be added before merge if that was the agreed direction.

psviderski:
>i dont understand the test failure, but this is ready for (another) review

These are flaky tests that sometimes fail on CI not being able to initialise a ucind cluster. I suspect they might be caused by system resource contention on GitHub Action machines. We probably need to reduce the number of clusters we create in parallel. Maybe adding more retries and increasing timeouts for the cluster creation operations. It's hard to reproduce this and properly fix.
We discussed some ideas about using a pool of clusters to reuse as much as possible here: https://github.com/psviderski/uncloud/pull/165#discussion_r2570333664

miekg:
well, those were useful comments. ptal

miekg:
thank you! I'll work on #283 to add "the rest"

Issue #284, opened 2026-03-24
[BUG] ucind runs x86 corrosion on arm64


miekg:
**Describe the bug**

ucind download the x64 docker image/tar file of corrosion instead of the arch64 one (when on arch). 
```
% docker cp 3bbbdede22b7:/images/corrosion.tar /tmp
# untar the tar and the tars in there
% cd usr/local/bin
% file corrosion
corrosion: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=baf9259b35532370e625a0055366df4563649e66, not stripped
```

**How to reproduce**

`% ./ucind cluster create miektest` on a arch64 machine

This will not come up because corrosion is not running: `DEBUG Retrying corrosion API request due to network error. error="dial tcp 127.0.0.1:51002: connect: connection refused"`

**Expected behavior**

An aarch64 version should be downloaded. Funny is that the Dockerfile has a lot of step to make that distinction, but some how it fails

**Environment:**

Asahi Linux/Fedora 43

- Uncloud versions: *main*

PR #285, opened 2026-03-24
Dockerfile: Get the correct corrosion version

miekg:
On arm64 the x86 version of corrosion was downloaded. This downloads the correct version. Also added comments to HACKING.md on how to clean up a broken cluster as I have done it twice now and forgot how to do it both times.

Fixes: #284

~~~ shell
% ./ucind cluster create miektest
Cluster "miektest" initialised with machine "machine-1" Waiting for cluster to be ready... done.
Cluster 'miektest' added to Uncloud config as the current context. Cluster 'miektest' created.
~~~

`uc machine ls` also works.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -47,6 +47,16 @@ Windows:
 GOOS=linux GOARCH=amd64 go build -o uncloudd ./cmd/uncloudd
 ```
 
+## Docker in Docker (ucind)
+
+When running ucind and running in trouble, you must manually remove the cluster, by running:
+
+```shell
+docker stop CONTAINERID
+docker rm CONTAINERID
+docker network rm CLUSTERNAME
+```
+
 ## Testing
 
 Run all tests (unit and e2e):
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -47,6 +47,16 @@ Windows:
...
 ```
 
+## Docker in Docker (ucind)
+
+When running ucind and running in trouble, you must manually remove the cluster, by running:
+
+```shell
+docker stop CONTAINERID
+docker rm CONTAINERID
+docker network rm CLUSTERNAME
```
psviderski, line 57: Sometimes when e2e tests crash, they may leave a dozen of running ucind containers. I typically use a command that removes all ucind-managed containers:
```
docker ps --filter "label=ucind.managed" -q | xargs docker rm -f
```

Just added a new `ucind:cleanup` command as a shortcut: https://github.com/psviderski/uncloud/commit/a0dbe4a983f3c37dd22b26c45ea2c1265ec63117

You can run with:
```
mise run ucind:cleanup
# or just (tab completion also works)
mise ucind:cleanup
```

I will update this section after merging.

psviderski (approved):
Nice fix, thank you!

HACKING.md
psviderski: Btw, did you try `ucind cluster rm` to remove the cluster containers that failed to start correctly?

miekg: yes I did but _only_ after I removed the container, then it complains that it can't connect, but also can't start a new one; the latter fix is to remove the network

PR #291, opened 2026-03-28
⚡ Performance: Unbuffered channel blocks Serf event loop

tang-vu:
## Problem

`nextCh` is an unbuffered channel. When `HandleEvent` receives a 'heads' event, it blocks on `b.nextCh <- userEvent.Payload` until the CRDT engine calls `Next()`. Because `HandleEvent` is typically called directly from Serf's event delivery loop, this synchronous handoff will block all Serf event processing (including node joins, leaves, and gossip) whenever the CRDT engine is busy and not actively waiting to read the next event.

**Severity**: `high`
**File**: `experiment/broadcaster.go`

## Solution

Initialize `nextCh` with a reasonable buffer to decouple the Serf event loop from the CRDT processing loop.

## Changes

- `experiment/broadcaster.go` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/experiment/broadcaster.go
+++ b/experiment/broadcaster.go
@@ -23,7 +23,7 @@ func NewSerfBroadcaster(ctx context.Context, serf *serf.Serf) *SerfBroadcaster {
 	return &SerfBroadcaster{
 		ctx:    ctx,
 		serf:   serf,
-		nextCh: make(chan []byte),
+		nextCh: make(chan []byte, 16),
 	}
 }
 


miekg:
looks odd, but anyway
 
how has the value of "16" been determined?

tang-vu:
Fair question. The value 16 was chosen as a reasonable heuristic - large enough to absorb bursts of `heads` events without blocking the Serf event loop, but small enough to avoid unbounded memory growth. In practice, CRDT head broadcasts are typically infrequent and small, so even a buffer of 4-8 would likely suffice.

If you have a preferred buffer size or a project convention for channel buffers, happy to adjust. The key insight is that *any* buffer > 0 prevents `HandleEvent` from blocking the entire Serf event delivery when the CRDT engine is busy.

miekg:
this sounds as a pretty weak argument.

you would have a stronger case if you have a short test demonstrating the issue

psviderski:
wtf

PR #296, opened 2026-03-31
feat: check autom if we can connect via Unix socket

### LINKED ISSUE #148: Install `uc` CLI on each cluster machine and allow running uc command agains local cluster

miekg:
When there is no config, but a unix socket does exist, connect via the unix socket. Prohibit saving the config if this is the case.

Fixes: #148

Use the new CutPrefix to shorten some code.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -54,10 +54,6 @@ func New(configPath string, conn *config.MachineConnection, contextName string) 
 }
 
 func (cli *CLI) CreateContext(name string) error {
-	if cli.conn.Unix != "" {
-		return fmt.Errorf("local unix socket connection '%s', not creating Uncloud config", cli.conn.Unix)
-	}
-
 	if _, ok := cli.Config.Contexts[name]; ok {
 		return fmt.Errorf("context '%s' already exists", name)
 	}
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -54,6 +54,10 @@ func New(configPath string, conn *config.MachineConnection, contextName string) 
 }
 
 func (cli *CLI) CreateContext(name string) error {
+	if cli.conn.Unix != "" {
+		return fmt.Errorf("local unix socket connection '%s', not creating Uncloud config", cli.conn.Unix)
+	}
+
 	if _, ok := cli.Config.Contexts[name]; ok {
 		return fmt.Errorf("context '%s' already exists", name)
 	}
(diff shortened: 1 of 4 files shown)


miekg:
note `fs.Exist` can be re-used in a few other places (grep for `os.Stat`)

internal/cli/cli.go
```diff
@@ -54,6 +54,10 @@ func New(configPath string, conn *config.MachineConnection, contextName string)
 }
 
 func (cli *CLI) CreateContext(name string) error {
+	if cli.conn.Unix != "" {
```
psviderski: We won't get that far with cluster initialisation when a `conn` is used instead of `Config` so this will never be reachable.

See this check earlier: https://github.com/psviderski/uncloud/blob/cb77d42107aa4583a5259a719bef5334b4d1cc29/cmd/uncloud/machine/init.go#L152-L158

Maybe we should instead check there that this is a connection to the local cluster and return a friendlier error message similar to this one explaining what's happening and why they should not try to initialise a new cluster? This error would be confusing as the user didn't pass the `--connect` flag.

Hm, but it's hard to distinguish whether they explicitly passed `--connect unix://...`. Maybe just skip this detail and only mention the connection to the cluster through the local machine. Feel free to suggest better alternatives.

psviderski (commented):
Looks good, just one little comment to address about a clear user error when connecting through the local machine.

internal/cli/cli.go
miekg: Ack, so remove the if- statement from CreateContext (for now?) and then in a future follow-up change the above error?

psviderski: Yeah, let's remove this if statement and add another condition to the error above for the new local case. Would be preferred in this PR to not come back to this in the future.

miekg: ok done. You may have even better ideas for the wording that I added

psviderski: Awesome, thanks!

PR #303, opened 2026-04-05
feat: add "StrictHostKeyChecking accept-new" to ssh

### LINKED ISSUE #297: ssh host verification
```
% ./uc machine init --no-install --no-dns --no-caddy --context uc1 test@uncloud1.vm.science.ru.nl
Error: connect to remote machine: connect to machine: SSH connection to 'test@uncloud1.vm.science.ru.nl': exit status 255: Host key verification failed.
```

correct, because I haven't logged in yet, and did the accept host dance. But I think we may not care and set ` -o "StrictHostKeyChecking no"` when using ssh.

`go+ssh` does already not care about this.

miekg:
Disable host key checking to prevent interaction with ssh.

Fixes: #297

### DIFF
--- a/pkg/client/connector/sshcli.go
+++ b/pkg/client/connector/sshcli.go
@@ -131,6 +131,8 @@ func (c *SSHCLIConnector) buildSSHArgs() []string {
 	// Disable interactive prompts (e.g., passphrase input) to prevent interference with the TUI.
 	// Authentication must succeed non-interactively via SSH agent or unencrypted key.
 	args = append(args, "-o", "BatchMode=yes")
+	// Disable host key checking for parity with go+ssh.
+	args = append(args, "-o", "StrictHostKeyChecking=no")
 	// Disable pseudo-terminal allocation to prevent SSH from executing as a login shell.
 	args = append(args, "-T")
 
--- a/pkg/client/connector/sshcli.go
+++ b/pkg/client/connector/sshcli.go
@@ -132,7 +132,7 @@ func (c *SSHCLIConnector) buildSSHArgs() []string {
 	// Authentication must succeed non-interactively via SSH agent or unencrypted key.
 	args = append(args, "-o", "BatchMode=yes")
 	// Disable host key checking for parity with go+ssh.
-	args = append(args, "-o", "StrictHostKeyChecking=no")
+	args = append(args, "-o", "StrictHostKeyChecking=accept-new")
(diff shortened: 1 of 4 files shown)


pkg/client/connector/sshcli.go
```diff
@@ -122,6 +122,8 @@ func (c *SSHCLIConnector) buildSSHArgs() []string {
 	// Disable interactive prompts (e.g., passphrase input) to prevent interference with the TUI.
 	// Authentication must succeed non-interactively via SSH agent or unencrypted key.
 	args = append(args, "-o", "BatchMode=yes")
+	// Disable host key checking for parity with go+ssh.
+	args = append(args, "-o", "StrictHostKeyChecking=no")
```
psviderski: How about using `StrictHostKeyChecking=accept-new` - accept unknown keys automatically but reject if a known key changes. Which is I believe is how most people actually want to use SSH interactively.
It's much safer than `no` because it catches key changes for known hosts that may signal a MITM or something that may need attention.

The same behaviour requires a bit more work for ssh+go connector so we can leave `InsecureIgnoreHostKey` there as we don't expect it to be widely used.

psviderski (commented):
Just one comment on using `StrictHostKeyChecking=accept-new` as it seems to provide more secure behaviour for free.

Please also update the failed unit tests accordingly: https://github.com/psviderski/uncloud/actions/runs/23996094206/job/69984307653?pr=303#step:7:566

pkg/client/connector/sshcli.go
miekg: oh, didn't know that existed. Much better idea. Yes.

pkg/client/connector/sshcli.go
```diff
@@ -131,6 +131,8 @@ func (c *SSHCLIConnector) buildSSHArgs() []string {
 	// Disable interactive prompts (e.g., passphrase input) to prevent interference with the TUI.
 	// Authentication must succeed non-interactively via SSH agent or unencrypted key.
 	args = append(args, "-o", "BatchMode=yes")
+	// Disable host key checking for parity with go+ssh.
+	args = append(args, "-o", "StrictHostKeyChecking=accept-new")
```
psviderski, line 135: There are failing tests for this as well

miekg: ah, sorry missed those. Brushed it off is ci test failing again

Issue #168, opened 2025-11-05
Cluster-aware image pulls (pull images from other machines)


psviderski:
With `uc image push` command ([v0.13.0](https://github.com/psviderski/uncloud/releases/tag/v0.13.0)), you can push images directly to machines in your cluster (all or only the specified ones). Now I’m considering whether to make image pulls cluster-aware too.

**Current behavior:** When deploying a service, if the image is missing on the target machine, it pulls from Docker Hub/external registry (standard Docker behaviour). Or fails if you don't use a registry and image is missing.

**Potential improvement:** Before pulling from external registries, check if the image exists on another machine in the cluster and pull from there instead. Basically treating the cluster as a **distributed image cache** and `uc image push` as a way to push an image to it.

It sort of blurs the boundaries between machines even further. You **push the image to the cluster** (you don’t need to worry about specifying which machines) and then **run a service in the cluster** (the image will be pulled to the required machines if missing). Somewhat similar to [Spegel](https://spegel.dev/) but in our case much simpler.

**Implementation considerations:** Every machine runs an unregistry API on `10.210.X.1:5000` which can be easily accessed from any other machine via the private cluster network. The Docker gRPC service can be extended with `PullImageFromMachine` call that uses existing `PullImage` to pull from another machine's registry. It's likely some tricks with prefixing the image name with the registry address and then retagging will be required.

Note also that people on Discord liked the proposal but considered it be more relevant for larger setups and suggested to focus on other more critical features first: https://discord.com/channels/1371726032104587335/1371726032838594565/1427954990860865608

Issue #75, opened 2025-06-20
Add support for Compose secrets from files


psviderski:
Similarly to configs support https://github.com/psviderski/uncloud/issues/43, add basic support for Compose secrets from files: https://docs.docker.com/reference/compose-file/secrets/

Note also that we need to support [`secrets`](https://github.com/compose-spec/compose-spec/blob/main/build.md#secrets) in the `build` section as well. See relevant issue: https://github.com/psviderski/uncloud/issues/181

jabr:
I was looking at implementing this, but it's not entirely clear to me what `secrets` do over `configs` in practice. The only things I see are:
1. defaults to `/run/secrets/{name}` for default path instead of `/{name}`
2. defaults to `0o400` permissions instead of `0o444`

And we could also maybe store them encrypted-at-rest in corrosion, but don't think the key could really be much more protected than the corrosion db (would need to be readable by uncloud user's processes and shared amongst all machines...) so I'm not sure that would add much.

Still worthwhile to do, even just for compatibility/convenience/semantics alone, but I'm thinking the implementation of it could effectively just be a flag to the existing configs implementation? A lot of the code would be the same between the two.

Naive LLM spike for reference: https://github.com/psviderski/uncloud/compare/main...jabr:uncloud:compose-secrets

psviderski:
Thank you so much for taking a look! I apologise we haven't updated the details of this issue as we've discussed it with @tonyo on the call after he implemented configs. I'm not sure if we have any notes from that discussion. From the top of my head, yeah we probably want to encrypt them before syncing to the Corrosion store. This will help us to worry less about protecting Corrosion, see https://github.com/psviderski/uncloud/issues/110. Also, we might want to support corrosion DB backups at some point so having secrets in plain text in a backup is obviously not great.

Or perhaps we can completely strip secrets out when syncing to corrosion like we do with env vars at the moment: https://github.com/psviderski/uncloud/blob/b046b78398562f40ec2125aeeb6543f405052783/internal/machine/store/container.go#L52-L55. Actually, this is another SQLite db (`machine.db`) we use for extending Docker concepts. For example, to store a `ServiceSpec` for each Docker container we create so that we can scale services without requiring a Compose file or comparing the old and new specs when deploying. Every node has its own `machine.db` that is not shared with other machines.

[...]

PR #306, opened 2026-04-08
feat: Add x-registry to compose files for private docker registries

miekg:
This allows setting credentials for private docker registries. The syntax is copied from Rancher.

Documentation is updated, there is no unit test, it was tested manually and works with the local private registry (GitLab).

Functionality for retrieving the auth tokens locally is removed, also because I had a hard time getting this to work, and seems awkward to a log into a remote machine to do a 'docker login'.

Fixes: #299

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -21,8 +21,6 @@ import (
 
 	"github.com/containerd/errdefs"
 	"github.com/distribution/reference"
-	dockercommand "github.com/docker/cli/cli/command"
-	dockerconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/filters"
@@ -276,14 +274,6 @@ func (s *Server) PullImage(req *pb.PullImageRequest, stream grpc.ServerStreaming
 	if len(req.Options) > 0 {
 		if err := json.Unmarshal(req.Options, &opts); err != nil {
 			return status.Errorf(codes.InvalidArgument, "unmarshal options: %v", err)
-		}
-	}
-
-	if opts.RegistryAuth == "" {
-		// Try to retrieve the authentication token for the image from the default local Docker config file.
-		dockerConfig := dockerconfig.LoadDefaultConfigFile(os.Stderr)
-		if encodedAuth, err := dockercommand.RetrieveAuthTokenFromImage(dockerConfig, req.Image); err == nil {
-			opts.RegistryAuth = encodedAuth
 		}
 	}
 
--- a/pkg/api/service.go
+++ b/pkg/api/service.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"maps"
@@ -11,6 +12,7 @@ import (
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/distribution/reference"
+	"github.com/docker/docker/api/types/registry"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/psviderski/uncloud/internal/machine/api/pb"
@@ -59,6 +61,8 @@ type ServiceSpec struct {
 	Configs []ConfigSpec
 	// Container defines the desired state of each container in the service.
 	Container ContainerSpec
+	// Registry defines credentials for connecting to (private) registries.
+	Registry RegistrySpec
 	// Mode is the replication mode of the service. Default is ServiceModeReplicated if empty.
 	Mode string
 	Name string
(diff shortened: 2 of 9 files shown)


miekg:
`uc run` would need to get --username and --password (and --password-stdin) for feature parity - just like docker - if that's desired.

internal/machine/docker/server.go
```diff
@@ -279,14 +277,6 @@ func (s *Server) PullImage(req *pb.PullImageRequest, stream grpc.ServerStreaming
...
 	}
 
-	if opts.RegistryAuth == "" {
-		// Try to retrieve the authentication token for the image from the default local Docker config file.
-		dockerConfig := dockerconfig.LoadDefaultConfigFile(os.Stderr)
-		if encodedAuth, err := dockercommand.RetrieveAuthTokenFromImage(dockerConfig, req.Image); err == nil {
-			opts.RegistryAuth = encodedAuth
-		}
-	}
-
```
psviderski, line 289: What are the downsides of leaving this as the default if the registry credentials are not explicitly provided?

I really like how this works for me both locally and on CI so I don't need to deal with any secrets in the Compose file. And local docker provides different kinds of credential stores, allowing to use also custom credential helpers to integrate with all kinds of secret managers.

This is a significant downgrade for me.

I wonder, did `uc deploy` have issues with pulling when you could `docker pull` the same image on the same machine where you run `uc deploy`? That's what I believe this covers.

pkg/client/compose/project.go
```diff
@@ -40,6 +40,7 @@ func LoadProject(ctx context.Context, paths []string, opts ...composecli.Project
 		composecli.WithExtension(MachinesExtensionKey, MachinesSource{}),
 		composecli.WithExtension(PortsExtensionKey, PortsSource{}),
 		composecli.WithExtension(PreDeployHookExtensionKey, PreDeployHook{}),
+		composecli.WithExtension(RegistryExtensionKey, RegistrySource{}),
```
psviderski, line 43: It feels like it should be possible to define the registry credentials once and use for all services in the project/compose file. Like in Kamal and Rancher 1.x it's a top-level attribute, not per-service.

pkg/api/service.go
```diff
@@ -59,6 +61,8 @@ type ServiceSpec struct {
 	Configs []ConfigSpec
 	// Container defines the desired state of each container in the service.
 	Container ContainerSpec
+	// Registry defines credentials for connecting to (private) registries.
+	Registry RegistrySpec
```
psviderski, line 65: If we put registry credentials in the spec, the same concerns as for secrets start to apply to them: https://github.com/psviderski/uncloud/issues/75#issuecomment-3587798391 and we don't have a good story for that yet.

Is this absolutely necessary at this stage? The client still invokes every image pull operation so we can just get the credentials from either the local docker or the compose file and pass them to the the pull operation without sharing them anywhere in the cluster store in plaintext.

The only issue I can think of is that `uc scale` may fail if it's unable to retrieve the creds locally (same as it works now). But for this case, we have a better story IMO - just pull the image from the other cluster machine instead of the external registry which should be even faster: https://github.com/psviderski/uncloud/issues/168

psviderski (commented):
Overall it's a great idea to allow specifying the registry credentials in the Compose file but I have a few concerns with the approach. Please see the comments

internal/machine/docker/server.go
miekg: I had trouble getting this work (I'm fine with leaving it in)... But how do you use it in practice?

You `docker login` locally *and* on the server before the container is pulled?

When not doing the above, `uc deploy` would fail to get the container.

pkg/client/compose/project.go
miekg: good point, will do

pkg/api/service.go
miekg: > Is this absolutely necessary at this stage? 

so you're saying a local docker login and then doing a `uc deploy` should be able to fetch from a private registry? I'll try again, but I failed to get this to work. 

Doing `docker login` remotely requires root-access - which if you can run docker privilged containers is easy to get - but I was planning to not give are using root access to the machines where uncloud runs.

miekg:
My bad. I'm *sure* I tried this, but indeed doing a *local* docker loging makes things work on the remote machine as well.

Works-for-me, and doesn't need this PR; it should be documented somewhere. Was somewhat counter-intuitive to me.

Happy to close this as we discuss secret handling first?

internal/machine/docker/server.go
psviderski: 1. Run `docker login` locally, no need to run it on the remote server
2. (Optional) test `docker pull IMAGE` locally works
3. `uc run IMAGE` or `uc deploy` for `compose.yaml` just works:
   ```
   services:
     app:
       image: IMAGE
   ```

See docs: https://uncloud.run/docs/guides/deployments/deploy-app#pull-from-a-private-registry

Does `docker pull` locally work for you but `uc deploy` doesn't? If so, then that's a bug we need to fix then.

psviderski: Same on CI, there are plenty of integrations to login against a registry like this GHA: https://github.com/docker/login-action
Then `uc deply` just picks it up automatically so no need to carry any secrets in the Compose file.

psviderski:
Ah just saw your another comment that `docker login` works for you now. It's documented here: https://uncloud.run/docs/guides/deployments/deploy-app#pull-from-a-private-registry
Maybe it's worth moving this to a separate short doc to help find it in the table of content.

You can close this PR if you don't need this. But I think we should support a top-level `registries` attribute because not all users may run Docker locally. And I haven't tested if this works with all kinds of distributions such as Rancher Desktop, Orbstack, Colima, etc. Could probably have issues with podman.

The current behaviour is the convenience for when you run Docker locally which is also currently the requirement if you build images locally.

miekg:
i'll close here, we can dig up this code whenever we need to; it's manually tested, it works, but it may not support all the fancy stuff `docker login` can do (i.e. credential-helpers)

Issue #142, opened 2025-10-13
CLI command to upgrade uncloud daemon on machines


psviderski:
Add `uc machine upgrade` or similar command to upgrade the `uncloudd` daemon to the specified or latest version on remote machines.

Will need to bundle this with reporting the daemon version in `uc machine ls` so that it's easier to see which machines needs upgrade.

jabr:
A `uc` command for this would be handy, but in case it is useful for anyone else in the meantime, I have a [Justfile](https://just.systems/man/en/) with this action:
```
upgrade-uncloudd version:
  https -o uncloudd.tar.gz -d https://github.com/psviderski/uncloud/releases/download/v{{version}}/uncloudd_linux_amd64.tar.gz
  tar -xf uncloudd.tar.gz
  sudo install uncloudd /usr/local/bin/uncloudd
  rm uncloudd uncloudd.tar.gz
  sudo systemctl restart uncloud
```

And can be run via ssh with:
```
ssh machine1 -t just upgrade-uncloudd 0.13.0
```

(That task uses [httpie](https://httpie.io/docs/cli) but can be replaced with the equivalent wget or curl command.)

psviderski:
Thank you Justin for sharing your Just recipe! That's exactly what the integrated command should do

pilasguru:
My script to update `uncloudd`;  It is easy to provide version from command line.  (I push this script to each node from Rundeck's job):

```
#!/bin/bash

VERSION="0.14.0"

ARCH=$(uname -m)

if [[ "$ARCH" == "x86_64" ]]; then
    URL="https://github.com/psviderski/uncloud/releases/download/v${VERSION}/uncloudd_linux_amd64.tar.gz"
elif [[ "$ARCH" == "aarch64" || "$ARCH" == "arm64" ]]; then
    URL="https://github.com/psviderski/uncloud/releases/download/v${VERSION}/uncloudd_linux_arm64.tar.gz"
else
    echo "Platform unsupported: $ARCH"
    exit 1
fi

echo "Updating Uncloudd to versión $VERSION to $ARCH"

curl -fsSL -o uncloudd.tar.gz "$URL" || {
    echo "❌ Error downloading $URL"
    exit 1
}

tar -xf uncloudd.tar.gz
install uncloudd /usr/local/bin/uncloudd
rm uncloudd uncloudd.tar.gz

systemctl restart uncloud

echo "✔ Daemon updated to v$VERSION"
```

miekg:
[...]

PR #308, opened 2026-04-09
ci: Build and publish latest (nightly) binaries

tonyo:
Publish latest builds to a rolling "Nightly" (pre-)release on GitHub, so that they can be installed and tested. 

Naming: we can do "nightly"; "latest" is already taken by GitHub to represent the latest release.


psviderski:
Goreleaser natively supports [nightly](https://goreleaser.com/customization/publish/nightlies/) builds and uses them itself: https://github.com/goreleaser/goreleaser/releases/tag/nightly
I did brief research recently, please see: https://claude.ai/share/80618728-c1b8-47ec-8476-087b78715169

So if we use goreleaser for all current releases, let's try using it for nightly as well.

Ah, it requires Pro which is a subscription, not a one-time payment hmmm. That's a bummer. I'm also interested if there a way to script/emulate such a behaviour using a regular release process.

psviderski:
A relevant comment from discord:

>Sounds like we probably need to invest some time in configuring the nightly build infrastructure so there is always builds with the latest changes are available. Combining this with `uc machine init/add --version nightly` and a command to upgrade daemons https://github.com/psviderski/uncloud/issues/142 sounds like a great way to test latest changes/fixes.

tonyo:
Switched to using `goreleaser` (also added it to Mise); renamed to nightly.

`uc machine init/add --version nightly` should work now, but will test properly when we have the release in place.

psviderski (approved):
Nice, ship it! 🚀

Issue #307, opened 2026-04-09
Add dynamic completions with ValidArgsFunction


miekg:
I would be cool is `uc` could complete arguments of things running in the cluster:

```
uc service inspect <TAB>
```

and it completes with the current services in the cluster. Similar for other commands. Cobra has support for this:
https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#dynamic-completion-of-nouns

I should be mostly: `clusterConnect`, check what needs finding/expanding, return values.

miekg:
if this is something we want, I'll do the rest in #309. 

I've updated all other commands as well...

miekg:
done

PR #309, opened 2026-04-10
Allow for argument completion

miekg:
This adds completion for cluster interaction for Volumes, Services and Machines. It further uses the machine completion for the -m flag in all commands that have one.

<s>Specifically excluded are local completion for the config file or compose file, ie `uc build` or `uc deploy`, but this can be easily added if we wish.</s>

`uc ctx use ...` added too.
`uc deploy/build` compose service completion added too

See: #307


psviderski:
Just before I look into the code, service listing is quite an expensive operation (not optimised) which can take a second or several seconds for setups with a dozens of containers and when the machines are located far from the client (additional latency). How does this feel in practice?
And also, how does the interactive progress `ClusterConnect` prints work with the completion? Is its stdout just ignored? You can test my prepending an ssh connection to the `connections` list in the config with some random unreachable IP so that `ClusterConnect` will fail on it and then try the next one.

miekg:
> How does this feel in practice?

If tested it with a cluster with one server and 1 service. Machine itself is US (i'm EU), this worked fine. And my shell (zsh) tells me it's completing and I can interrupt it.
(tbf making this faster is orthogonal to having completion work)

>  Is its stdout just ignored?

Yes, at least that's what is happening, because I see no output what so ever during the completion.

I'll try to find some time to test this somewhat more intensive

miekg:
https://github.com/user-attachments/assets/317491cc-02c2-4b72-b7ee-62e7ffc0784f

internal/cli/completion/context.go
```diff
@@ -0,0 +1,24 @@
...
+
+func Contexts(ctx context.Context, uncli *cli.CLI, args []string, toComplete string) ([]cobra.Completion, cobra.ShellCompDirective) {
+	contexts := slices.Sorted(maps.Keys(uncli.Config.Contexts))
+
+	names := []cobra.Completion{}
+	for _, context := range contexts {
+		if slices.Contains(args, context) {
+			continue
+		}
+		names = append(names, context)
```
psviderski, line 24: It doesn't filter by prefix as others do. What's the difference?

psviderski (approved):
Looks great!
What do you think of moving the `internal/completion` package to `internal/cli/completion` as this is scoped to CLI only?

internal/cli/completion/context.go
miekg: ah, forgot I guess.

miekg:
moved the import path

internal/cli/completion/context.go
miekg: added

Issue #311, opened 2026-04-15
Implement a shell prompt to display current cluster context


psviderski:
There are shell prompts for k8s, for example:
- https://github.com/jonmosco/kube-ps1
- https://github.com/Ladicle/fish-kubectl-prompt
- https://github.com/superbrothers/zsh-kubectl-prompt

Implement something similar or explore the ways to integrate into existing most common shell themes/prompts.

We can implement a new `uc ctx show` command for displaying the current context for this. But it can easily be scripted now with `uc ctx ls | grep ✓ | awk '{ print $1 }'`

Relevant issue: https://github.com/psviderski/uncloud/issues/298

PR #314, opened 2026-04-16
feat: uc ctx ls add --short option

miekg:
This just prints the default context and nothing on error. This allows the use of `uc ctx ls --short` in your shell prompt to print the current context however you like. Is also saves on execs, as you don't need a little pipeline to filter stuff.

See: #311

### DIFF
--- a/cmd/uncloud/context/ls.go
+++ b/cmd/uncloud/context/ls.go
@@ -10,32 +10,52 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type listOptions struct {
+	short bool
+}
+
 func NewListCommand() *cobra.Command {
+	opts := listOptions{}
 	cmd := &cobra.Command{
 		Use:     "ls",
 		Aliases: []string{"list"},
 		Short:   "List available cluster contexts.",
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


psviderski (commented):
Thanks for the change but I think the flag naming is a bit inconsistent with other commands. Other `ls` commands and the Docker CLI that `uc` models itself after use `-q/--quiet` for scripting-friendly output, not `--short`.

For example:
- `uc volume ls -q` - only display volume names
- `docker context ls -q`, `docker ps -q`, `docker image ls -q` - only display names or IDs

But the semantics don't match `-q/--quiet` either.
   
That's why I suggested `uc ctx show` subcommand (mirrors `docker context show`). `ls` stays about listing.

There are two main reasons for trying to mirror docker CLI:
1. The projects are closely interconnected so users should appreciate that their Docker knowledge is easily transferable to uncloud or at least they can build intuition. There are new commands and exceptions of course but still many commands could be almost identical.
2. As more people start using AI agents to perform their ops tasks, such similarity builds intuition for LLMs as well and requires less explicit documentation and guiding.

miekg:
makes perfect sense, and I'll take this into account with future prs...
(closing this, no sense in reworking this, I'll open a new one)

PR #316, opened 2026-04-17
feat: ucind: implement cluster list

miekg:
I keep forgetting my ucind clusters, that add the missing list command:
```
% ./ucind cluster list
NAME    MACHINES
miek1   machine-1
```

Just the name would have been enough, but to actually make use of the tui table a machines list is handy as well.
(There isn't much bubbletea in ucind, but I thought lets start now).

### DIFF
--- a/internal/ucind/cluster.go
+++ b/internal/ucind/cluster.go
@@ -323,7 +323,11 @@ func (p *Provisioner) WaitClusterReady(ctx context.Context, c Cluster, timeout t
 }
 
 func (p *Provisioner) ListClusters(ctx context.Context) ([]Cluster, error) {
-	nets, err := p.dockerCli.NetworkList(ctx, network.ListOptions{})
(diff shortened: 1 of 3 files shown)


internal/ucind/cluster.go
```diff
@@ -322,6 +322,22 @@ func (p *Provisioner) WaitClusterReady(ctx context.Context, c Cluster, timeout t
 	return nil
 }
 
+func (p *Provisioner) ListClusters(ctx context.Context) ([]Cluster, error) {
+	nets, err := p.dockerCli.NetworkList(ctx, network.ListOptions{})
```
psviderski: Let's filter only `ucind.managed` networks so it's obvious why we list and use network here.

psviderski (approved):
Thanks 👍  Just one comment about filtering uncloud-only networks.

internal/ucind/cluster.go
miekg: done

PR #317, opened 2026-04-17
feat: add uc ctx show

miekg:
This only shows the current context and nothing more. On error nothing is printed (also no error). This is useful to include in your prompt.

See: #314


psviderski (approved):
👍

Issue #140, opened 2025-10-10
Add advanced "exec" capabilities


tonyo:
"uc exec" will be added as part of https://github.com/psviderski/uncloud/pull/139, but it'll mostly replicate "docker exec" in terms of CLI-interface and also limitations.

We don't have to stop there and can give "uc exec" superpowers, e.g. in the form of https://github.com/iximiuz/cdebug. It'll allow users to debug slim and restricted containers, similar to "docker debug" [1] or "kubectl debug".

[1] https://docs.docker.com/reference/cli/docker/debug/

miekg:
TIL about debug containers, that *is* handy. from looking a bit at *cdebug*, we can easily lift that into uncloud, the gist is `ContainerAttach` which is a docker API. Looks to be one or two functions and hooking it up to the infra we already have (and of course working out all of the details)

miekg:
we have: `uc exec [OPTIONS] SERVICE [COMMAND ARGS...] [flags]`
docker has:  `docker debug [OPTIONS] {CONTAINER|IMAGE}`
(of course this is not docker compose).

I can't figured out if you are supposed to configure the debug image as well, seems like something useful.

our container option is useful here as well:
`uc exec uc exec --container d792e web-service ls -la
ls -la`

So I guess: `uc debug --container d792e web-service`
with the --shell options *and* an extra option to set the debug image seems like a good idea

tonyo:
One question we should probably answer is whether we need both `uc exec` and `uc debug` (assuming it's similar to `cdebug`). As in, can we perhaps extend `exec` with additional flags that enable the `cdebug` features?

Thinking out loud, arguments in favor of keeping them separate:
* kubectl/docker also have two separate commands for `exec` and `debug`, so there'll be semantic/mental parity with the existing tools.
* Each command has a similar, but still different contract: `exec` = run in existing container, debug = attach to the existing container but with a user-defined execution context.
* `exec` stays scriptable and composable (-T, pipes, CI usage) without debug-specific flags polluting its surface.
* `debug` can evolve independently and add some ambitious features (debugging crashed/stopped containers) without touching exec.
* Security boundary is more explicit: exec reuses container perms; debug is inherently privileged/root, so keeping them separate makes the security model obvious.

Arguments against:
* "debug" was a later addition in both kubectl and docker, so it was natural not to overload the existing and stable `exec` with new features. With Uncloud, we can rethink the approach and pack both exec and debug features in one command.
[...]

PR #318, opened 2026-04-19
feat: debug container: add AttachContainer infra

miekg:
This adds the lowlevel grpc and docker calling code to call AttachContainer, called ContainerAttach here.

cdebug/cmd/exec/exec_docker.go has served as a guide on what is needed to run a debug container, and AttachContainer is the low-level building block.

See: #140


psviderski:
I think there is much more to it to fully support the `AttachContainer` API call. I expect it to be similar to how we implemented `ContainerExecAttach` for attaching to the container created as part of the `uc service exec` command: https://github.com/psviderski/uncloud/blob/1c8b77054a22fcb076879b33891dceb6c427222b/internal/machine/docker/server.go#L1402-L1430

It's a bit cumbersome because we essentially need to convert the TTY IO twice to proxy it through the gRPC API:
```
Docker container TTY <-> gRPC stream <-> CLI TTY
```

See full PR: https://github.com/psviderski/uncloud/pull/139

I wonder if we can piggy back the `exec` for implementing `debug`.

The essential part of the debug implementation is running a new debug container that shares all the namespaces of the container we want to debug. This is very similar to what `ContainerExecCreate` Docker API call does. But it uses the same image as the target container.

We want to use a different image but Docker API doesn't provide such a method so we need to create our own. We can run something like `pause`/`sleep infinity` inside this container and then just run the debug shell/command using `exec` instead of attaching to the container's main process.

This way we can reuse most of the `exec` routine without having to implement `AttachContainer`. If we still need to implement a new `AttachContainer` method, I'd suggest reusing most of the exec's implementation as they will be almost identical.

miekg:
ack, closing with 'good to have, not like this', can always resurrect.

Issue #324, opened 2026-04-22
 `uc machine logs uncloud -f` hangs 


miekg:
@miekg FYI `uc machine logs uncloud -f` hangs in my tests and prints a warning about stalled streams. Also zombie `journalctl` processes start to appear on the remote machine(s):
```
ps ax | grep journal
    242 ?        Ss    21:59 /lib/systemd/systemd-journald
 286152 ?        Z      0:00 [journalctl] <defunct>
 286153 ?        Z      0:00 [journalctl] <defunct>
 286180 ?        Z      0:00 [journalctl] <defunct>
 286181 ?        Z      0:00 [journalctl] <defunct>
 286197 ?        Z      0:00 [journalctl] <defunct>
 286218 ?        Z      0:00 [journalctl] <defunct>
 286219 ?        Z      0:00 [journalctl] <defunct>
 286230 ?        Z      0:00 [journalctl] <defunct>
 286238 ?        Z      0:00 [journalctl] <defunct>
 286248 ?        Z      0:00 [journalctl] <defunct>
 286479 ?        Z      0:00 [journalctl] <defunct>
 286487 ?        Z      0:00 [journalctl] <defunct>
 286488 ?        Z      0:00 [journalctl] <defunct>
 286496 ?        Z      0:00 [journalctl] <defunct>
 286497 ?        Z      0:00 [journalctl] <defunct>
 286499 ?        Z      0:00 [journalctl] <defunct>
```

Note that `-f -n N` where N is any number < 100 works but hangs >=100. This could be related to this constant:
https://github.com/psviderski/uncloud/blob/cef047221c4f103f2997837e2209a317c389319f/pkg/client/logmerger.go#L15

But I haven't seen this behaviour for service logs. So I suspect we're not correctly handling timestamps and heartbeats when streaming journal logs.
Zombie processes might be an unrelated issue.

_Originally posted by @psviderski in https://github.com/psviderski/uncloud/issues/283#issuecomment-4288381645_

miekg:
Just running `uc machine logs uncloud -f` shows:

```
Apr 22 08:50:42.985 machine-9q7y uncloud uncloudd[777]: DEBUG Sent log stream heartbeat. unit=uncloud stream_id=3d70246e850 timestamp=2026-04-22T06:50:42.784Z
Apr 22 08:50:43.167 machine-9q7y uncloud uncloudd[777]: DEBUG Syncing containers to cluster store triggered by a regular interval. interval=30s
[...]

PR #325, opened 2026-04-22
fix: call cmd.Wait()

miekg:
Even though the context is cancelled we still need to call cmd.Wait() after a cmd.Start() to clean up the child (reap) process. Not doing so results in a zombie journalctl.

I have manually tested this, as I'm still not sure how to e2e test for this in a simple manner.

Before:

```
root@uncloud2:~# ps aux|grep jou
root         313  0.0  1.2  42272 23352 ?        S<s  Apr20   0:10 /usr/lib/systemd/systemd-journald
root       15751  0.0  0.0      0     0 pts/0    Z+   07:02   0:00 [journalctl] <defunct>
root       15754  0.0  0.0      0     0 pts/0    Z+   07:02   0:00 [journalctl] <defunct>
```

After:

```
root         313  0.0  1.2  42272 23592 ?        S<s  Apr20   0:10 /usr/lib/systemd/systemd-journald
```

### DIFF
--- a/internal/journal/logs.go
+++ b/internal/journal/logs.go
@@ -16,7 +16,7 @@ func Logs(ctx context.Context, unit string, opts api.ServiceLogsOptions) (<-chan
 		return nil, fmt.Errorf("journal logs: invalid unit: %s", unit)
 	}
 
-	reader, err := logs(ctx, unit, opts)
+	reader, wait, err := logs(ctx, unit, opts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


miekg:
Fixes: #324

internal/journal/logs.go
```diff
@@ -25,7 +25,7 @@ func Logs(ctx context.Context, unit string, opts api.ServiceLogsOptions) (<-chan
 
 	go func() {
 		defer close(outCh)
-		follow(ctx, reader, outCh)
```
psviderski, line 28: We can probably defer wait here so follow doesn't even need to know anything about it

psviderski (approved):
👍 one nit comment but otherwise looks good

internal/journal/logs.go
miekg: let me check, that would be nicer indeed

PR #327, opened 2026-04-22
feat: add uc context create

miekg:
This gets the context from the remote machine and add it to the correct place in your uncloud config.

### DIFF
--- a/cmd/uc/context/create.go
+++ b/cmd/uc/context/create.go
@@ -1,43 +1,45 @@
-package machine
+package context
 
 import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/goccy/go-yaml"
 	"github.com/psviderski/uncloud/internal/cli"
 	"github.com/psviderski/uncloud/internal/cli/config"
+	"github.com/psviderski/uncloud/internal/machine/api/pb"
 	"github.com/spf13/cobra"
 )
 
-type contextOptions struct {
+type createOptions struct {
 	context string
 	sshKey  string
 	write   bool
 }
 
-func NewContextCommand() *cobra.Command {
-	opts := contextOptions{}
+func NewCreateCommand() *cobra.Command {
+	opts := createOptions{}
 	cmd := &cobra.Command{
-		Aliases: []string{"context"},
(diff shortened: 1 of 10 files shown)


miekg:
```
% ./uc machine context test@uncloud1.vm.science.ru.nl -c bla   
uncloud:
  connections:
    - ssh: test@131.174.88.77
      machine_id: c22073beee04bf9b1861e53c21023d2d
    - ssh: test@131.174.88.78
      machine_id: 212fd73438b53ebfb7dd531da4c6ed41
```
the context is `uncloud` because the machine(s) ID already are in that section in my uncloud config

miekg:
where should this be tested? test/e2e/machine_test.go ?

miekg:
weird

miekg:
a user just came in asking fo such a feature. If you have specifics on why this doesn't work in some usecases let me know so I can brush up the PR

psviderski:
From the Discord discussion, I think something like the following feels like a more generic approach:
```
uc context create mycluster --connect user@host1 [--connect user@host2]
```
this generates and adds the following to `~/.config/uncloud/config.yaml`:
```
contexts:
  mycluster:
    connections:
      - ssh: user@host1
        machine_id: .... # inspected for each user@hostX
```
and optionally sets the `current_context`.

`PublicIP` doesn't work in the cases when machines don't have public IPs, which is a common use cases for office/home clusters behind NAT or hybrid clusters when only a subset of machines are public. Public IP in those cases is the IP of the egress router which won't work as the connection to the machine.

That's why I'm suggesting to require a user to explicitly specify what connections to use and their hosts/IPs.

We don't need to list all the machines in the connections. So I'd say listing only 1 machine with the correct user@host is more preferable than all machines with potentially incorrect addresses.

miekg:
> We don't need to list all the machines in the connections.

technically correct, but this is also something that serves as documentation, which machines do we actually have in our cluster. (Yes uc machine ls also does this - but having a complete listing in the config would help).

Point taken on the public ips, those machines prolly need to be filtered out? (I.e. a different route then your suggestion)

miekg:
this now is: us context create and only add machines that have a public IP address

PTAL

PR #333, opened 2026-04-25
feat: allow binding to an interface

### LINKED ISSUE #332: Support binding to an named interface in x-ports
> of course this is what I need:
> ```
>     x-ports:
>       - 80:80@host
>       - 443:443@host
>       - 443:443/udp@host
>       - 131.174.88.77:53:53@host
>       - 131.174.88.77:53:53/udp@host
>       - 131.174.88.78:53:53@host
>       - 131.174.88.78:53:53/udp@host
> ```
> 
> but this fails because of IP doesn't match. What is needed here is an interface name instead of an IP address, so I can bind to an interface and have it be a global service 

 _Originally posted by @miekg in [#320](https://github.com/psviderski/uncloud/issues/320#issuecomment-4318516739)_

miekg:
feat: x-port add interface specification
When using x-ports in the host mode you can now also specify an
interface name. Uncloudd will when starting the container resolve the IP
address from that interface and use those to bind to.

When running a global service you may need this, as you can bind to IP
addresses these are obviously different on each machine. But using an
interface name this is more portable.

Fixes: https://github.com/psviderski/uncloud/issues/332

Signed-off-by: Miek Gieben <miek@miek.nl>

### DIFF
--- a/cmd/uncloud/service/run.go
+++ b/cmd/uncloud/service/run.go
@@ -100,7 +100,8 @@ func NewRunCommand(groupID string) *cobra.Command {
 			"  -p app.example.com:8080/https  Publish port 8080 as HTTPS via reverse proxy with custom hostname\n"+
 			// TODO: add support for publishing L4 tcp/udp ports.
 			//"  -p 9000:8080                   Publish port 8080 as TCP port 9000 via reverse proxy\n"+
-			"  -p 53:5353/udp@host            Bind UDP port 5353 to host port 53")
+			"  -p 53:5353/udp@host            Bind UDP port 5353 to host port 53"+
+			"  -p enp1s0:54:5443@host         Bind TCP port 5453 to host port 54 on interface enp1s0")
 	cmd.Flags().StringVar(&opts.pull, "pull", api.PullPolicyMissing,
 		fmt.Sprintf("Pull image from the registry before running service containers ('%s', '%s', '%s').",
 			api.PullPolicyAlways, api.PullPolicyMissing, api.PullPolicyNever))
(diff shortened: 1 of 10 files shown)


miekg:
with this compose' x-ports
``` yaml
    x-ports:
      - 80:80@host
      - 443:443@host
      - 443:443/udp@host
      - enp1s0:53:53@host
      - enp1s0:53:53/udp@host
```

and I see:
```
root@uncloud1:~# docker ps
CONTAINER ID   IMAGE                                                COMMAND                  CREATED              STATUS              PORTS                                                                                                                                                                                 NAMES
d9494ed8c838   registry.science.ru.nl/cncz/sys/image/caddy:latest   "caddy run -c /confi…"   About a minute ago   Up About a minute   0.0.0.0:80->80/tcp, [::]:80->80/tcp, 131.174.88.77:53->53/tcp, 131.174.88.77:53->53/udp, 0.0.0.0:443->443/tcp, [::]:443->443/tcp, 0.0.0.0:443->443/udp, [::]:443->443/udp, 2019/tcp   caddy-g9hq
```

psviderski:
@miekg I was going to leave a comment on this PR about the idea but you've just closed it. What's the reason?

>When running a global service you may need this, as you can bind to IP addresses these are obviously different on each machine. But using an interface name this is more portable.

I don't really like this idea to be honest as the interface name doesn't make the Compose configuration more portable. I have several linux boxes and interface names are all different. I guess it got worse when Linux distributions started using interface names based on the hardware location such as `ensX` or `enpXsY` instead of traditional `eth0`, `eth1`.

miekg:
ah fuck, not enough coffee! re-opened

miekg:
yeah, it's slightly _more_ portable, at least in our case.

miekg:
As I said on discord, rework this to allow a cidr notation? `131.174.115.0/24:bla` this would do similar matching, but is more inline with the current way of doing things, can be properly syntax checked *and* solves my problem (and more portable)

I'll do this anyway, can discuss later anyway

miekg:
closing in favor of #358

PR #334, opened 2026-04-25
fix: sync local machine name from cluster store on machine changes

jabr:


internal/machine/cluster.go
```diff
@@ -541,6 +543,29 @@ func (cc *clusterController) handleMachineChanges(ctx context.Context) error {
...
+
+		cc.state.mu.Lock()
+		cc.state.Name = m.Name
+		if err := cc.state.Save(); err != nil {
+			slog.Error("Failed to save local machine state after name change.", "err", err)
+		} else {
+			slog.Info("Synced local machine name from cluster store.", "name", cc.state.Name)
+		}
+		cc.state.mu.Unlock()
+		return
```
miekg, line 565: this seems to do only the first machine in the list, is that intentional?

jabr: It's just looking for itself in the list. It skips entries until it finds its id (`cc.state.ID`). Then if the name has changed, it updates and saves its `machine.json` file.

miekg:
Ack. Then LGTM

psviderski (commented):
The fix itself looks good but I think we probably should simply get rid of the machine name from the `machine.json` state file and don't use it anywhere. Let me give you some context.

This file was introduced long time ago, before we even settled on using Corrosion IIRC, and put the authoritative data there. Then we started sharing the machine info via the `machines` table in Corrosion.

I was still thinking we should treat this file as the source of truth for the machine info, hence to update it, we need to route requests to a particular machine which will update the info in `machines.json` and then sync it to the `machines` in Corrosion. So similar to `containers`, it should be possible to easily recreate this table from the sources of truth and make it harder for machines to update the info of their peers.

However, an external contributor implemented machine info updates right in corrosion: https://github.com/psviderski/uncloud/pull/91. I was a bit hesitant with this approach but decided to not worry about this too much so went with this approach. The `machines` table became the source of truth at least for the machine names and WG endpoints.

I still don't like the fact that any machine in the cluster can update any data in Corrosion. This is just the Corrosion design limitation and I don't know how we will address this yet but we will need to at some point.

It's a bit ironic that we started with the idea to sync `machine.json` to Corrosion and ended up with this PR suggesting to sync Corrosion back to `machines.json`.

To not complicate things and reduce duplications, I would suggest to not do that and just remove the machine name field from the machine state. Or at least start with replacing all places that use the machineName from it that is not updated on `uc machine rename`.

Then we will at least have only one source where the name comes from (Corrosion) so it will be easier to track and change it in the future.

psviderski:
Alternatively, we can update the `UpdateMachine` logic to route the request to the machine being updated and do the update in both Corrosion and `machine.json` state.

Tbh, this feel more correct and safer to me. And we can also return an error for requests that try to update the machine info for the machine that is not the one receiving the request. This doesn't solve the Corrosion permission model but at least it won't continue weakening it even further.

Like with containers, we route requests to specific machines to manage Docker containers on them. And then containers from Docker are synced to Corrosion. I prefer thinking of Corrosion as the mechanism to share data in the cluster, and not as the primary source for the data that any cluster member can manipulate without any limitation.

psviderski:
Fixed this properly by making each machine the owner of its configuration and syncing local state to cluster store. Update/rename requests are also routed through the machines they update/rename.

https://github.com/psviderski/uncloud/commit/23d6b4409a85ad3d97f514b4f6dbf066328f6771

PR #344, opened 2026-04-30
feat: add completion for ucind

miekg:
For feature parity with uc. Only _really_ usefull for ucind cluster rm, but still useful, and not many lines of code.


psviderski (approved):
👍

PR #342, opened 2026-04-29
Add --output=json to ls subcommands

miekg:
This PR add --output=json to _only_ the command that have an ls.go. The json keys are copied from docker and are camelCase, json struct tags are added where needed.

It does _not_ change any of the inspect commands, that should be another issue. It also doesn't do any prework for future output formats, it only handles json.

Fixes: #338 #337

PR #345, opened 2026-04-30
WIP: add jsonconv

miekg:
This package converts snake case (public_key) to camel case (PublicKey) for json output. We need this so all json generated by Uncloud has similar type of keys.

Specificaly generate json from grpc using different keys (snake case) and this isn't easily changed, hence to two pass to fix things.

This isn't a huge performance issue as this output is generated by running `uc`, it is not used for internal rpc or some such.

See discussion in #342 if we want this or not.

### DIFF
--- a/internal/jsonconv/conv.go
+++ b/internal/jsonconv/conv.go
@@ -1,0 +1,68 @@
+// Package jsonconv converts JSON keys into camel case.
+package jsonconv
+
+import (
+	"encoding/json"
+	"strings"
+)
+
+// Marshal converts x into JSON, while also normalizing the keys.
+func Marshal(x any) ([]byte, error) {
+	data, err := json.Marshal(x)
+	if err != nil {
+		return nil, err
+	}
+
+	var jdata any
+	if err := json.Unmarshal(data, &jdata); err != nil {
+		return nil, err
+	}
+	processed := normalize(jdata)
+	return json.MarshalIndent(processed, "", "  ")
+}
+
+func normalize(v any) any {
+	switch obj := v.(type) {
+	case map[string]any:
+		m := make(map[string]any, len(obj))
+		for k, val := range obj {
+			m[camelCase(k)] = normalize(val)
+		}
+		return m
+	case []any:
+		for i, val := range obj {
+			obj[i] = normalize(val)
+		}
+		return obj
+	}
+	return v
+}
+
+// camelCase converts 'public_key' to 'PublicKey'. Words like id, url are recognized and converted to ID and
+// URL respectively. If a key contains dots (mostly labels, it is left alone).
+func camelCase(s string) string {
+	if strings.Contains(s, ".") {
+		return s
+	}
+
+	if strings.ToLower(s) != s { // probably alredy in camelcase
+		return s
+	}
+
+	splits := strings.Split(s, "_")
+	sb := strings.Builder{}
+	for _, split := range splits {
+		switch split {
+		case "id", "Id":
+			fallthrough
+		case "url", "Url":
+			fallthrough
+		case "tcp", "Tcp":
+			sb.WriteString(strings.ToUpper(split))
+		default:
+			sb.WriteString(strings.ToUpper(split[0:1]))
+			sb.WriteString(strings.ToLower(split[1:]))
+		}
+	}
+	return sb.String()
+}
--- a/internal/jsonconv/conv_test.go
+++ b/internal/jsonconv/conv_test.go
@@ -1,0 +1,16 @@
+package jsonconv
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCamelCase(t *testing.T) {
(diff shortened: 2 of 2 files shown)


psviderski:
What do you think about not directly exposing/using the internal gRPC structs for the command output? Instead, create a new clean "view" struct for the Machine where we can define the fields we want to show, the case, and the values format, e.g. for enums. We should probably use that new struct in the client as well as the public API so a user doesn't need to work with more low-level auto-generated structs from the proto.

It's a bit more work to convert between them but fully owning the `api` structs provides friendlier Go-native experience and allows to easily add custom methods to them if needed.
IDK, am I not using proto the right way?

miekg:
Yeah thought about it, it also works, but more work to keep them in sync
(also we might be able to autogen most of it?)
In the long run this probably the best solution.

For now, this might work as an intermediate, but happy to attempt doing the
better thing now.

Also I think we should very clear about the unstable nature of the json
output for the time being.

miekg:
I think I have an even better idea, even though this is hacky, but I think it will work.

We do _own_ the protobuf definition and it is just Go code, so why don't we rewrite the generated protobuf files and just update the - in our eyes - correct json tags..., I don't think they are used for something else

that means reworking this PR in a go generate that becomes part of the build process. 

and look it already exists: https://github.com/fatih/gomodifytags
(shold prolly need a little more work, but ....)

miekg:
```
gomodifytags -file machine.pb.go -all -remove-tags json | gomodifytags -file /dev/stdin -all -add-tags json -add-options json=omitempty -transform pascalcase | less
```
til this is actually called pascalcase. 
```go
type MachineInfo struct {
        state         protoimpl.MessageState  `json:"State,omitempty"`
        sizeCache     protoimpl.SizeCache     `json:"SizeCache,omitempty"`
        unknownFields protoimpl.UnknownFields `json:"UnknownFields,omitempty"`

        Id       string         `protobuf:"bytes,1,opt,name=id,proto3" json:"Id,omitempty"`
        Name     string         `protobuf:"bytes,2,opt,name=name,proto3" json:"Name,omitempty"`
        Network  *NetworkConfig `protobuf:"bytes,3,opt,name=network,proto3" json:"Network,omitempty"`
        PublicIp *IP            `protobuf:"bytes,4,opt,name=public_ip,json=publicIp,proto3" json:"PublicIp,omitempty"`
}
```
is exactly what we want, and saves (a quite the hassle) redefining everything

psviderski:
>I don't think they are used for something else

The serialised JSONs are stored in Corrosion in the `machines` table. So this will be a breaking change we need to account.

It fixes JSON keys but does nothing for Go ergonomics. The caller still receives `*pb.IP`, `*pb.IPPort`, `*pb.MachineMember`, custom `Duration` and `Timestamp` and have to call `ToAddr()`, `ToPrefix()`, handle `nil`s and errors everywhere. In some places I even started ignoring the `err` returned by these conversion methods when I was sure the data was validated before. It's just so much boilerplate.

Proto types can't carry domain methods (`Validate`, `Clone`, `Equal`) and you still can't automatically attach methods like `Equals` on regenerated types without risk anyway.

I think we still want to maintain Go-native *domain* data types at least for commonly used resources such as `Machine`.

They also not always map 1:1 to the RPC wire data types. For example, we have `Service` that doesn't exist at the RPC level and is comprised of individual container calls fully on the client. This might change in the future and we can change the RPC wire level potentially without even affecting the user-facing `Service`.

Our `client` methods are also not always 1:1 mapping to the RPC methods because we often need to do the low level proxy/broadcast routine. So far we've been pretty successful at hiding these complexities from the user of the `client` but still provide a clear interface, for example, to specify if a particular action needs to run on a specific machine using a simple argument or a field in an options struct.

Now, it would be great to draw a clear line what we should mirror/duplicate and what not.

Mirroring all the proto types feels like it defeats the whole purpose of using proto in the first place. I don't know to be honest and I'm fine to not overthink it.

I would probably not bother to recreate type like `UpdateMachineRequest`, `JoinClusterRequest`, `AddMachineRequest` which are used only in one method and which are used very rarely. The only problem is inconsistency. But again we have much bigger fish to fry at the moment than to try to perfect these types.
I'm sure we will work out some intuition and maybe better ideas after working more with those.

miekg:
> The serialised JSONs are stored in Corrosion in the machines table. So this will be a breaking change we need to account.

ah, then this route is a no-go

> The serialised JSONs are stored in Corrosion in the machines table. So this will be a breaking change we need to account.

Very true. Hence the WIP status here, and good to have some docs on subject. Also saying somewhere that json outout just like the text output is not stable seems like a good starting point.

psviderski:
>just like the text output is not stable seems like a good starting point.

What do you mean when you say JSON is not stable? What scenario do you have in your mind you want to warn about?

miekg:
user parsers json to do a thing, we update the json (`public_key` -> `PublicKey`), json parsing fails.

"Until version *X*, the JSON output can change in backwards incompatible ways between Uncloud versions" - or something like that.

PR #351, opened 2026-05-02
feat: uncloudd add periodic image pruning

### LINKED ISSUE #350: uncloudd: prune images
Via discord:

Should uncloudd just prune images by default. Prune images that haven't seen use in 24 (?) hours.

Pruning images _always_ comes up in docker deployments.

miekg:
Run a go-routine that prunes images older than 24h every hour.

Fixes: #350


miekg:
Once we have a uc prune command, this can also done from a image having `uc` and the socket... Although it also be done via docker prune in that image, albeit not with all those options

miekg:
Note this can also be done externally: https://miek.nl/2026/may/14/uncloud-image-pruning/, which is even more elegant

psviderski:
>Note this can also be done externally: https://miek.nl/2026/may/14/uncloud-image-pruning/, which is even more elegant

Looks like you're slowly building kubernetes with dozens of extensions and controllers running on the cluster lol

miekg:
> > Note this can also be done externally: https://miek.nl/2026/may/14/uncloud-image-pruning/, which is even more elegant
> 
> Looks like you're slowly building kubernetes with dozens of extensions and controllers running on the cluster lol

oh god... yes

PR #353, opened 2026-05-04
feat: error when using relative volumes sources in compose

### LINKED ISSUE #346: Return a clear error when bind-mounting relative path
When attempting to deploy the following `compose.yaml`:
```yaml
services:
  atomdns:
    #    image: registry.science.ru.nl/cncz/sys/image/atomdns:latest
    image: debian:bookworm-slim
    entrypoint:
      - /bin/sh
      - "-c"
      - "ls -l /conf/Conffile"
    x-ports:
      - ctrl.uncld.science.ru.nl:443/https

    deploy:
      replicas: 1

    volumes:
      - ./atomdns:/conf:ro
      - atomdns_data:/var/lib/atomdns

volumes:
  atomdns_data:
```

it will attempt to mount `/Users/richard/projects/uc/atom/atomdns` on the remote machine but this is the path resolved from `./atomdns` on the local machine.

```
$ docker container inspect d32e134ec731
...
 "Mounts": [
            {
                "Type": "bind",
                "Source": "/Users/richard/projects/uc/atom/atomdns",
                "Destination": "/conf",
                "Mode": "",
                "RW": false,
                "Propagation": "rprivate"
            },
            {
                "Type": "volume",
                "Name": "atomdns_data",
                "Source": "/var/lib/docker/volumes/atomdns_data/_data",
                "Destination": "/var/lib/atomdns",
                "Driver": "local",
                "Mode": "z",
                "RW": true,
                "Propagation": ""
            }
        ],
...
```

We should fail with a clear error message that mounting a relative path won't work as a user might expect. And we can hint that using [`configs`](https://uncloud.run/docs/concepts/configs) is likely what the user needs.

See original issue: https://github.com/psviderski/uncloud/issues/330

miekg:
Add a warning when using compose with a relative volume source, as this like does not do what you want.

Warning:
```
WARNING: Invalid volume mount, volume name './atomdns' is relative. If you intended to pass a host host directory or file,
use absolute path, or configs might be better suited for this use case. See https://docs.docker.com/reference/compose-file/configs/
```

This warning is slightly amended from `uc run` which straight up disallows these kind of paths:

```
Error: invalid volume mount './atomdns:/conf:ro': volume name './atomdns'
includes invalid characters, only '[a-zA-Z0-9][a-zA-Z0-9_.-]' are allowed. If you intended to pass a host directory or file, use absolute path
```

Fixes: #346

### DIFF
--- a/pkg/client/compose/project.go
+++ b/pkg/client/compose/project.go
@@ -129,9 +129,9 @@ func checkRelativeVolumeMount(data any, _ tree.Path, _ bool) (any, error) {
 	}
 	if strings.HasPrefix(source, "./") || strings.HasPrefix(source, "../") { // only check actual paths, not volumes _names_
 		// uc run also warns against this
-		tui.PrintWarning("Invalid volume mount, volume name '" + source + "' is relative. If you intended to pass a host " +
-			"host directory or file, use absolute path, or configs might be better suited for this use case. " +
-			"See https://docs.docker.com/reference/compose-file/configs/")
+		return nil, fmt.Errorf("invalid volume mount, volume name '%s' is relative. If you intended to pass a host "+
+			"host directory or file, use absolute path, or configs might be better suited for this use case. "+
+			"See https://docs.docker.com/reference/compose-file/configs/", source)
 	}
 
 	return data, nil
--- a/pkg/client/compose/project.go
+++ b/pkg/client/compose/project.go
@@ -22,6 +22,7 @@ var registerComposeOverrides sync.Once
 func LoadProject(ctx context.Context, paths []string, opts ...composecli.ProjectOptionsFn) (*types.Project, error) {
 	registerComposeOverrides.Do(func() {
 		transform.RegisterDefaultValue("services.*.deploy.update_config", setUpdateConfigDefaults)
+		transform.RegisterDefaultValue("services.*.volumes.*.source", checkRelativeVolumeMount)
 	})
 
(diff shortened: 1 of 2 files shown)


pkg/client/compose/project.go
```diff
@@ -119,3 +120,19 @@ func setUpdateConfigDefaults(data any, _ tree.Path, _ bool) (any, error) {
...
+
+// checkRelativeVolumeMount check if a volume mount uses a relative path.
+func checkRelativeVolumeMount(data any, _ tree.Path, _ bool) (any, error) {
+	source, ok := data.(string)
+	if !ok || filepath.IsAbs(source) {
+		return data, nil
+	}
+	if strings.HasPrefix(source, "./") || strings.HasPrefix(source, "../") { // only check actual paths, not volumes _names_
+		// uc run also warns against this
+		tui.PrintWarning("Invalid volume mount, volume name '" + source + "' is relative. If you intended to pass a host " +
```
psviderski: Why warning? Let's fail loudly. I can't think of any reasonable use case where a relative path resolved on the client makes any sense on the remote server.

psviderski (commented):
One comment about maybe making it an error instead of a warning.

Regarding `uc run`, did you implement something and forgot to include in this PR?

miekg:
> Regarding uc run, did you implement something and forgot to include in this PR?

No, that already errors

PR #358, opened 2026-05-06
feat: allow ip prefix in x-ports

miekg:
This allows you to specify a, portable but global compose that binds only to certain interfaces. Namely those that are contained in the prefix spec.

This should a better idea than #333

### DIFF
--- a/cmd/uncloud/service/run.go
+++ b/cmd/uncloud/service/run.go
@@ -97,11 +97,12 @@ func NewRunCommand(groupID string) *cobra.Command {
 			"Supported protocols: tcp, udp, http, https (default is tcp). If a hostname for http(s) port is not specified\n"+
 			"and a cluster domain is reserved, service-name.cluster-domain will be used as the hostname.\n"+
 			"Examples:\n"+
-			"  -p 8080/https                  Publish port 8080 as HTTPS via reverse proxy with default service-name.cluster-domain hostname\n"+
-			"  -p app.example.com:8080/https  Publish port 8080 as HTTPS via reverse proxy with custom hostname\n"+
+			"  -p 8080/https                       Publish port 8080 as HTTPS via reverse proxy with default service-name.cluster-domain hostname\n"+
+			"  -p app.example.com:8080/https       Publish port 8080 as HTTPS via reverse proxy with custom hostname\n"+
 			// TODO: add support for publishing L4 tcp/udp ports.
 			//"  -p 9000:8080                   Publish port 8080 as TCP port 9000 via reverse proxy\n"+
-			"  -p 53:5353/udp@host            Bind UDP port 5353 to host port 53")
+			"  -p 53:5353/udp@host                 Bind UDP port 5353 to host port 53\n"+
+			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on all interfaces that contain the prefix 192.168.76.0/24")
 	cmd.Flags().StringVar(&opts.pull, "pull", api.PullPolicyMissing,
 		fmt.Sprintf("Pull image from the registry before running service containers ('%s', '%s', '%s').",
 			api.PullPolicyAlways, api.PullPolicyMissing, api.PullPolicyNever))
--- a/cmd/uncloud/service/run.go
+++ b/cmd/uncloud/service/run.go
@@ -93,7 +93,7 @@ func NewRunCommand(groupID string) *cobra.Command {
 		"Give extended privileges to service containers. This is a security risk and should be used with caution.")
 	cmd.Flags().StringSliceVarP(&opts.publish, "publish", "p", nil,
 		"Publish a service port to make it accessible outside the cluster. Can be specified multiple times.\n"+
-			"Format: [hostname:]container_port[/protocol] or [host_ip:]host_port:container_port[/protocol]@host\n"+
+			"Format: [hostname:]container_port[/protocol] or [host_ip|host_prefix:]host_port:container_port[/protocol]@host\n"+
 			"Supported protocols: tcp, udp, http, https (default is tcp). If a hostname for http(s) port is not specified\n"+
 			"and a cluster domain is reserved, service-name.cluster-domain will be used as the hostname.\n"+
 			"Examples:\n"+
@@ -102,8 +102,8 @@ func NewRunCommand(groupID string) *cobra.Command {
 			// TODO: add support for publishing L4 tcp/udp ports.
 			//"  -p 9000:8080                   Publish port 8080 as TCP port 9000 via reverse proxy\n"+
 			"  -p 53:5353/udp@host                 Bind UDP port 5353 to host port 53\n"+
-			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on all interfaces that have addresses\n"+
-			"                                      that are contained in the prefix 192.168.76.0/24")
+			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on every host IP address\n"+
+			"                                      contained in the prefix 192.168.76.0/24")
 	cmd.Flags().StringVar(&opts.pull, "pull", api.PullPolicyMissing,
 		fmt.Sprintf("Pull image from the registry before running service containers ('%s', '%s', '%s').",
 			api.PullPolicyAlways, api.PullPolicyMissing, api.PullPolicyNever))
--- a/cmd/uncloud/service/run.go
+++ b/cmd/uncloud/service/run.go
@@ -102,7 +102,8 @@ func NewRunCommand(groupID string) *cobra.Command {
 			// TODO: add support for publishing L4 tcp/udp ports.
 			//"  -p 9000:8080                   Publish port 8080 as TCP port 9000 via reverse proxy\n"+
 			"  -p 53:5353/udp@host                 Bind UDP port 5353 to host port 53\n"+
-			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on all interfaces that contain the prefix 192.168.76.0/24")
+			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on all interfaces that have addresses\n"+
+			"                                      that are contained in the prefix 192.168.76.0/24")
 	cmd.Flags().StringVar(&opts.pull, "pull", api.PullPolicyMissing,
 		fmt.Sprintf("Pull image from the registry before running service containers ('%s', '%s', '%s').",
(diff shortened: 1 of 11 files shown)


miekg:
finally came around to actually deploy this, with the fixed I pushed:

~~~
% uc deploy -y     
Deployment plan

+ create service caddy (global)
  │   image: registry.science.ru.nl/cncz/sys/image/caddy:v0.1.14
  │
  ├── +   run container caddy on machine-lyrk
  ╰── +   run container caddy on machine-vkre

────────────────────────────
2 create · across 2 machines

[+] Deploying 6/6
 ✔ Container caddy-el2a on machine-lyrk                                   Running                                                      7.3s
   ✔ Image registry.science.ru.nl/cncz/sys/image/caddy:v0.1.14 on machine-lyrk  Pulled                                                 1.9s
 ✔ Container caddy-yntr on machine-vkre                                   Running                                                      6.5s
   ✔ Image registry.science.ru.nl/cncz/sys/image/caddy:v0.1.14 on machine-vkre  Pulled                                                 1.1s
%    
% uc ls        
NAME    MODE     REPLICAS   IMAGE                                                 ENDPOINTS
caddy   global   2          registry.science.ru.nl/cncz/sys/image/caddy:v0.1.14   (custom Caddy config)
%   
% uc ls|cat 
Connecting to ssh://test@uncloud2.vm.science.ru.nl
Connected to cluster.
NAME    MODE     REPLICAS   IMAGE                                                 ENDPOINTS
caddy   global   2          registry.science.ru.nl/cncz/sys/image/caddy:v0.1.14   (custom Caddy config)
~~~

with
``` yaml
    x-ports:
      - 80:80@host
      - 443:443@host
      - 443:443/udp@host
      - 131.174.88.0/24:53:53/tcp@host
      - 131.174.88.0/24:53:53/udp@host
```

pkg/api/port.go
```diff
@@ -23,6 +23,9 @@ type PortSpec struct {
 	Hostname string
 	// HostIP is the host IP to bind the PublishedPort to. Only valid in host mode.
 	HostIP netip.Addr
+	// HostPrefix is the host prefix IP to bind the PublishedPort to. Only valid in host mode. Either HostIP
```
psviderski:
```suggestion
	// HostPrefix is the host prefix to bind the PublishedPort to. Only valid in host mode. Either HostIP
```

pkg/api/port.go
psviderski, line 60: We should validate forbid `HostPrefix` outside the host mode and reject simultaneous `HostIP` + `HostPrefix` for host mode. Please add tests for this as well.

pkg/api/port.go
```diff
@@ -138,24 +145,33 @@ func ParsePortSpec(port string) (PortSpec, error) {
...
-		switch protocol {
-		case ProtocolHTTP, ProtocolHTTPS, ProtocolTCP, ProtocolUDP:
-			spec.Protocol = protocol
-			specifiedProtocol = protocol
-		default:
-			return spec, fmt.Errorf("unsupported protocol: '%s'", protocol)
-		}
+	switch len(parts) {
+	case 2:
+		// 53:5353/udp@host
```
psviderski: This is also the protocol-less case now: `192.168.76.0/24:53:5353@host`
It might be worth rewriting the parsing logic to split by colons `:` first and then parse slashes in individual components to handle the ambiguity and provide more meaningful errors.
Please cover this case in tests as well.

pkg/api/port.go
```diff
@@ -211,7 +227,11 @@ func ParsePortSpec(port string) (PortSpec, error) {
 			}
 
 			if spec.HostIP, err = netip.ParseAddr(ip); err != nil {
-				return spec, fmt.Errorf("invalid host IP '%s': %w", parts[0], err)
+				spec.HostPrefix, err = netip.ParsePrefix(ip)
+				if err != nil {
+					// error with the original error
+					return spec, fmt.Errorf("invalid host IP '%s': %w", parts[0], err)
```
psviderski, line 248: The original error is now overridden, so you might want to use different variables for the errors, e.g. `errIP`, `errPrefix`

internal/machine/docker/interface.go
```diff
@@ -0,0 +1,34 @@
+package docker
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+)
+
+// addrOfPrefix checks the interface and returns the address of each interface that container the prefix.
```
psviderski: Should likely be ... that the prefix contains

internal/machine/docker/interface.go
```diff
@@ -0,0 +1,34 @@
...
+				continue
+			}
+			nip, _ := netip.ParseAddr(ipnet.IP.String()) // round about way is needed to get ipv6 addrs, not mapped v4 in v6.
+			if prefix.Contains(nip) {
+				addrs = append(addrs, nip.String())
+			}
+		}
+	}
+	if len(addrs) == 0 {
+		return nil, fmt.Errorf("no addresses are contained in prefix '%s'", prefix)
```
psviderski:
```suggestion
		return nil, fmt.Errorf("no host addresses are contained in prefix '%s'", prefix)
```

cmd/uncloud/service/run.go
```diff
@@ -97,11 +97,12 @@ func NewRunCommand(groupID string) *cobra.Command {
...
 			"Examples:\n"+
-			"  -p 8080/https                  Publish port 8080 as HTTPS via reverse proxy with default service-name.cluster-domain hostname\n"+
-			"  -p app.example.com:8080/https  Publish port 8080 as HTTPS via reverse proxy with custom hostname\n"+
+			"  -p 8080/https                       Publish port 8080 as HTTPS via reverse proxy with default service-name.cluster-domain hostname\n"+
+			"  -p app.example.com:8080/https       Publish port 8080 as HTTPS via reverse proxy with custom hostname\n"+
 			// TODO: add support for publishing L4 tcp/udp ports.
 			//"  -p 9000:8080                   Publish port 8080 as TCP port 9000 via reverse proxy\n"+
-			"  -p 53:5353/udp@host            Bind UDP port 5353 to host port 53")
+			"  -p 53:5353/udp@host                 Bind UDP port 5353 to host port 53\n"+
+			"  -p 192.168.76.0/24:53:5353/udp@host Bind UDP port 5353 to host port 53 on all interfaces that contain the prefix 192.168.76.0/24")
```
psviderski: I think the description is a bit off as the prefix should contain addresses, not the other way around

website/docs/3-concepts/2-ingress/2-publishing-services.md
```diff
@@ -45,20 +45,23 @@ extension:
...
+| ------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------- |
+| `8000/http`                          | Publish port 8000 as HTTP via Caddy using hostname `<service-name>.<cluster-domain>`                                            |
+| `app.example.com:8080/https`         | Publish port 8080 as HTTPS via Caddy using hostname `app.example.com`                                                           |
+| `127.0.0.1:5432:5432@host`           | Bind TCP port 5432 to host port 5432 on loopback interface only                                                                 |
+| `53:5353/udp@host`                   | Bind UDP port 5353 to host port 53 on all network interfaces                                                                    |
+| `192.168.76.0/24:5432:5432/udp@host` | Bind TCP port 5432 to host port 5432 on all interfaces that have 192.168.76.\* as an address, not `/udp` or `/tcp` is mandatory |
```
psviderski: According to the description and intention (Postgres port?), the example should have `/tcp@host`

Not sure I understand what that means:
> not `/udp` or `/tcp` is mandatory

website/docs/3-concepts/2-ingress/2-publishing-services.md
```diff
@@ -78,8 +81,8 @@ services:
     image: app:latest
     x-ports:
       - example.com:8000/https
-      - www.example.com:8000/https  # The same port can be published with multiple hostnames
-      - api.domain.tld:9000/https   # Another port can be published with a different hostname
+      - www.example.com:8000/https # The same port can be published with multiple hostnames
+      - api.domain.tld:9000/https # Another port can be published with a different hostname
```
psviderski: Is this auto-formatting in IDE? Not a big deal but the original version is IMO a bit easier to read

psviderski (commented):
Overall looks good but please see a comment about another unhandled case when splitting by slashes. We might want to rework the entire logic to start splitting by colons first.

website/docs/3-concepts/2-ingress/2-publishing-services.md
miekg: yeah, some autofmt, I'll revert,as it should not be part of this change.

pkg/api/port.go
miekg: done. That little format is pretty tricky...

pkg/api/port.go
miekg: done

pkg/api/port.go
miekg: split of in different error

miekg:
ptal, should be done now

B08Z:
Thanks for adding prefix-scoped host ports.

The snag is with **globally-deployed** services on a cluster where the nodes aren't all on the same subnet. `addrOfPrefix` errors when a machine has no interface in the prefix:

```go
if len(addrs) == 0 {
    return nil, fmt.Errorf("no host addresses are contained in prefix '%s'", prefix)
}
```

That error is propagated by the caller, so `create container` fails on every node outside the prefix and the whole `mode: global` deploy aborts. It only works if *all* nodes share the subnet.

**Motivating case — DNS on two LANs.** I want one global DNS service to listen on `:53` on each LAN it's attached to, scoped per LAN so it never collides with `uncloudd`'s mesh DNS:

```yaml
x-ports:
  - 192.168.10.0/24:53:53@host   # only node-A is on this LAN
  - 192.168.50.0/24:53:53@host   # only node-B is on this LAN
```

With the current behavior this can't be expressed at all: the prefix is checked **per-prefix, per-machine**, so node-A errors on the `192.168.50.0/24` prefix and node-B errors on `192.168.10.0/24` — every node fails on the prefix it doesn't match. The intended outcome is: node-A binds the `.10` port and skips `.50`, node-B binds `.50` and skips `.10`, and any node on neither LAN skips both — deploy still succeeds.

Could a zero-match prefix on a `@host` port be treated as "publish nothing on this machine, carry on" rather than a hard error (at least for global services)? If skipping by default feels too implicit, an explicit opt-in marker would be fine too — e.g. `192.168.10.0/24:53:53@host?` meaning "best-effort, skip where unmatched."

---
Not a problem if this seems too niche its probably a use case that only affect me...

miekg:
I'll be brief here, because this is better discussed in a new issue. I think this can be done, I'm a little worried if we can detect "none of the IPs could be bound to" and only then return an error

PR #359, opened 2026-05-07
feat(dns): implement extra dns names for machine in m.internal

### LINKED ISSUE #340: Create more DNS names for machines
Via discord:

Ah right, we need to add dns names for the machines as well. Like we already have [machine-scoped DNS names for services](https://uncloud.run/docs/concepts/services/internal-dns#machine-id-scoped-service-name), we can have a namespace for machine names, e.g. `m.internal`, to separate from service names:

- `m.internal` - IPs of all machines in the cluster
- `<machine-name>.m.internal` - single machine by name
- `<machine-id>.m.internal` - single machine by ID

TXT records could be added later to return additional metadata for the machines when we have a use case for this.

`.m` might not be very clear what it means for new folks. Other options: `.machine`, `_machine`, `_m`.

I personally like shortcuts and when I need to type less. That's why I even added an alias for the `machines` command to type it quicker with `uc m ls`. We also have `-m` flag used in many commands so the `.m` DNS namespace seems to nicely fit the current naming in other places.

miekg:
This implements \<machinename\>.m.internal and <machineid>.m.internal as names that can be queried in the cluster.

<s>Listing all machines/services is not implemented, this may come later.</s>

Fixes: #340

### DIFF
--- a/internal/machine/dns/resolver.go
+++ b/internal/machine/dns/resolver.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/psviderski/uncloud/internal/machine/api/pb"
+	"github.com/psviderski/uncloud/internal/machine/network"
 	"github.com/psviderski/uncloud/internal/machine/store"
 )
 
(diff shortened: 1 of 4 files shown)


jabr (approved):

miekg:
> Listing all machines/services is not implemented, this may come later.

listing all machine for a query for m.internal is now implemented too

PR #357, opened 2026-05-06
feat: uc image rm

miekg:
Allow for image removal via the uc cmd line. This just does 'rm' not 'prune', but adds a missing option to uncloud to at least have the possibility to remove images remotely. SSH is of course still an option.

All the plumbing and docker rpcs is done. It needs an e2e test still.

See #167

PR #364, opened 2026-05-08
feat: image completion

### LINKED ISSUE #361: completion for uc image
Would be nice to have completion here as well.

1 complication is that you can have remote and local images

miekg:
this adds completion for both local and remote images. The current commands that deal with images are using this.

Fixes: #361


miekg:
Once https://github.com/psviderski/uncloud/pull/357 is merged, this should be added there too

PR #368, opened 2026-05-13
goreleaser: build uc instead of uncloud

### LINKED ISSUE #367: [BUG] nightly uncloud build contains uncloud instead uc as binary
This is minor, but the uncloud tar.gz doesn't contain uncloud as `uc`, but as `uncloud`. not a biggy, but having the uc binary in there makes most sense

miekg:
Fixes: #367

Seems to build the right thing as shown below:
(obvs can't test in production)

```sh
[1] % GOOS="linux" \
GOARCH="arm64" \
/tmp/goreleaser build --single-target --snapshot --clean
  • skipping validate...
  • cleaning distribution directory
  • loading environment variables
  • getting and validating git state
    • ignoring errors because this is a snapshot     error=git doesn't contain any tags - either add a tag or use --snapshot
    • using tags                                     previous=<unknown> current=v0.0.0
    • pipe skipped or partially skipped              reason=disabled during snapshot mode
  • parsing tag
  • setting defaults
    • DEPRECATED: archives.builds should not be used anymore, check https://goreleaser.com/deprecations#archivesbuilds for more info
    • brews is being phased out in favor of homebrew_casks, check https://goreleaser.com/deprecations#brews for more info
  • partial
  • snapshotting
    • building snapshot...                           version=0.1.0-nightly-519825c
  • running before hooks
    • running                                        hook=go mod tidy
  • ensuring distribution directory
  • setting up metadata
  • writing release metadata
  • loading go mod information
  • build prerequisites
  • building binaries
    • partial build                                  match=target=linux_arm64_v8.0
    • partial build                                  match=target=linux_arm64_v8.0
    • building                                       paths=cmd/uncloudd binaries=uncloudd target=linux_arm64_v8.0
    • building                                       paths=cmd/uncloud binaries=uc target=linux_arm64_v8.0
  • size reports
    • 49.5MiB                                        path=dist/uncloudd_linux_arm64_v8.0/uncloudd
    • 93.25MiB                                       path=dist/uncloud_linux_arm64_v8.0/uc
  • writing artifacts metadata
  • you are using deprecated options, check the output above for details
  • build succeeded after 2s
  • thanks for using GoReleaser!
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -22,7 +22,7 @@ snapshot:
 builds:
   - id: uncloud
     main: ./cmd/uncloud
-    binary: uncloud
+    binary: uc
     env:
       # Use CGO to be able to build fsevent required by compose on macOS.
       - CGO_ENABLED={{ if eq .Os "darwin" }}1{{ else }}0{{ end }}


tonyo:
`uc` is a symlink to `uncloud`, created by the [installation script](https://github.com/psviderski/uncloud/blob/main/scripts/install_cli.sh), which will break with this change.

miekg:
hmm, as does the deb:
```
% dpkg-deb -c uncloud_0.19.0-1+forky_arm64.deb | grep uc
lrwxrwxrwx root/root         0 2026-04-24 11:56 ./usr/bin/uc -> /usr/bin/uncloud
```

psviderski:
Btw, at the beginning I wasn't sure whether to call the CLI `uncloud` or just `uc` so decided to have both (`uncloud` + `uc` symlink).
Now all the docs mention only `uc` so I assume everyone uses `uc`. Is there any value in keeping `uncloud` as well? Maybe we should just rename `cmd/uncloud` -> `cmd/uc` and build and install only `uc` for consistency?

This is a breaking change for 3rd-party packaging but as far as I know we only have two (arch and debian). And it's much easier to make this change now rather than later.

@tonyo @miekg what do you think?

miekg:
Yes that would be best thing to do. Also saves a few key strokes whenever I
compile uc: gob -o uc

Cheers

PR #376, opened 2026-05-25
fix: reject extra args in context commands

immanuwell:
## What
`uc context` commands were a bit too chill with extra args.

On `main`, these bad calls still run:
```bash
uc context show extra
uc context ls extra
uc context use prod extra
```

`show` and `ls` ignore the extra arg and exit 0.
`use` drops into the interactive selector. Kinda weird.

This adds Cobra arg validation across the context command family, plus a small table test.

## Repro
Use any config with at least one context, then run the commands above.

Expected:
all of them fail fast with an arg error.

## Checks
```bash
go test ./cmd/uncloud/context
go test ./cmd/uncloud/... ./internal/cli/... ./pkg/api/... ./pkg/client/...
```

### DIFF
--- a/cmd/uncloud/context/args_test.go
+++ b/cmd/uncloud/context/args_test.go
@@ -1,0 +1,103 @@
+package context
+
+import (
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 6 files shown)


psviderski (approved):
Thanks!

PR #378, opened 2026-05-26
Fix completion with direct connections

### LINKED ISSUE #377: [BUG] completion with only --connect is a noop
When there is no incloud config the completion is a noop:

```
mv ~/.config/uncloud/config.yaml /tmp
% uc --connect test@uncloud1.vm.science.ru.nl inspect <TAB>
...nothing...
```

bit strange 'cause I would have expected it to work out-of-the-box, but connection string isn't seen/detected by the completion function or some such

codeafridi:
fixes #377 

## Summary

- Initialize the CLI before Cobra completion runs.
- Read global flags during completion so `--connect` works without a config file.
- Disable connection progress output during completion.

## Testing

- `GOCACHE=/tmp/go-build GOMODCACHE=/tmp/go-mod go test ./cmd/uncloud ./internal/cli/completion`
- `GOCACHE=/tmp/go-build GOMODCACHE=/tmp/go-mod go build -o /tmp/uc ./cmd/uncloud`
- `HOME=/tmp/uc-nohome /tmp/uc __complete --connect tcp://bad inspect ''`

Full `go test -p 1 ./...` passed for non-e2e packages. `test/e2e` failed with existing Docker cluster runtime errors about invalid IPv6 addresses.

### DIFF
--- a/cmd/uncloud/main.go
+++ b/cmd/uncloud/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/psviderski/uncloud/internal/machine"
 	"github.com/psviderski/uncloud/internal/version"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 type globalOptions struct {
@@ -43,6 +44,8 @@ func main() {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Cobra initializers run before completion and may hit an init error
+			// before PersistentPreRunE gets a chance to run.
 			if initErr != nil {
 				return initErr
 			}
@@ -124,6 +127,9 @@ func initCLI(cmd *cobra.Command, opts globalOptions) error {
 	cli.BindEnvToFlag(cmd, "connect", "UNCLOUD_CONNECT")
 	cli.BindEnvToFlag(cmd, "context", "UNCLOUD_CONTEXT")
 	cli.BindEnvToFlag(cmd, "uncloud-config", "UNCLOUD_CONFIG")
+	if opts.connect == "" {
+		opts.connect = connectFlagFromCompletionArgs(os.Args[1:])
+	}
 
 	var conn *config.MachineConnection
 	if opts.connect != "" {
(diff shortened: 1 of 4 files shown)


cmd/uncloud/main.go
```diff
@@ -36,64 +36,17 @@ func main() {
...
-						Unix: opts.connect[len("unix://"):],
-					}
-				} else {
-					// Default: system ssh CLI command (no prefix or ssh:// prefix).
-					dest := strings.TrimPrefix(opts.connect, "ssh://")
-					conn = &config.MachineConnection{
-						SSH: config.SSHDestination(dest),
-					}
-				}
+			if initErr != nil {
```
miekg, line 49: might be nice to add a comment that initErr may be set in initCLI and we check for that here, is a bit non-obvious

cmd/uncloud/main.go
```diff
@@ -158,3 +115,102 @@ func main() {
...
+			opts.configPath = args[i+1]
+			i++
+		case strings.HasPrefix(arg, "--uncloud-config="):
+			opts.configPath = strings.TrimPrefix(arg, "--uncloud-config=")
+		}
+	}
+	return opts
+}
+
+func setCLIContext(cmd *cobra.Command, uncli *cli.CLI) {
```
miekg, line 193: why is this changed from

> `cmd.SetContext(context.WithValue(cmd.Context(), "cli", uncli))`

?

cmd/uncloud/main.go
```diff
@@ -158,3 +115,102 @@ func main() {
...
+
+	uncli, err := cli.New(configPath, conn, opts.context)
+	if err != nil {
+		return fmt.Errorf("initialise CLI: %w", err)
+	}
+	setCLIContext(cmd.Root(), uncli)
+	return nil
+}
+
+func globalOptsFromArgs(opts globalOptions, args []string) globalOptions {
```
miekg: this looks odd and fragile... you have to do your own parsing for something that is a package that is supposed to do this?

miekg (changes requested):

cmd/uncloud/main.go
codeafridi: addded a comment explaining this. initCLi can run from Cobras initializer before PersistentPreRunE

cmd/uncloud/main.go
codeafridi: completion callbacks can run on the leaf command context so setting it only on the root command is not enough. so added comment here too

cmd/uncloud/main.go
codeafridi: yeah i shortened it now and it reads --connect with pflag for the completion path

miekg:
```
% ./uc --connect test@uncloud1.vm.science.ru.nl inspect  <TAB>                                                                               ~miekg/uncloud/cmd/uncloud pr-378
 completions
alertmanager   caddy          debug          grafana-data   ip             ip-php         prometheus     unprune
atomdns        caddy-metrics  grafana        helloworld     ip-data        ntfy           unprom         unredis
```
while I have moved my config.yaml out of the way. Works for me.

miekg (approved):

psviderski:
Thank you for the PR. I ended up reworking the fix on main using a bit simpler approach: https://github.com/psviderski/uncloud/commit/351698c280e1a2368bbb5dde3e3a4c8991b15bb9                                                                                 
I kept your `ConnectClusterWithOptions(ctx, cli.ConnectOptions{})` change in the completion helpers.

PR #382, opened 2026-05-29
Migrate DNS server to dnsv2

### LINKED ISSUE #341: move dns impl to dnsv2
Move the current dns code to codeberg/miekg/dns, the v2 impl. This is a much more performant impl. and it would be good exercise to surface missing features in the dns lib (unlikely), but still good to check.

Also miekg/dns v1 is deprecated from my point of view.

zasdaym:
Closes #341.


miekg:
That was easier than I thought. Haven't look indepth yet. But for
forwarding you could now just resend Msg.Data without fiddling with it.
Should be less code and less policy in uncloudd

internal/machine/dns/server.go
```diff
@@ -271,15 +270,12 @@ func (s *Server) forwardRequest(req *dns.Msg, proto string) (*dns.Msg, error) {
...
-		Net:     proto,
-		Timeout: forwardingTimeout,
-	}
+	ctx, cancel := context.WithTimeout(ctx, forwardingTimeout)
+	defer cancel()
 
 	var lastErr error
 	for _, server := range s.upstreamServers {
-		resp, _, err := client.Exchange(req, server.String())
+		resp, err := forwardRaw(ctx, req.Data, proto, server.String())
```
miekg: this should be as simple as io.Copy, if it's not are you ran into problem I would like to know, because then the lib. needs to change.

internal/machine/dns/server.go
```diff
@@ -362,14 +457,29 @@ func parseNameserversFromResolvConf() ([]netip.Addr, error) {
...
 }
 
+func newResponse(req *dns.Msg, rcode uint16) *dns.Msg {
+	resp := new(dns.Msg)
+	dnsutil.SetReply(resp, req)
+	resp.Rcode = rcode
+	return resp
+}
+
+func truncate(msg *dns.Msg, maxSize int) {
```
miekg: this is probably wrong, as you will returns split up RRsets (which is an unfortunate DNS thing). But apparantly this is difficult enough that this needs a helper.  https://codeberg.org/miekg/dns/issues/882

internal/machine/dns/server.go
```diff
@@ -362,14 +457,29 @@ func parseNameserversFromResolvConf() ([]netip.Addr, error) {
...
-	if !dns.IsSubDomain(InternalDomain, name) {
+	name = dnsutil.Canonical(name)
+	if !dnsutil.IsBelow(InternalDomain, name) {
 		return name
 	}
 
 	return strings.TrimSuffix(name, "."+InternalDomain)
 }
 
+func newResponse(req *dns.Msg, rcode uint16) *dns.Msg {
```
miekg: you can, prolly 100% of the time, just the incoming message Reset it and reuse it.

zasdaym: Yeah, let's make it simpler.

internal/machine/dns/server.go
zasdaym: Why io.Copy instead of dns.Exchange?

go.mod
```diff
@@ -351,10 +353,23 @@ require (
...
+// codeberg.org/miekg/dns transitively requires newer versions of the modules below than Caddy v2.8 can
+// compile against, so pin them back to the versions Caddy supports. The left side of a replace directive is
+// version-specific: if a future dependency bump selects a newer version of one of these modules, the
+// corresponding replace silently stops applying and must be updated. Remove these replaces once Caddy is
+// upgraded to a version compatible with the newer module versions.
+replace go.uber.org/zap/exp v0.3.0 => go.uber.org/zap/exp v0.2.0
+
+replace github.com/caddyserver/certmagic v0.25.2 => github.com/caddyserver/certmagic v0.21.4
+
+replace github.com/caddyserver/zerossl v0.1.5 => github.com/caddyserver/zerossl v0.1.3
```
psviderski: @miekg why do we need this Caddy stuff for a DNS library? Does it provide a DNS server now as well?

As we have migrated from the JSON caddy config (https://github.com/psviderski/uncloud/blob/5cc005a4237bf7e5367e4570cce5e5f90748d155/internal/machine/caddyconfig/jsonconfig.go) to Caddyfile generation, we can now drop the `caddy/v2` Go dependency which is quite heavy.

But this PR seems to brings some of the caddy packages back.

miekg: honestly don't know - this pr is on my list to review again
(or do it myself first and then see what I come up with)

It should not have to do anything with caddy, as both dns and dnsv2 can be used independently. Even though the deps for dnsv2 are heavier than they should be, as *atomdns* is also there - on the TODO to rip that out as well. That should prolly by done before including it here as a dep - but as said, I want to look at this more closely.

(there is no rush to merge this btw)

zasdaym: @psviderski we can remove zerossl but not with certmagic.

psviderski: Thank you both for the context. I'll then leave this PR to you guys to decide how we go with this migration

PR #383, opened 2026-05-29
fix: corrosion logs are now service logs

miekg:
Get the system logs for corrosion from docker now that corrosion is a system image instead of the systemd unit.

(Nice to see that the logging refactoring makes this super easy)


miekg:
not tested, my clusters are broken atm

miekg:
and trying to test: 
`Error: stream logs for system container 'uncloud-corrosion': inspect service: not found`

which makes sense because it is *not* a service in that sense, that would have been nice

miekg:
ok, this doesn't work. We need ContainerLogs which gives us 2 extra problems:
1. requires a machine ID, to proxy to that machine
2. LogEntry vs ServerLogEntry which needs another conversion.

PR #384, opened 2026-05-29
feat: add uc proxy command

### LINKED ISSUE #379: kubectl port-forward equiv in uc
kubectl has [port forward](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_port-forward/), which allows you to connect to a non-exposed service. This would be useful here as well. Connect to a DB or check of prometheus has a metric, etc.

I don't think this will be hard to implement, have to double check what kubectl has for niceties, but I believe uc already implements most of it (i.e. connect to a service, uc will choose container etc.)

Also port-forward is a terrible name.

miekg:
This allow to set up a proxy from the current host to a container running inside the cluster.

Fixes: #379

### DIFF
--- a/cmd/uncloud/proxy.go
+++ b/cmd/uncloud/proxy.go
@@ -24,10 +24,10 @@ type proxyOptions struct {
 func NewProxyCommand() *cobra.Command {
 	opts := proxyOptions{}
 	cmd := &cobra.Command{
-		Use:   "proxy [LOCAL_PORT:]CONTAINER:REMMOTE_PORT",
-		Args:  cobra.ExactArgs(1),
+		Use:   "proxy SERVICE [LOCAL_PORT:]REMOTE_PORT",
+		Args:  cobra.ExactArgs(2),
 		Short: "Proxy to service.",
-		Long: `Proxy to a container on the remote port.
+		Long: `Proxy to a service on the remote port.
 
 If no local port is provided a random port will be chosen.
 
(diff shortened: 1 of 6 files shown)


miekg:
Need:

- to get the local addres, but proxy.Run() does the listen, so we need to get afterwards
- need p.Run() check for errors, and then wait for ^C
- some tui stuff

miekg:
I tried using the progress stuff, but nothing showed up on the screen, no idea why.

But this works, if we like it, I'll finish it

cmd/uncloud/proxy.go
```diff
@@ -0,0 +1,137 @@
...
+		OnError: func(err error) {
+			fmt.Printf("Failed to proxy to '%s': %v\n", remoteAddr, err)
+			cancel()
+		},
+	}
+
+	go p.Run(ctx)
+
+	// The internal URL should not be doing TLS, as caddy would terminate that, so prefix with http:// with
+	// the assumption that it is running a webservice as this allow control-click on most terminals.
```
psviderski: I'm a bit hesitant with such a bold assumption. I bet proxying a port to a DB is a valid quite popular use case so http:// will be misleading for it.

I'm thinking maybe let's try to print http:// for a list of popular ports that are most likely http, e.g. 80, 8000, 8080, 9090

psviderski (approved):
Looks nice, thank you! I'll push a few minor changes to this branch and merge now

cmd/uncloud/proxy.go
miekg: yeah, you're right. It's mostly "nice to click on in the terminal". Fine with whatever changes you see fit!

Issue #76, opened 2025-06-20
Add support for Compose secrets with custom drivers


psviderski:
Basic secrets support from files: https://github.com/psviderski/uncloud/issues/75

I'm thinking of extending the Compose secrets spec using custom drivers in a similar way compose uses drivers for volumes or logging. Still keep them client side but allow to fetch them from popular secrets managers. See how Kamal supports various secret adapters: https://kamal-deploy.org/docs/commands/secrets/

Something like this:
```
secrets:
  my_secret:
    driver: 1password|bitwarden|vault
    driver_opts:
      path: Vault/secret_name/field_name
```

fredix:
Hi !
maybe I can suggest to replace vault with openbao (https://openbao.org/). With the acquisition of Hashicorp, the Vault API could diverge, and only Openbao is under a free license. Unless you think you can handle both ?

hrstoyanov:
I was gonna ask for this in discord, but very happy it is on your radar already! Kamal's integrations are genius - I love how they leverage the various CLIs. Not sure how much of these can be incorporated in UC, but an AI agent can look at the Kamal code and translate them to Go and UC quickly - these wrappers are not very difficult code, imo.

psviderski:
Just stumbled upon a project that implements 1password secret driver for Docker Swarm: https://github.com/MrMarble/opsd

hrstoyanov:
@psviderski there are a lot of secrets stores out there, I would suggest use the cli as kamal does, it would be hard to get apis for all.

psviderski:
I agree, supporting an arbitrary command would cover most of the use cases so this should definitely be supported first 👍

hrstoyanov:
A suggested hack - get your favorite AI thing and ask it to translate all the Kamal secrets  Ruby code into Go! I think the  Kamal team covered almost everything out there by now  .. you will have 80% done in no time. The rest should be mostly testing ...

miekg:
seems both configs and secrets could benefit from an `x-command` option that runs the command, captures stdout and uses *that* as the config/secret.

For secrets this would mean a single API to support, while keeping the entirety of how to do secrets out of uncloud.

## update

it does make your compose file non-portable... I can see that being "not good"

miekg:
#401 implements something similar to 'secrets engine', if we want to support defining how to get the secret *in the compose file*, like so:
```yaml
services:
  alpine:
    image: alpine
    environment:
      BLA: secret://geheim

secrets:
  geheim:
[...]

PR #385, opened 2026-06-03
feat: add secrets

miekg:
This adds (limited) support for secrets. In the end we want to support the new 'secrets engine' from docker, but this still requires secrets to be put in a container.

This piggybacks on configs as much as possible to make the PR as small as possible.

We use /run/secrets to store them this is a tmpfs so nothing lands on disk.
If a container path is used, we symlink that to /run/secrets, so only a symlink lands on disk, not the secret it self.

When storing the container in corrosion we use base64 encryption, see the TODO there to do proper encrytion.

Update the compatiblity matrix to say we have limited support for secrets.

See: #75 and possibly #76

Note for #76, we will def. want to use the secrets engine of docker: https://github.com/docker/secrets-engine

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -716,6 +716,13 @@ func (s *Server) CreateServiceContainer(
 		return nil, status.Errorf(codes.Internal, "inject configs: %v", err)
 	}
 
+	// Inject secrets into the created container
+	if err = s.injectSecrets(ctx, resp.ID, spec.Secrets, spec.Container.SecretMounts); err != nil {
+		// Remove the container if secret injection fails
+		_ = s.client.ContainerRemove(ctx, resp.ID, container.RemoveOptions{RemoveVolumes: true})
+		return nil, status.Errorf(codes.Internal, "inject secrets: %v", err)
+	}
+
 	respBytes, err := json.Marshal(resp)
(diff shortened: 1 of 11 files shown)


joeblew999:
This would help us directly. We inject two tokens today via `${VAR}` env substitution — `CLOUDFLARE_API_TOKEN` for Caddy's ACME **DNS-01** challenge, and `MOLTIS_TOKEN` for an app — and the env path appends a trailing `\n` (#393), which on the CF token makes caddy-dns/cloudflare reject it, so we currently `tr -d '\r\n'` it in the container. Mounting these as `/run/secrets` **files** instead looks like the clean fix (and is more secure than env), and Caddy can read its token from a file — so we'd drop the trim workaround. 👍

One thing worth confirming given #393: does the **secret file content** itself stay byte-exact, or could it pick up a trailing `\n` (from the stored value or the file write)? If a file-mounted token gains a `\n`, a consumer like caddy-dns/cloudflare would reject it the same way — so it'd be great if the file is the raw bytes with no added newline. Happy to test this end-to-end (CF DNS-01 via a file secret) once it lands.

miekg:
note this fails, but looks unrelated, can't click re-run tests,

miekg:
we have secrets

Issue #369, opened 2026-05-16
uc down or uc undeploy would be nice


miekg:
I think I'm missing the opposite of `uc deploy` to stop a previous started deployment. `compose` has *down*, but then again it also has *up*...

think *deploy* should be an alias of *up* and we should implement a *down* as well

jabr:
Perhaps `uc undeploy`? 😄

psviderski:
Wil throw some ideas that may influence the design. I've been thinking of adding a new Application (app) concept. A compose file defines an app that consists of several services. This introduces a namespace and grouping for services.

The top-level [`name`](https://github.com/compose-spec/compose-spec/blob/main/spec.md#name-top-level-element) attribute in Compose can define the name of the app (defaults to the working dir name).

```
uc app ls
uc app rm
uc app inspect
```
when working with services, we should likely start specifying what app we want to use:
```
uc service ls -a app
uc service -a app rm SERVICE
```
Having to specify `-a` is a bit annoying. Maybe all shortcut `uc` commands should only work with the app defined in the local Compose file. If no compose, `-a` is required.

I think this is closer to Docker Swarm's `docker stack deploy/rm` than to `docker compose up/down`.

miekg:
oh yes! The namespace bit is really nice to have, never used swarm, but to have `uc ls` include a namespace, i.e.

```
monitoring/prometheus
monitoring/alertmanager
```
would also be a cool to exactly see what belongs together.

Adding the namespace in dns is then also in the cards? `prometheus.monitoring.internal` ? Together with everything under `monitoring.internal`.

psviderski:
I'm thinking of apps as not generic namespaces like in k8s but as 1:1 mapping to what you define in the Compose file - typically related services that comprise an "app". I've never liked specifying `-n NAMESPACE` with `kubectl` so that's why some defaults would be helpful when working on a particular app.

>Adding the namespace in dns is then also in the cards? prometheus.monitoring.internal ?

Yeah, I think this is necessary to do as well.

miekg:
yep, totally agree. It _only_ is a _namespace_, I'm now mimicking it in compose with an `un` prefix for our stuff. Was hoping `un/xxxx` was allowed, but no. Namespaces would fix this.

Volumes (and maybe other things) are excluded from this?

psviderski:
>Volumes (and maybe other things) are excluded from this?

[...]

PR #394, opened 2026-06-07
feat: add destroy

miekg:
feat: add destroy

This adds destroy (aliases: down, undeploy) which calls out to client.RemoveService to remove services from the cluster. It is the opposite of "deploy" (which now has the alias "up")

This was mostly cut&paste of existing functionality.

Cancelling the operation:

    % ~miekg/uncloud/cmd/uncloud/uc destroy
    NAME    MODE     REPLICAS
    unkey   global   3

    Proceed with destroy? [y/N] n

    Destroy cancelled. No changes were made.

Giving a service:

     ~miekg/uncloud/cmd/uncloud/uc destroy ds
     Error: select services: no such service: ds

Destroying takes place:

     NAME    MODE     REPLICAS
    unkey   global   3

    Proceed with destroy? [y/N] y

    [+] Destroy service unkey 3/3
     ✔ Container unkey/cafab403586d on machine-11zv  Removed  3.9s
     ✔ Container unkey/f0773dc8c97d on machine-z00h  Removed  4.6s
     ✔ Container unkey/b1b428fabcd8 on machine-36i2  Removed  4.1s

Signed-off-by: Miek Gieben <miek@miek.nl>

### DIFF
--- a/cmd/uc/deploy.go
+++ b/cmd/uc/deploy.go
@@ -38,8 +38,9 @@ type deployOptions struct {
 func NewDeployCommand() *cobra.Command {
 	opts := deployOptions{}
 	cmd := &cobra.Command{
-		Use:   "deploy [FLAGS] [SERVICE...]",
-		Short: "Deploy services from a Compose file.",
+		Use:     "deploy [FLAGS] [SERVICE...]",
+		Aliases: []string{"up"},
+		Short:   "Deploy services from a Compose file.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cli.BindEnvToFlag(cmd, "yes", "UNCLOUD_AUTO_CONFIRM")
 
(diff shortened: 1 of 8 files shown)


miekg:
fixes: #369

miekg:
Do we want to move this to a deployment plan thing? Or just call stop.Service which also has nice output. A deployment plan seems also a bit overkill as the plan is 'remove all containers'

psviderski:
I'm a bit confused about what you want to actually achieve with the `undeploy`. From the issue discussion, I thought you basically want to remove all the services defined in your Compose file. Maybe `destroy` is a better name. This PR stops them. What's the practical use case?

For destructive operations we don't seem to have a consistent prompt. But I don't think we need anything fancy, just list the resources that will be removed like `uc volume ls` does.

miekg:
>  This PR stops them. What's the practical use case?

s/stops/remove 

Thanks for the feedback, I'll move towards that.

PR #395, opened 2026-06-07
fix: machine init/add fall back to plain output when stdout isn't a TTY (#386)

### LINKED ISSUE #386: uc machine init requires a TTY in non-interactive shells (`could not open TTY`)
Uncloud version: `uc version 0.19.0`

## Problem

`uc machine init` cannot run in a non-interactive shell (CI, automation, a piped/backgrounded process) because its progress UI needs a TTY. Even with `-y/--yes` set, it fails:

```
$ uc machine init root@<ip> --no-dns -y    # run without a controlling TTY
Error: wait for cluster to be ready: bubbletea: error opening TTY: bubbletea: could not open TTY: open /dev/tty: device not configured
```

The `-y/--yes` help text says it "should be explicitly set when running non-interactively, e.g., in CI/CD pipelines", which implies non-interactive use is supported — but `-y` only auto-confirms prompts; the bubbletea progress view still requires `/dev/tty`. `uc ctx` (e.g. `uc ctx rm`) has the same TTY requirement.

## Steps to reproduce

1. From a process with no controlling terminal (CI runner, `nohup`, a script whose output is piped), run:
   `uc machine init root@<server-ip> --no-dns -y`
2. It errors with `could not open TTY` instead of completing.

## Workaround

Allocating a pseudo-TTY works — the cluster is actually created fine; only the progress spinner needs the TTY:

```
script -q /dev/null uc machine init root@<server-ip> --no-dns -y
```

## Request

A non-interactive / plain-output mode for `uc machine init` (and `uc ctx`): detect the absence of a TTY and fall back to plain line-by-line log output (or add a `--no-tui`/`--plain` flag). That would let Uncloud be driven from CI and IaC tooling without a PTY hack.

## Context

Hit while automating Uncloud-on-Hetzner provisioning with mise + OpenTofu: https://github.com/joeblew999/vm-uncloud

joeblew999:
Fixes #386.

`uc machine init` and `uc machine add` use `huh` readiness spinners that open `/dev/tty`, so they die in shells with no controlling terminal (CI, SSH automation, task runners) even with `-y`. Today the only workaround is wrapping the command in a PTY (`script -q /dev/null …`), which also swallows the real error on failure.

This guards both spinners with `tui.IsStdoutTerminal()` and falls back to a plain log + direct wait when stdout isn't a terminal — the exact pattern `connectClusterWithProgress` already uses. Headless now works without a PTY, and errors surface normally.

- `cmd/uncloud/machine/init.go` — "Waiting for the cluster to be ready..."
- `cmd/uncloud/machine/add.go` — "Waiting for the machine to join the cluster..."

No behaviour change in a terminal (spinner as before).

### DIFF
--- a/cmd/uncloud/machine/add.go
+++ b/cmd/uncloud/machine/add.go
@@ -163,19 +163,28 @@ func add(ctx context.Context, uncli *cli.CLI, remoteMachine *cli.RemoteMachine, 
 	}
 
 	// Wait for the cluster to be initialised on the machine to be able to deploy the Caddy service.
-	err = spinner.New().
-		Title(" Waiting for the machine to join the cluster...").
-		Type(spinner.MiniDot).
-		WithTheme(spinner.ThemeFunc(func(isDark bool) *spinner.Styles {
-			return &spinner.Styles{
-				Spinner: lipgloss.NewStyle().Foreground(lipgloss.Yellow),
-				Title:   lipgloss.NewStyle(),
-			}
-		})).
-		ActionWithErr(func(ctx context.Context) error {
-			return machineClient.WaitClusterReady(ctx, 5*time.Minute)
-		}).
-		Run()
+	// The spinner opens /dev/tty, so it dies headless (CI, SSH, mise tasks). Fall
+	// back to a plain wait when stdout isn't a terminal — mirrors
+	// connectClusterWithProgress, and lets `uc machine init/add` run without a PTY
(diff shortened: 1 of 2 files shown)


miekg:
Think it may make more sense to put more things I the client/tui package
and do things there. This also allows for a, maybe, different kind of
spinner used in both cases.

The duplicated comment also points to the need for something slightly more
generic imo

Cheers

Issue #209, opened 2025-12-07
Provide windows CLI with scoop


gpetrov:
Please compile the uncloud Go cli for windows target and provide the binary to be installed with scoop.sh

This will define increase the user base of uncloud and will make it much more accessible.

Also it should work all fine because of the great Go target compilation support for windows as well the great support for ssh and ssh agent that works fine out of the box on latest windows 11.

psviderski:
Thank you for the request!
Is this a hypothetical nice-to-have or you're a potential user who misses the native windows binary?

I wonder how much worse the experience is when using WSL as noted [here](https://uncloud.run/docs/getting-started/install-cli)?
It's very hard for me to judge and navigate the Windows ecosystem as I haven't touched windows for more than 10 years. So having someone using windows and interested in testing uncloud on it would be very helpful. Please let me know if this is you who reads this (not necessarily @gpetrov).

gpetrov:
Well more that 50% of our users use windows to publish their web sites to various hosting and cloud providers.

So yes the windows user base is pretty large.

Using WSL adds an additional layer of having an embedded Linux vm on windows that uses huge amount of additional memory and takes time to spin up (as well it often crashes or gets stuck) so it has a bad user experience.

As I said Go can compile natively to windows target so it is a no brainer to just add windows as target in go and allow uncloud cli to be used by all windows users directly.

You might want to see what happened to bun.sh (NodeJS alternative) when they also added windows support. Now they have huge user base and are getting investments and even bought by Anthropic

jabr:
Out of curiosity, I tried building the cli with `GOOS=windows` and it hit (at least) these issues where even the CLI has some unix (macOS, linux, etc) dependencies:

- internal/machine/docker/client_exec.go:74:28: undefined: unix.SIGWINCH
- internal/machine/machine.go:573:17: undefined: sockets.NewUnixSocket
- internal/machine/cluster.go:36:27: undefined: network.WireGuardNetwork
- internal/machine/cluster.go:72:24: undefined: network.NewWireGuardNetwork

It might be possible to disable some of the CLI functionality to get a Windows build, but full support will likely require some platform-specific work. Unfortunately, I also don't have and haven't used Windows in more than a decade either.

psviderski:
[...]

PR #399, opened 2026-06-11
feat(uc): build uncloud CLI for Windows

ApexDevelopment:
This PR makes it possible to build the `uc` tool (not the daemon) for Windows, making it possible to orchestrate clusters and deploy images from a Windows machine. To accomplish this without stubbing out Windows-specific functions in the daemon side of the code, a refactor was required to separate out code that the daemon needs but that `uc` does not.

OpenSSH on Windows does not have the ControlMaster multiplexing feature. Instead, the Windows build takes advantage of `ssh -D` to create a SOCKS proxy through which multiple connections can be made via one SSH session. This bypasses the need to open multiple SSH connections to the cluster machines, which can trip max session limits.

Tested on my own Windows machine deploying my own production services. Still builds fine within WSL, so Linux build should be unaffected, but should be verified on an actual Linux machine. Test suite passes on my machine.

Issue #209 related


ApexDevelopment:
Rebased on 6266e9e

internal/machine/token/token.go
```diff
@@ -1,4 +1,4 @@
-package machine
+package token
```
miekg, line 1: why is this changed in this PR?

internal/machine/network/mtu_windows.go
```diff
@@ -0,0 +1,10 @@
+//go:build windows
+
+package network
+
+import "errors"
+
+// detectEgressMTU is a stub for Windows. The machine daemon that performs detection only runs on Linux.
```
miekg: This PR is only about the `uc` client running on windows, no? Is this needed?

internal/machine/docker/client_exec_windows.go
```diff
@@ -0,0 +1,7 @@
+package docker
+
+import "os"
+
+// notifyResizeSignal is a no-op on Windows, which does not deliver terminal
+// resize notifications via signals. The initial window size is still sent.
+func notifyResizeSignal(ch chan<- os.Signal) {}
```
miekg: elsewhere you do a runtime `GOOS` check, here it's a build tag, doing a runtime check in `handleTerminalResize` makes everything shorter as well (if that builds) and doesn't conflate two ways of doing things in the same PR

internal/machine/constants/constants.go
```diff
@@ -5,4 +5,11 @@ const (
 	MachineAPIPort = 51000
 	// UnregistryPort is the port for the embedded container registry listening on the machine IP.
 	UnregistryPort = 51500
+
+	// DefaultMachineSockPath is the default path to the machine API Unix socket.
+	DefaultMachineSockPath = "/run/uncloud/machine.sock"
+	// DefaultUncloudSockPath is the default path to the Uncloud API Unix socket.
+	DefaultUncloudSockPath = "/run/uncloud/uncloud.sock"
+	// DefaultSockGroup is the Linux group that owns the API sockets, granting access without root.
+	DefaultSockGroup = "uncloud"
```
miekg: why this change in this PR?

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +230,96 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
 	return conn, nil
 }
 
+// socksTunnelDialer lazily starts a single `ssh -D` SOCKS proxy to the destination and returns a dialer that
```
miekg: what is lazily about this?

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +230,96 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+// routes connections through it. The tunnel is started once per connector and reused by all subsequent dials,
+// so any number of concurrent connections are multiplexed over a single SSH connection. The tunnel is torn
+// down by Close. It uses the system ssh client, so it honours ~/.ssh/config host aliases, agents and keys
+// exactly like the rest of the connector.
+func (c *SSHCLIConnector) socksTunnelDialer(ctx context.Context) (proxy.ContextDialer, error) {
+	c.socksOnce.Do(func() {
+		port, err := freeLocalPort()
```
miekg: I never seen something like `freeLocalPort()`, you just give port 0 to listen directly and let the kernel figure it out at that point. Surely there is a way to get the actual port number if you need it

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +230,96 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+		return 0, err
+	}
+	defer l.Close()
+	return l.Addr().(*net.TCPAddr).Port, nil
+}
+
+// waitForSOCKSReady polls addr until a TCP connection succeeds, the context is cancelled, or a 15s deadline
+// is reached. It gives the ssh SOCKS proxy time to authenticate and bind its local port.
+func waitForSOCKSReady(ctx context.Context, addr string) error {
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
```
miekg: the highest wait in this file in *main* is 10s, otherwise 5s, why does this need 15s max?

internal/machine/network/mtu_windows.go
ApexDevelopment: The import graph reaches it, so it's needed in that sense. I intended to remove these kinds of stubs, so I can do that here. Doing so will probably mean removing the existing Darwin stub too. Is that ok?

internal/machine/docker/client_exec_windows.go
ApexDevelopment: Got it, will do it that way.

pkg/client/connector/sshcli.go
ApexDevelopment: Nothing, weird Claude-ism. Will remove.

pkg/client/connector/sshcli.go
ApexDevelopment: No reason, just a choice Claude made, will adjust to 10s.

pkg/client/connector/sshcli.go
ApexDevelopment: We do need it so the Go SOCKS dialer can connect to it, I'm having a hard time finding a good way to get what the port number is. Working on it now.

ApexDevelopment: Okay, OpenSSH does not appear to accept port 0 as an argument for -D. It needs a concrete port and I don't see a way around it. I can make searching for an open port more durable by opening a TCP listener on 127.0.0.1:0, getting the port, closing the listener, and then trying to use that port for ssh. Technically something could intervene in the time it takes to pass the port to ssh, but a bounded retry loop should help recover. What do we think?

internal/machine/token/token.go
ApexDevelopment: This was done to avoid depending on daemon code for the CLI build so that Windows stubs could be removed, as requested by @psviderski.

internal/machine/constants/constants.go
ApexDevelopment: `DefaultUncloudSockPath` needed to move out for import graph untangling, same reason as my reply for `token.go`. `DefaultSockGroup` and `DefaultMachineSockPath` were for consistency, not technically necessary, so could be reversed if you wanted, just a code hygiene kind of thing.

miekg: ack. Personally I like PRs a short as possible, if something is not needed it should become an issue or deferred to another (follow-up) PR to make things nicer

pkg/client/connector/sshcli.go
miekg: ah, that's the what the whole dance is for `-D <port>`, and as you say its not guaranteed. Think we should pick a port with an option to override, makes the error message back to the user better as well, now there is a (small) chance it just fails and you can only re-run `uc` hoping it would work. In a new error message we can say fiddle with this option.

internal/machine/constants/constants.go
ApexDevelopment: Reverted.

pkg/client/connector/sshcli.go
ApexDevelopment: Any default port preference? 😁

internal/machine/network/mtu_windows.go
ApexDevelopment: Bumping this thread 😁

pkg/client/connector/sshcli.go
miekg: dunno, `uncloudd` has ports in the 5xxxx range, might do that here as well? 51022 ?

internal/machine/network/mtu_windows.go
miekg: dunno. Let's first minimize the entire PR and see what's left to fix/do different

ApexDevelopment:
Ready for more review. SOCKS port can be overridden with a new env var `UNCLOUD_SSH_SOCKS_PORT`.

miekg:
starting to look good, eyeballed https://patch-diff.githubusercontent.com/raw/psviderski/uncloud/pull/399.diff
will check this pr out tomorrow and do a final review.

thanks

internal/machine/docker/client_exec.go
```diff
@@ -69,15 +70,18 @@ func setupTerminal(ctx context.Context, stream pb.Docker_ExecContainerClient) (f
...
 	if size, err := term.GetWinsize(inFd); err == nil {
 		_ = sendResizeRequest(stream, size)
 	}
 
+	if runtime.GOOS == "windows" {
+		return nil
+	}
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.Signal(28))
```
miekg: what's 28?

... and the next file `mtu.go` is still using build tags. I think this entire PR should be changed to using build tags for the windows build; currently it makes the unix build unreadable. Whereby handleTerminalResize should be entirely stubbed out, not making some helper function

pkg/client/connector/sshcli.go
```diff
@@ -114,6 +138,13 @@ func (c *SSHCLIConnector) Connect(ctx context.Context) (*grpc.ClientConn, error)
 // include control socket settings for connection reuse if the path is configured and useControlMaster is true.
 // The remote command is not included and should be appended by the caller.
 func (c *SSHCLIConnector) buildSSHArgs(useControlMaster bool) []string {
```
miekg, line 140: I can't really see why this change is needed?

Again using build tags and stubbing this out is prolly much cleaner

pkg/client/connector/sshcli.go
```diff
@@ -19,6 +21,14 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	defaultSSHSOCKSHost = "127.0.0.1"
+	// defaultSSHSOCKSPort is the local port used for the Windows ssh -D SOCKS tunnel. It can be
+	// overridden with UNCLOUD_SSH_SOCKS_PORT if another process already uses the default.
+	defaultSSHSOCKSPort = 51022
+	sshSOCKSPortEnv     = "UNCLOUD_SSH_SOCKS_PORT"
```
miekg: can't this not be done in cli by letting *cobra* handle this, just like the rest?

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +239,126 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+			c.socksErr = err
+			return
+		}
+
+		// The tunnel must outlive the dial context that triggered it, so Close cancels this background context.
+		tunnelCtx, cancel := context.WithCancel(context.Background())
+		args := append(c.sshOptions(false), "-o", "ExitOnForwardFailure=yes", "-N", "-D", socksAddr, c.config.Destination())
+		cmd := exec.CommandContext(tunnelCtx, "ssh", args...)
+		var stderr strings.Builder
+		cmd.Stderr = &stderr
```
miekg:
```suggestion
		var stderr &bytes.Buffer{}
		cmd.Stderr = stderr
```

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +239,126 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+			return "", 0, fmt.Errorf("invalid %s %q: must be a TCP port from 1 to 65535", sshSOCKSPortEnv, value)
+		}
+	}
+
+	return net.JoinHostPort(defaultSSHSOCKSHost, strconv.Itoa(port)), port, nil
+}
+
+func isLocalForwardBindFailure(detail string) bool {
+	detail = strings.ToLower(detail)
+	return strings.Contains(detail, "address already in use") ||
```
miekg: this is super fragile to say the least... just return the error if not nil and suggest a different port could used

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +239,126 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+// waitForSOCKSReady polls addr until a TCP connection succeeds, the context is cancelled, or a 10s deadline
+// is reached. It gives the ssh SOCKS proxy time to authenticate and bind its local port.
+func waitForSOCKSReady(ctx context.Context, addr string, sshDone <-chan struct{}, sshErr *error) error {
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+	for {
+		conn, err := (&net.Dialer{Timeout: time.Second}).DialContext(ctx, "tcp", addr)
```
miekg: this net.Dailer can be moved out of the loop, usage here also looks somewhat funny

pkg/client/connector/sshcli.go
```diff
@@ -194,6 +239,126 @@ func (c *SSHCLIConnector) DialContext(ctx context.Context, network, address stri
...
+	c.socksOnce.Do(func() {
+		socksAddr, socksPort, err := sshSOCKSAddr()
+		if err != nil {
+			c.socksErr = err
+			return
+		}
+
+		// The tunnel must outlive the dial context that triggered it, so Close cancels this background context.
+		tunnelCtx, cancel := context.WithCancel(context.Background())
+		args := append(c.sshOptions(false), "-o", "ExitOnForwardFailure=yes", "-N", "-D", socksAddr, c.config.Destination())
```
miekg: sshBuildARG has been changed, but then isn't used here? #confused

miekg (changes requested):

pkg/client/connector/sshcli.go
ApexDevelopment: On vacation currently, will get back to this on Sunday. In the meantime would appreciate a unified answer from both you and @psviderski on whether I should be using stubs or avoiding stubs

miekg:
think we should merge #414 first and then focus on getting sock proxy support. That support can be useful in general case as well, not just for windows. Then we can make it so that we default to `true` for sock proxy use under windows only

ApexDevelopment:
> think we should merge #414 first and then focus on getting sock proxy support. That support can be useful in general case as well, not just for windows. Then we can make it so that we default to `true` for sock proxy use under windows only

Okay, so, with all due respect, #414 is nearly exactly the same as the first commit in this PR. If that's what you wanted, I would have happily submitted just that commit.

The SOCKS proxy is necessary because without multiplexing, `uc` tries to open too many simultaneous SSH connections to the machines which hits concurrent session limits. I ran into this reliably when working on my machine until I implemented the proxy.

miekg:
hmm, yes, I see - I never looked at the initial commit, but that's indeed very much like #414 
the subsequent changes didn't make it better though :(

ApexDevelopment:
> hmm, yes, I see - I never looked at the initial commit, but that's indeed very much like #414 the subsequent changes didn't make it better though :(

The subsequent changes were entirely driven by maintainer feedback - but either way, new changes from your recent feedback are in. If you would rather merge #414 first, that's fine, I can rebase my changes on that, but it _will_ produce nonfunctional Windows builds until the SOCKS proxy is in.

ApexDevelopment:
Squashed messy commits after f9cc739 and rebased branch on b7e224a.

PR #403, opened 2026-06-25
initial support for 'x-command' and 'driver: exec' secrets, referenced as secret://name in env

### LINKED ISSUE #76: Add support for Compose secrets with custom drivers

psviderski:
Support defining secrets that get their value from a command output:
```yaml
secrets:
  token:
    x-command: op read op://vault/myapp/token
```
which is a short form of the following long syntax:
```yaml
secrets:
  token:
    driver: exec
    driver_opts:
      command: op read op://vault/myapp/token
```

The secret value could also be fetched from a file or environment variable as defined in the standard compose spec: https://github.com/compose-spec/compose-spec/blob/main/09-secrets.md

Then a secret value could be referenced as an environment variable value:
```yaml
services:
  foo:
    image: foo
    environment:
      TOKEN: secret://token

secrets:
  token:
    x-command: op read op://vault/myapp/token
```

A secret definition is kept separate from the reference format (not encoded as e.g. `uc://plugin/reference`) to have only one way of defining secrets - top level `secrets` section - and then reference them in `environment` or mount as files: https://github.com/compose-spec/compose-spec/blob/main/05-services.md#secrets (will be implemented later).

Resolves #76

### DIFF
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -111,9 +111,9 @@ func ResolveSecrets(ctx context.Context, project *types.Project) error {
 }
 
 // secretValue resolves a secret to its value depending on its source: an 'exec' driver command,
-// an environment variable, or a file. The command output is trimmed of surrounding whitespace as
-// command-line tools commonly append a trailing newline; environment and file values are returned
-// verbatim.
+// an environment variable, or a file. A single trailing newline ('\n' or '\r\n') is stripped from the command
+// output as command-line tools commonly append one. All other whitespace is kept. Environment and file values are
+// returned verbatim.
 func secretValue(ctx context.Context, secret types.SecretConfig, project *types.Project) (string, error) {
 	switch {
 	case secret.Environment != "":
@@ -136,7 +136,11 @@ func secretValue(ctx context.Context, secret types.SecretConfig, project *types.
 		if err != nil {
 			return "", err
 		}
-		return strings.TrimSuffix(out, "\n"), nil
+		// Strip a single trailing newline, treating it as '\r\n' on Windows.
+		if trimmed, ok := strings.CutSuffix(out, "\n"); ok {
+			out = strings.TrimSuffix(trimmed, "\r")
+		}
+		return out, nil
 	default:
 		return "", fmt.Errorf("secret has no source: define one of 'file', 'environment', '%s', or 'driver'",
 			SecretCommandExtensionKey)
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/compose-spec/compose-go/v2/types"
+	"github.com/mattn/go-shellwords"
 )
 
 const (
@@ -137,9 +138,18 @@ func secretValue(ctx context.Context, secret types.SecretConfig, project *types.
 	}
 }
 
-// runSecretCommand runs command via 'sh -c' in workingDir with the given environment and returns its stdout.
+// runSecretCommand runs the command in workingDir with the given environment and returns its stdout.
+// The command runs directly without a shell, so shell features need an explicit shell, e.g. 'sh -c "cmd1 | cmd2"'.
 func runSecretCommand(ctx context.Context, command, workingDir string, env []string) (string, error) {
-	cmd := exec.CommandContext(ctx, "sh", "-c", command)
+	args, err := shellwords.Parse(command)
+	if err != nil {
+		return "", fmt.Errorf("parse command '%s': %w", command, err)
+	}
+	if len(args) == 0 {
+		return "", fmt.Errorf("command '%s' is empty", command)
+	}
+
+	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
 	cmd.Dir = workingDir
 	cmd.Env = env
 
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -1,0 +1,160 @@
+package compose
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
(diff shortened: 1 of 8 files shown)


pkg/client/compose/secret.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	return nil
+}
+
+// secretValue resolves a secret to its value depending on its source: an 'exec' driver command,
+// an environment variable, or a file. The command output is trimmed of surrounding whitespace as
+// command-line tools commonly append a trailing newline; environment and file values are returned
+// verbatim.
+func secretValue(ctx context.Context, secret types.SecretConfig, project *types.Project) (string, error) {
```
miekg, line 137: secrets in docker are `[]byte`, don't think it matters much

pkg/client/compose/secret.go
```diff
@@ -0,0 +1,160 @@
...
+		return strings.TrimSuffix(out, "\n"), nil
+	default:
+		return "", fmt.Errorf("secret has no source: define one of 'file', 'environment', '%s', or 'driver'",
+			SecretCommandExtensionKey)
+	}
+}
+
+// runSecretCommand runs command via 'sh -c' in workingDir with the given environment and returns its stdout.
+func runSecretCommand(ctx context.Context, command, workingDir string, env []string) (string, error) {
+	cmd := exec.CommandContext(ctx, "sh", "-c", command)
```
miekg: this seems overkill... if you want the shell, you should put this in the `x-command` line

pkg/client/compose/secret_test.go
```diff
@@ -0,0 +1,313 @@
...
+    image: foo
+    environment:
+      TOKEN: secret://token
+  bar:
+    image: bar
+    environment:
+      TOKEN: secret://token
+secrets:
+  token:
+    x-command: "echo run >> %s; printf abc"
```
miekg: you want `/bin/echo` here

pkg/client/compose/secret_test.go
```diff
@@ -0,0 +1,313 @@
...
+			name: "command fails with stderr",
+			content: `
+services:
+  foo:
+    image: foo
+    environment:
+      TOKEN: secret://token
+secrets:
+  token:
+    x-command: echo boom >&2; exit 3
```
miekg: `/bin/echo`

pkg/client/compose/secret_test.go
```diff
@@ -0,0 +1,313 @@
...
+	// '$$' escapes the '$' so Compose doesn't interpolate it; the shell expands it at run time.
+	composeYAML := `
+services:
+  foo:
+    image: foo
+    environment:
+      TOKEN: secret://token
+secrets:
+  token:
+    x-command: printf '%s' "$$DOTENV_ONLY"
```
miekg:
```suggestion
    x-command: /bin/sh -c 'printf %s "$$DOTENV_ONLY"'
```
hope i didn't mess up the quotes

pkg/client/compose/secret_test.go
```diff
@@ -0,0 +1,313 @@
...
+			name: "short form x-command",
+			content: `
+services:
+  foo:
+    image: foo
+    environment:
+      TOKEN: secret://token
+secrets:
+  token:
+    x-command: printf 'topsecret'
```
miekg, line 62: these all need to be `sh -c`'s or actual executables

pkg/client/compose/secret.go
miekg: it's somewhat annoying as you need to do the shell split yourself here then. With something like https://pkg.go.dev/github.com/buildkite/shellwords

miekg:
we prolly want to make a difference between `sh -c` and plain exec, i.e. the docker cmd vs entrypoint, or shell-form and exec-form

pkg/client/compose/secret.go
psviderski: What do you mean by secrets in docker? Secrets engine, swarm secrets, or which ones?
I was thinking that if we're mostly dealing with env vars which are strings it makes sense to use string here as well which doesn't interpret bytes on its own in Go anyway.

pkg/client/compose/secret.go
psviderski: Yes, you're right, this format is inconsistent with the existing service `command`, `endpoint`, and `x-pre_deploy.command`. I'd want to reuse their `ShellCommand` type to support both string and list formats but we're limited by the string type in `driver_opts` (`map[string]string`).

We can still do `ShellCommand` for `x-command` only but I don't like to have even more inconsistency. So I guess a list won't be supported.

And I still think it's good to keep the long syntax with `driver: exec` to have the possibility to options in the future if needed, e.g. a custom timeout or something like that.

pkg/client/compose/secret_test.go
psviderski: Why? `echo` works well in both exec and shell formats

pkg/client/compose/secret.go
psviderski: Removed the `sh -c` wrapping so it's up to the user to add it explicitly if shell features are needed

pkg/client/compose/secret.go
miekg: yeah, I not sure it matters as long as you don't range over the strings or that kind of actions.

But anything secret related in the docker code is a `[]byte`

PR #406, opened 2026-06-30
chore: use cli.Cancelled in all cases

miekg:
Use `cli.Cancelled` in all cases where we cancel an operation instead of Printf && return nil


psviderski (approved):
Thanks! 👍

PR #411, opened 2026-07-06
Add machine: scale all global services

### LINKED ISSUE #388: feat: scale all global services when adding a machine
Would make life slightly easier. For caddy this is already supported. But caddy maybe one of many global services

miekg:
This lists all global services and scales them to the new machine that just was added. This re-uses a bunch of code from add.go in the new scale.go

Fixes: #388

### DIFF
--- a/cmd/uc/machine/scale.go
+++ b/cmd/uc/machine/scale.go
@@ -49,18 +49,18 @@ func scale(ctx context.Context, clusterClient, machineClient *client.Client, unc
 			}
 			d, err := clusterClient.NewCaddyDeployment(caddyImage, "", api.Placement{})
 			if err != nil {
-				return fmt.Errorf("create caddy deployment: %w", err)
+				return fmt.Errorf("create %s deployment: %w", service.Name, err)
 			}
-			plan, err = d.Plan(ctx)
+			if plan, err = d.Plan(ctx); err != nil {
+				return fmt.Errorf("plan %s deployment: %w", service.Name, err)
+			}
 
 		default:
 			spec := service.Containers[0].Container.ServiceSpec
 			d = clusterClient.NewDeployment(spec, nil)
-			plan, err = d.Plan(ctx)
-		}
-
-		if err != nil {
-			return fmt.Errorf("plan %s deployment: %w", service.Name, err)
+			if plan, err = d.Plan(ctx); err != nil {
+				return fmt.Errorf("plan %s deployment: %w", service.Name, err)
+			}
 		}
 
 		if len(plan.Operations) == 0 {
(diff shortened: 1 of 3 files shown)


miekg:
I need a way to test this, worst case: manual testing

website/docs/9-cli-reference/uc_machine_add.md
```diff
@@ -22,7 +22,7 @@ uc machine add [USER@]HOST[:PORT] [flags]
 ```
   -h, --help                  help for add
   -n, --name string           Assign a name to the machine. (default is the machine's hostname)
-      --no-caddy              Don't deploy Caddy reverse proxy service to the machine.
+      --no-caddy              Don't deploy Caddy reverse proxy service to the machine. This also cancels scaling any global services.
```
psviderski, line 25: >This also cancels scaling any global services.

This feels odd. Do we even need this flag for `machine add` if we scale all global services now?
If we want to keep the flag, maybe `--no-deploy`, `--no-scale`, or `--no-global-scale` are better names.

cmd/uc/machine/scale.go
```diff
@@ -0,0 +1,120 @@
...
+			if err != nil {
+				return fmt.Errorf("create %s deployment: %w", service.Name, err)
+			}
+			if plan, err = d.Plan(ctx); err != nil {
+				return fmt.Errorf("plan %s deployment: %w", service.Name, err)
+			}
+
+		default:
+			spec := service.Containers[0].Container.ServiceSpec
+			d = clusterClient.NewDeployment(spec, nil)
```
psviderski, line 60: What do you think about making it possible to reduce the target destination (machines) for a particular service deployment? So that you can specify a subset of machines you want to rollout your service to before rolling them out to all machines specified in the compose.

I'm thinking it's useful for testing new changes on only one replica. For example:
```
uc deploy -m <only-this-machine>
```
which deploys to `x-machines` intersected with `<only-this-machine>`. This might not be the ideal interface as it becomes less flexible when having multiple replicas running on one machine. Any better ideas?

Same for scaling, specify which machine(s) to scale the service to:
```
uc scale SERVICE -m <this-machine>
```

Then with this capability, we will be able to scale all global service to only the new machine here without running a new deployment across the entire cluster. In a general case, cluster-scope deployments may have more undesired side effects and failures that IMO shouldn't be tight to the `machine add` command. It should set up a machine and deploy stuff related only to it, not fix degraded or partially failed deployments on other machines.

cmd/uc/machine/scale.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+
+		if len(plan.Operations) == 0 {
+			fmt.Printf("Service %s is up to date.\n", service.Name)
+			continue
+		}
+
+		if !yes {
+			confirmed, err := tui.Confirm("Proceed with deployment?")
```
psviderski, line 72: Did you consider merging plans for N services into one plan instead of printing N plans and prompts?

miekg: yeah, this should be one prompt, or none at all - I don't know why someone *wouldn't* want this

PR #415, opened 2026-07-18
fix(proxy): don't tear down the proxy when a client connection aborts

panakour:
## What happened 

Opening a proxied service in a browser kills `uc proxy` within seconds: 
``` 
127.0.0.1:8090 → 10.210.2.2:8090 (beszel/02eedae88481) 
Failed to proxy to '10.210.2.2:8090': data copy: ... write: broken pipe 
``` 

Browsers abort connections all the time (cancelled prefetches, favicon retries). Each per-connection error goes through `OnError`, which the proxy command handles with `cancel()` — so one aborted request tears down the whole tunnel. 

## The fix 
- Per-connection errors (remote dial, data copy) now go to a new non-fatal `OnConnError` callback the proxy logs them and keeps running. 
- Normal peer-disconnect noise (`EPIPE`, `ECONNRESET`, `net.ErrClosed`) is not even reported. 
- `OnError` + `cancel()` is now only used for listener accept errors. Tested against a live cluster: before, one aborted request killed the tunnel; after, repeated aborts are ignored and the tunnel keeps serving.


miekg:
How can  I reproduce in uncloud? What should I run, just any (static?) website and then `uc proxy`?

panakour:
> How can I reproduce in uncloud? What should I run, just any (static?) website and then `uc proxy`?

Hi @miekg I found this while running Beszel behind `uc proxy` and opening the UI in a browser kept killing the tunnel. 

But it reproduces with plain nginx:

```sh
uc run -n web nginx
uc exec web sh -c 'head -c 5M /dev/urandom > /usr/share/nginx/html/big.bin'
uc proxy web 8080:80
```

Then in another terminal, abort a download mid-transfer:

```sh
curl --limit-rate 50k --max-time 0.5 -o /dev/null http://127.0.0.1:8080/big.bin
```

Screenshot 2026-07-19 at 16 03 23

internal/proxy/proxy.go
```diff
@@ -124,9 +130,15 @@ func (p *Proxy) handleConnection(ctx context.Context, localConn net.Conn) {
...
+			if err != nil && !isBenignConnError(err) && p.OnConnError != nil {
+				p.OnConnError(fmt.Errorf("data copy: %w", err))
 			}
 		}
 	}
 }
+
+// isBenignConnError reports whether err is expected connection noise (the peer
+// closed or aborted the connection mid-transfer) that is not worth reporting.
+func isBenignConnError(err error) bool {
```
miekg:
```suggestion
func isTemporaryError(err error) bool {
```
the *net* package actually has `Temporary()` as a method too, but it's marked deprecated. Make sense to use that language here too

internal/proxy/proxy.go
```diff
@@ -16,6 +18,10 @@ type Proxy struct {
 	RemoteAddr  string
 	DialContext func(ctx context.Context, network, address string) (net.Conn, error)
 	OnError     func(error)
+	// OnConnError is called for non-fatal per-connection errors (remote dial
+	// and data copy failures). A single failed or aborted client connection
+	// must not tear down the whole proxy.
+	OnConnError func(error)
```
miekg: can we not just re-use `OnError` for this?

miekg (changes requested):

internal/proxy/proxy.go
panakour: done!

miekg (approved):

psviderski:
Thank you for the fix. I reworked it in https://github.com/psviderski/uncloud/commit/fa77edf53e7dd9dd4f607a40b7c11964f1edb150 to still surface all errors from the `Proxy` but leave it to the caller to make decisions on how to process certain errors. The `Proxy` implementation is pretty generic so I'm not sure if there will be use cases where silently ignoring errors due to connection abortion won't be desired.

Also improved the current implementation and made it correctly handle all edge cases during context cancellation.

PR #419, opened 2026-07-29
feat: support stdin_open and tty

miekg:
This can be useful to leave a container running without specifying a command like `sleep`, and probably in other situations as well.


psviderski (approved):
Thanks! I pushed a few fixes for the logs streaming to work for containers with allocated TTY

PR #423, opened 2026-08-19
secret: Resolve secrets in x-pre_deploy too

### LINKED ISSUE #422: [BUG] PreDeploy.Env is not stripped before replication to the cluster store, while Container.Env is
## Describe the bug

`normaliseContainerForStore` removes environment variables from a container record before it is written to the replicated Corrosion store, with the comment "Remove the environment variables to avoid leaking secrets". It clears `Config.Env` and `ServiceSpec.Container.Env`, but `ServiceSpec.PreDeploy.Env` holds the same class of data and is left in place.

The result is that the same value is treated differently depending on which compose block it sits in. A value in `environment:` stays on the machine running the container. The identical value in `x-pre_deploy.environment:` replicates to every machine in the cluster.

`internal/machine/store/container.go:91-94`

```go
func normaliseContainerForStore(ctr *api.ServiceContainer) {
	// Remove the environment variables to avoid leaking secrets.
	ctr.Config.Env = nil
	ctr.ServiceSpec.Container.Env = nil
	...
}
```

`ServiceSpec.PreDeploy.Env` is defined at `pkg/api/service.go:461` and is user settable from compose:

* Extension struct: `pkg/client/compose/predeploy.go:14` (`Environment types.MappingWithEquals`)
* Mapped into the spec: `pkg/client/compose/service.go:152-159`

### Path into the store

1. The client marshals the full spec, including `PreDeploy`, into the create request (`pkg/client/container.go:88`).
2. The machine writes the spec to its local `machine.db` (`internal/machine/docker/server.go:754`).
3. `InspectServiceContainer` reads it back onto `ctr.ServiceSpec` (`internal/machine/docker/service.go:127`).
4. The controller syncs both regular and hook containers to the store (`internal/machine/docker/controller.go:163-190`), calling `normaliseContainerForStore` and then inserting into `containers`.

The regular service container carries the full spec too, so the value stays in the replicated store for the life of the service. It is not limited to the short lived hook container.

## How to reproduce

Verified on a 3 machine `ucind` cluster built from this repository at commit `b7e224a`.

1. Create `leak-test.yaml`:

```yaml
services:
  leaktest:
    image: nginx:alpine
    x-machines: machine-1
    environment:
      CONTROL_TOKEN: control-value-aaa
    x-pre_deploy:
      command: ["/bin/true"]
      environment:
        HOOK_TOKEN: hook-value-bbb
```

2. Create the cluster and deploy:

```
make ucind-cluster
uc deploy -f leak-test.yaml
```

3. Read the replicated store on `machine-2`, which is not running the container:

```
docker cp ucind-default-machine-2:/var/lib/uncloud/corrosion/store.db /tmp/store-m2.db
docker cp ucind-default-machine-2:/var/lib/uncloud/corrosion/store.db-wal /tmp/store-m2.db-wal
docker cp ucind-default-machine-2:/var/lib/uncloud/corrosion/store.db-shm /tmp/store-m2.db-shm

sqlite3 /tmp/store-m2.db "SELECT container FROM containers;" \
  | grep -o -E "control-value-aaa|hook-value-bbb" | sort -u
```

Result:

```
hook-value-bbb
```

`control-value-aaa` is absent, confirming that `Container.Env` stripping works. `hook-value-bbb` is present.

The stored value:

```
sqlite3 /tmp/store-m2.db "SELECT container FROM containers;" | grep -o '"PreDeploy":{[^}]*}'

"PreDeploy":{"Command":["/bin/true"],"Env":{"HOOK_TOKEN":"hook-value-bbb"}
"PreDeploy":{"Command":["/bin/true"],"Env":{"HOOK_TOKEN":"hook-value-bbb"}
```

Two rows, the service container and the hook container, both carrying the value.

`machine-2` was running no service container of its own at the time (`docker exec ucind-default-machine-2 docker ps` listed only `uncloud-corrosion`), so the record reached it by replication.

## Expected behavior

`PreDeploy.Env` should be stripped alongside `Container.Env` before the record is written to the replicated store, so that a value behaves the same way regardless of which compose block it is written in.

## Environment

* Uncloud versions:
  * Control (client) node (`uc version`): 999.0.0-dev, git commit `b7e224a1eff98813b1d1a32034d977be24be994e`
  * Uncloud daemon (`uncloudd version`): 999.0.0-dev, built from the same tree via `docker build --target ucind`
* OS version (`uname -a`):
  * Client: `Linux 6.1.0-52-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.180-1 (2026-08-03) x86_64 GNU/Linux`
  * Server: ucind container, Alpine Linux v3.23, same host kernel

`uc version` reports `Git state: dirty`. That comes only from untracked files in my working directory. No tracked source is modified, so the tested code matches `b7e224a` exactly.

### Why this matters in practice

The documented use case for a pre-deploy hook is database migrations. That is exactly where a user would put a database URL or password:

```yaml
x-pre_deploy:
  command: ["./migrate.sh"]
  environment:
    DATABASE_URL: postgres://user:password@db/app
```

Nothing signals that this behaves differently from `environment:` one block above. The field most likely to hold a real credential is the one that is not stripped.

### Scope

This is a gap in an existing control rather than a high severity vulnerability.

* `ResolveSecrets` only walks `service.Environment` (`pkg/client/compose/secret.go:91-99`), so a `secret://` reference placed in hook env passes through unresolved as a literal string. Managed secrets cannot leak through this path.
* Only values written directly into the compose file are affected.
* Reading the store requires root or the `uncloud` service account on a cluster machine, which already has broader access.

The concern is that the asymmetry is invisible to the user.

### Root cause

The stripping is a denylist that names fields explicitly. Git history suggests the omission was accidental rather than deliberate:

* `4cc1e55` (2025-08-14) added the stripping, when the store began holding `ServiceContainer` including the spec.
* `772b31b` (2026-04-06) added the pre-deploy hook feature, eight months later, introducing a second env carrying field on `ServiceSpec`.

Nothing links the two. There is no test in `internal/machine/store/` pinning the stripping behaviour, so the new field passed review without anyone noticing the interaction.

### Suggested fix

Minimal:

```go
ctr.ServiceSpec.Container.Env = nil
if ctr.ServiceSpec.PreDeploy != nil {
	ctr.ServiceSpec.PreDeploy.Env = nil
}
```

Worth considering separately: the current approach requires every future field on `ServiceSpec` that can hold a secret to be remembered and added here, by someone who has no reason to know this function exists. A test over `normaliseContainerForStore` would at least catch the next occurrence.

miekg:
This adds the secret:// resolving to the Environment in x-pre_deploy too. Further more the Environment is set to nil when commiting the container to the store.

Fixes: #422

### DIFF
--- a/internal/machine/store/container.go
+++ b/internal/machine/store/container.go
@@ -92,6 +92,9 @@ func normaliseContainerForStore(ctr *api.ServiceContainer) {
 	// Remove the environment variables to avoid leaking secrets.
 	ctr.Config.Env = nil
 	ctr.ServiceSpec.Container.Env = nil
+	if ctr.ServiceSpec.PreDeploy != nil {
+		ctr.ServiceSpec.PreDeploy.Env = nil
+	}
 
 	// Docker returns Mounts in a non-deterministic order so sort them.
 	slices.SortFunc(ctr.Mounts, func(a, b container.MountPoint) int {
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"maps"
 	"os"
 	"os/exec"
 	"strings"
@@ -132,9 +131,6 @@ func ResolveSecrets(ctx context.Context, project *types.Project) error {
 // a command.
 func HasCommandSecretRefs(project *types.Project) bool {
 	for _, service := range project.Services {
-		if predeploy, ok := service.Extensions[PreDeployHookExtensionKey].(PreDeployHook); ok {
-			maps.Copy(service.Environment, predeploy.Environment)
-		}
 		for _, v := range service.Environment {
 			if v == nil {
 				continue
@@ -145,6 +141,21 @@ func HasCommandSecretRefs(project *types.Project) bool {
 			}
 			if secret, ok := project.Secrets[name]; ok && secret.Driver == secretExecDriver {
 				return true
+			}
+
+			if predeploy, ok := service.Extensions[PreDeployHookExtensionKey].(PreDeployHook); ok {
+				for _, v := range predeploy.Environment {
+					if v == nil {
+						continue
+					}
+					name, ok := secretRefName(*v)
+					if !ok {
+						continue
+					}
+					if secret, ok := project.Secrets[name]; ok && secret.Driver == secretExecDriver {
+						return true
+					}
+				}
 			}
 		}
 	}
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -147,7 +147,6 @@ func HasCommandSecretRefs(project *types.Project) bool {
 				return true
 			}
 		}
-
 	}
 	return false
 }
--- a/pkg/client/compose/secret.go
+++ b/pkg/client/compose/secret.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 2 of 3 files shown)


vky5:
Opened #424 before I saw this, same two fixes. This one doesn't cover the gap for `container.go` though, no test on the storage-side half of the bug. That's the one the issue title is about. I added one for it in 424, plus a test that a secret referenced from both `environment` and `x-pre_deploy.environment` only resolves once.

Also pulled the resolve loop into a shared helper instead of duplicating it in `ResolveSecrets` and `HasCommandSecretRefs`. Not required, just less to keep in sync next time someone adds a field.

Whichever you want, I'll rebase the tests onto this one, or close 424.

PR #426, opened 2026-08-24
feat: add <machine-name>.m.<service>.internal

iRazvan2745:
dont name your machines `nearest` or `rr` :heart:

### DIFF
--- a/internal/machine/dns/resolver.go
+++ b/internal/machine/dns/resolver.go
@@ -101,6 +101,10 @@ func (r *ClusterResolver) updateServiceIPs(containers []store.ContainerRecord) {
 		serviceNameWithMachineID := record.MachineID + ".m." + ctr.ServiceName()
 		newServiceIPs[serviceNameWithMachineID] = append(newServiceIPs[serviceNameWithMachineID], ip)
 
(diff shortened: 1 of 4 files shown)


miekg:
would it make sense to put these names in a different namespace that don't also have nearest or rr, OR make nearest and rr sit in a namespace that can never be a machine name, i.e. `_rr` or `_nearest`, or even `nearest._m...`?

iRazvan2745:
that would cause issues with people using those already. id honestly just not allow those names as machine names, its @psviderski 's call

jabr:
We might want a new "namespace" (say `n` instead of `m`) like `{machine-name}.n.{service}.internal` regardless. 

But we could also not support the nearest/rr options on both of these (machine name or ID), as it would be the same behavior (round robin) either way -- all service instances on that machine would be the same "nearest".

Also, iirc, I considered adding a machine name option way back when I added the machine-id lookup. I think Pasha had concerns with it because machine names can change...

iRazvan2745:
With namespaces it would be M-NAME.m.SERVICE.NAMESPACE.internal.

if the user changes their machine names frequently they can still use the id.

PR #428, opened 2026-08-24
feat(compose): support mac_address

### LINKED ISSUE #427: add mac_address to uncloud
We will move some license servers to uncloud, but these servers are annoying and some require a mac address to be set (as this is registered upstream a long time ago) .

I don't think we support it currenty, and adding it should be easy.

See https://docs.docker.com/reference/compose-file/services/#mac_address

miekg:
Add mac_address supported.

Fixes: #427

### DIFF
--- a/internal/machine/docker/server.go
+++ b/internal/machine/docker/server.go
@@ -562,9 +562,10 @@ func (s *Server) CreateServiceContainer(
 			api.LabelServiceMode: spec.Mode,
 			api.LabelManaged:     "",
 		},
-		User:      spec.Container.User,
-		Tty:       spec.Container.Tty,
-		OpenStdin: spec.Container.OpenStdin,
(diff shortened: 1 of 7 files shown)


miekg:
```
=== NAME  TestServiceSpecFromCompose/full_spec
    service_test.go:310: 
        	Error Trace:	/home/runner/work/uncloud/uncloud/pkg/client/compose/service_test.go:310
        	Error:      	Received unexpected error:
        	            	validating /home/runner/work/uncloud/uncloud/pkg/client/compose/testdata/compose-full-spec.yaml: services.test additional properties 'mac_adress' not allowed
        	Test:       	TestServiceSpecFromCompose/full_spec
```
seems a docker error, which is - I think - the note I've shown in #427 that newer docker engines don't support this anymore.

ok, bummer, need to do this with some command hacking

PR #429, opened 2026-09-02
feat: validate remaining unsupported items of the support-matrix

miekg:
Validate unsupported compose features: external configs, container_name, and placement.

This makes the code match the support-matrix as of now (added comment to that effect as well).

### DIFF
--- a/pkg/client/compose/project_test.go
+++ b/pkg/client/compose/project_test.go
@@ -322,6 +322,40 @@ volumes:
 `,
 			shouldErr: true,
 		},
+		{
+			name: "external configs",
+			composeYAML: `services:
+  app:
+    image: myapp:latest
+configs:
+  http_config:
+    external: true
+`,
+			warnCount:    1,
+			warnContains: []string{"external"},
+		},
+		{
+			name: "container_name",
+			composeYAML: `services:
+  app:
+    image: myapp:latest
+    container_name: my_container
+`,
+			warnCount:    1,
+			warnContains: []string{"container_name"},
+		},
+		{
+			name: "placement",
+			composeYAML: `services:
+  app:
+    image: myapp:latest
+    deploy:
+      mode: replicated
+      replicas: 2
+`,
+			warnCount:    1,
+			warnContains: []string{"placement"},
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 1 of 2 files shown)


pkg/client/compose/service.go
```diff
@@ -489,6 +498,24 @@ func validateServicesFeatures(project *types.Project) []error {
 		if service.StorageOpt != nil {
 			errs = append(errs, err(service.Name, "storage_opt"))
 		}
+		if service.ContainerName != "" {
+			errs = append(errs, err(service.Name, "container_name"))
+		}
+		if service.Deploy != nil {
+			if len(service.Deploy.Placement.Constraints) > 0 || len(service.Deploy.Placement.Preferences) > 0 || service.Deploy.Placement.MaxReplicas > 0 || service.Deploy.Mode != "" {
```
psviderski: `Deploy.Mode` is supported

pkg/client/compose/service.go
```diff
@@ -489,6 +498,24 @@ func validateServicesFeatures(project *types.Project) []error {
...
+				errs = append(errs, fmt.Errorf(
+					"service '%s': placement is not supported, "+
+						"use x-machines instead: %s", service.Name,
+					"https://uncloud.run/docs/compose-file-reference/extensions#x-machines"))
+				if service.Deploy.RestartPolicy != nil {
```
psviderski: These checks shouldn't likely be nested under the `Placement` condition

pkg/client/compose/project_test.go
```diff
@@ -322,6 +322,40 @@ volumes:
...
+		},
+		{
+			name: "placement",
+			composeYAML: `services:
+  app:
+    image: myapp:latest
+    deploy:
+      mode: replicated
+      replicas: 2
+`,
```
psviderski, line 369: This is a valid compose. Please fix the checks below and add tests for other `deploy` fields

psviderski (changes requested):
Thanks! Please fix minor bugs and will merge then

pkg/client/compose/service.go
miekg: damn, missed that. Fixed

pkg/client/compose/service.go
miekg: ack, done and added better checks/warning to what is actually supported.

pkg/client/compose/project_test.go
miekg: done

